PR #108, opened 2014-09-30
use window.dump to make firefox log to stdout

fippo:
This allows outputting the trace messages to stdout using https://developer.mozilla.org/de/docs/DOM/window.dump

Useful to make firefox more active in the testing procedures described either in http://googletesting.blogspot.se/2014/08/chrome-firefox-webrtc-interop-test-pt-1.html or https://blog.andyet.com/2014/09/29/testing-webrtc-applications


KaptenJansson:
What's the latest status here?

fippo:
@phoglund interested in this? If not let's close it here and don't reopen it.

phoglund:
So this improves firefox logging in the test? I'm positive to that in general but I don't have any immediate need of this. Closing is fine.

fippo:
it makes the logs grep-able which enables stuff like turn prober.

PR #127, opened 2014-10-03
minor turn-prober improvements

fippo:
this moves the cleanup to a trap function. Also kills Xvfb the hard way, I ended up having lots of dead Xvfb processes when testing.


KaptenJansson:
Ping, if not usable, please close.

PR #174, opened 2014-10-23
Avoiding duplicate and allowing unset of codec parameter

minyuel:
Avoiding duplicate and allowing unset of codec parameter.

This allows unsetting Opus FEC or other codec parameter. This also avoids duplication.


KaptenJansson:
minyuel, hows the unit test looking as requested by juberti?

minyuel:
I've not got time to add them.

I see that there is a new (since Dec) test framework. I'll add tests in it.

Issue #265, opened 2014-12-04
Record events for AppRTC and other demos


samdutton:
For example, to record numbers for those who, having navigated to apprtc.appspot.com: 
• call gUM
• request a connection
• get a connection
• get errors
...

Could also add events for icon/button clicks on AppRTC and other demos.

vr000m:
In addition to those listed above we have:

- iceDisconnected: if the interface disappears. If iceConnected again on the same or a different candidate pair, then we can pair disconnections and connections to calculate connectivity disruptions.

PR #281, opened 2014-12-06
figure out the local type preference of the active candidate

fippo:
mostly to find out if we're using turn/udp, turn/tcp or turn/tls. Will be useful for logging in #265.
This uses the same trick as the trickle-page to determine the local type preference, based on the comment in https://groups.google.com/d/msg/discuss-webrtc/NfCgW-8FhkE/3chOYjjPcboJ

It's probably in the wrong place and should be done in the completed event (which somehow I don't get?) but this should be enough to get the idea. Currently doesn't properly work for IPv6 (I think) since the local ip will have [] in the stats but not in the candidate line.

I'd still prefer getting the serverUrl in the candidate stats, but until then this hack allows gathering the data

### DIFF
--- a/samples/web/content/apprtc/js/infobox.js
+++ b/samples/web/content/apprtc/js/infobox.js
@@ -65,16 +65,25 @@ function updateInfoDiv() {
       localAddr = activeCandPair.stat('googLocalAddress');
       remoteAddr = activeCandPair.stat('googRemoteAddress');
       localAddrType = activeCandPair.stat('googLocalCandidateType');
-      remoteAddrType = activeCandPair.stat('googRemoteCandidateType');
+      remoteAddrType = activeCandPair.stat('googRemoteCandidateType')
     }
     if (localAddr && remoteAddr) {
-      var localTypePref = getCandidateTypePreference(pc, activeCandPair);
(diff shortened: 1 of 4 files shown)


fippo:
in case anyone stumbles over this: there is another way to do this if one does not want to call getStats... chrome updates the c= line and the port in the m-line with the currently used candidate so all information required for determining the currently active connection

Firefox doesn't (unfortunately/yet)

juberti:
changes are in... you can get the local candidate via connection.localCandidateId -> candidate.priority

fippo:
updated... even though I think this PR is dead. But it's still good for some discussion. Can make a new one if you want.

**formatTypePreference** might go into adapter.js -- Firefox uses 5 for TURN/UDP and (currently) 0 for TURN/TCP. For the remote candidate this is tricky, as one needs to have information about the remote implementation to appropriately format this. Which makes me wonder if it should be displayed for the remote party at all.

**getStatsReport** can't filter by id instead of a property name. Special-casing 'id' is ugly but works.

But hey, I have a month of raw data already and that is what I'm really interested in :-)

juberti:
I think this is still useful for apprtc, can you refile this PR there with the new (candidate stats object-based) approach?

fippo:
will do.

fippo:
done

PR #303, opened 2014-12-16
Add Closure support

jiayliu:
How to use:
Run setup-closure.sh from the same dir as apprtc.py. This will download the closure library and compiler.
Run closure-build.sh to compile the files in closure-src. The output is in closure-out as compile.js, which can be referenced from the HTML file in a script tag.

closure-src/ only has a utils.js for now. We can convert the other pieces gradually.

@phoglund


KaptenJansson:
ping phoglund

samples/web/content/apprtc/setup-closure.sh
```diff
@@ -0,0 +1,15 @@
+rm -rf closure-library
+echo "Downloading Closure library:"
+git clone https://github.com/google/closure-library
+echo "Downloading Closure library done."
+
+echo "Downloading Closure compiler:"
+curl -o compiler.zip http://dl.google.com/closure-compiler/compiler-latest.zip
```
phoglund: Consider passing -s or -sS here depending on how verbose the curl output is.

phoglund:
First, sorry, I completely forgot about this one. Are we still going ahead with this?

I think this should work on the Travis boxes since everything is downloaded on the fly. You can try it by hooking up a shell job which invokes your .sh in the gruntfile (like for the python tests). Then make a new commit and push to your branch. Travis will run on that (similarly to the green checkbox next to your Add Closure Support commit above) and you will see if it works.

How will this work then, will one have to run the .sh before bringing up the apprtc server so all the closure stuff gets generated?

juberti:
I think it makes sense to do this so that we can download a single minified JS lib, but ideally it would be possible to run unoptimized (i.e. what we have now) as well.

jiayliu:
We added a 'build' task to grunt for packaging the Chrome App. We can add
the closure build script to the build task.

We can also use closure to generate unoptimized code into one JS file for
debugging.

I'll update this patch to make that happen.

On Tue, Jan 20, 2015 at 12:00 PM, Justin Uberti <notifications@github.com>
wrote:

> I think it makes sense to do this so that we can download a single
> minified JS lib, but ideally it would be possible to run unoptimized (i.e.
> what we have now) as well.
>
> —
> <https://github.com/GoogleChrome/webrtc/pull/303#issuecomment-70721949>.
>



-- 
Jiayang Liu

jiayliu:
I'll create a new PR since this one is compared against an old branch.

PR #327, opened 2014-12-20
adapter.js: also check for mozRTCPeerConnection

fippo:
apparently there are some custom builds of FF that have disabled webrtc peerconnections but allow GUM. Checking both mozGetUserMedia and mozRTCPeerConnection should catch those as well.

Arguably the same could be done for chromium but...

### DIFF
--- a/samples/web/js/adapter.js
+++ b/samples/web/js/adapter.js
@@ -45,7 +45,7 @@ function maybeFixConfiguration(pcConfig) {
   }
 }
 
-if (navigator.mozGetUserMedia) {
+if (navigator.mozGetUserMedia && mozRTCPeerConnection) {
   console.log('This appears to be Firefox');
 
   webrtcDetectedBrowser = 'firefox';


juberti:
We still want adapter to work if peerconnection is disabled, right? Or do I misunderstand the intent here?

fippo:
We'll probably want adapter to still shim GUM if GUM is available. But setting things like setting
```
window.RTCSessionDescription = mozRTCSessionDescription;
```
should only be tried when mozRTCPeerConnection is set. Makes sense?

Style-wise do you have any preference between
```
if (some kind of GUM is supported) {
   shim GUM stuff
}
if (RTCPeerConnection is supported) {
  shim RTCPeerConnection stuff
}
```
and
```
if (mozrtcpeerconnection ||  mozgetusermedia) {
   if (mozrtcpeerconnection) { shim this}
  if (mozgetusermedia) { shim this}
} else if (webkitrtcpeerconnection || navigator.webkitgetusermedia) {
  ... do the same for chrome
}
```

The second variant seems easier to maintain but when modularizing adapter I'd tend to split GUM and RTC stuff into separate modules so applications which don't need GUM (datachannel-only) don't pull in that dependency.

juberti:
There's some global stuff that needs to happen for both gUM and RTCPC, but I probably need to see a specific example to decide which approach is the best.

fippo:
browser-detection at toplevel
https://github.com/fippo/webrtc/blob/variant1/samples/web/js/adapter.js

feature detection at toplevel:
https://github.com/fippo/webrtc/blob/variant2/samples/web/js/adapter.js
I didn't really know what to do with the global stuff...

variant 2 (which I prefer) seems to me like this should be four modules:
- [RTCPeerConnection](https://www.npmjs.com/package/rtcpeerconnection)
- [getUserMedia](https://www.npmjs.com/package/getusermedia)
- [attachMediaStream](https://www.npmjs.com/package/attachmediastream)
- [webrtcsupport](https://www.npmjs.com/package/webrtcsupport)
I am slightly biased here obviously :-)

fippo:
will file a new bug on adapter.

PR #328, opened 2014-12-22
Prevent vertical scrollbar in report dialog (Linux)

KaptenJansson:
There was an vertical scrollbar appearing at certain vertical resolutions, e.g. 600px. This fixes it by hiding the overflow.


KaptenJansson:
I will look into this next week.

PR #329, opened 2014-12-22
Testrtc wifi scan

andresusanopinto:
Add a test for detecting wifi scans that periodically inject delays into the channels due to the network card stop to scan all other frequencies in a go.

The test is based on measuring delay suffered by a low rate 10Hz data stream when looped back via a TURN server. For now the test has no expectations and due to its delay nature it is only enabled when explicitely enabling it with: "?test_filter=WiFi Periodic Scan".

Plan is to push this into prod and ask for a few reports to write the expectations based on.

### DIFF
--- a/samples/web/.jshintrc
+++ b/samples/web/.jshintrc
@@ -10,6 +10,7 @@
   "undef": true,
   "unused": "strict",
   "globals": {
+    "addExplicitTest": true,
     "addTest": true,
     "arrayAverage": true,
     "arrayMax": true,
--- a/samples/web/content/testrtc/js/bandwidth_test.js
+++ b/samples/web/content/testrtc/js/bandwidth_test.js
@@ -162,3 +162,45 @@ function testVideoBandwidth(config) {
     testFinished();
   }
 }
+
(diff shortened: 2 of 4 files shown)


andresusanopinto:
Ping

KaptenJansson:
I think it makes sense to add some results to the UI, not just in the "Download report" log file. Currently if it fails, it just marks it as failed and the test panel cannot be expanded. It does not tell why it has failed.  

For the result I think it makes sense to display the average delay, min and max. Also if possible display how many times > 150ms was seen and for bonus points at what interval or at least since last time (use start of test as a reference for the first time), for instance on my MBA I can see the delay jumping up to 180-400ms every 2 minutes or so, presumably when the wifi card scans networks.

If you do not agree, at least printout that the user should look at the "Download report" file for the result.

andresusanopinto:
Yep it was failing due to test complete without success or failure.
Added a basic expectations on:
 max delay < (min_delay+100)*2

Also printed the min/max/avg for the user. WIll merge.

PR #336, opened 2014-12-25
trickle demo: use bundle in gathering

fippo:
The description "Gather unbundled RTCP candidates" is slightly misleading since googUseRtpMUX only affects bundle, not rtcp-mux. Also changes the default semantics so one has to turn this off.


juberti:
Unbundled RTCP (i.e. the current state) will be the default for a PeerConnection, so I think that's the right thing to default to here as well.

PR #337, opened 2014-12-25
another trickle demo: measuring setup time

fippo:
Since the original trickle demo is now more a demo of candidate gathering here is one that lets the user compare trickle ice and non-trickle. Mashed up from the trickle and munge demos.
Online version: http://fippo.github.io/webrtc/samples/web/content/peerconnection/trickle-effect/

It measures the time between createOffer and iceconnectionstatechange->connected. For the biggest effect I recommend disabling trickle, using a TURN/TCP or TLS server and having a virtualbox interface.

@samdutton: might be useful as part of the codelab since it demonstrates why trickle ice is such a great idea :-)

### DIFF
--- a/index.html
+++ b/index.html
@@ -96,6 +96,8 @@
 
       <p><a href="samples/web/content/peerconnection/trickle-ice">ICE candidate gathering from STUN/TURN servers</a></p>
 
+      <p><a href="samples/web/content/peerconnection/trickle-effect">Examine the effects of Trickle ICE</a></p>
+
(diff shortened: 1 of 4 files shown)


juberti:
I'm not sure we need another sample page for this - since it is so similar to the existing ice-servers page, could something like this be added into that page (i.e. some estimate of the time saved by trickling?)

fippo:
if my lxcbr0 interface is ignored (soon) this doesn't really make a good example anymore.

PR #341, opened 2014-12-30
Add a status request handler to collider

jiayliu:
This will make it easy to check the status of collider and make it possible to implement a prober with GAE scon service.


jiayliu:
Addressed comments. PTAL.

jiayliu:
Ping @pthatcherg

PR #316, opened 2014-12-18
Objectize the AppRTC javascript and add unit tests

jiayliu:
This is a big refactoring to make the javascript more object oriented and MVC.
I haven't added unit tests for all classes yet, but want to get some feedback before all reviewers are gone for vacation.
New classes:
AppController: the controller that connects the UI and the model "Call"
Call: manages everything needed to make a call
SignalingChannel: wrapper of the WebSocket connection
PeerConnectionClient: wrapper of RTCPeerConnection
InfoBox: wrapper of the info div

CC: @samdutton @tkchin @juberti

PR #345, opened 2015-01-07
Add basic support for packaged Chrome App

chuckhays:
* Adds files required for chrome app
* Creates a landing page with UI to choose a room
* Changes navigation model to single page because chrome app can't navigate between pages
* Adds new GAE calls to support new navigation model
* Adds web fonts to local package for chrome app.
* Adds script to fetch css and fonts.
.

There are a couple of outstanding issues still to be addressed, but they probably make sense to do as separate changes:
* ~~Keyboard key handling conflicts with text entry~~ - completed
* ~~Input into the textbox is not yet sanitized/transformed to ensure it is valid as a room name~~ - completed
* App does not send BYE message when closed - chrome apps have a different model for close events that will require some restructuring
* CSP model in chrome apps is different, the GET requests don't include the required headers, solution needs to be coordinated with server side
* loopback.js and vr.js files are commented out, haven't been investigated

@juberti 
@jiayliu


samples/web/content/apprtc/apprtc.py
```diff
@@ -288,35 +288,42 @@ def get_room_parameters(request, room_id, client_id, is_initiator):
   media_constraints = make_media_stream_constraints(audio, video,
                                                     firefox_fake_device)
   wss_url, wss_post_url = get_wss_parameters(request)
+  jsonParams = json.dumps({
```
jiayliu: This will break mobile clients which depend on the output of get_room_parameters. Why do you have to change this?

samples/web/content/apprtc/js/main.js
```diff
@@ -358,3 +399,142 @@ function toggleFullScreen() {
     trace(event);
   }
 }
+
+roomIdInput.addEventListener('input', function(){
```
jiayliu: Could you move the room related code into a separate file and make it more object oriented? E.g. what https://github.com/GoogleChrome/webrtc/pull/316 tries to do.
Also please avoid using global data variables.

samples/web/content/apprtc/js/util.js
```diff
@@ -83,3 +85,97 @@ function filterTurnUrls(urls, protocol) {
...
+function isChromeApp() {
+  return (typeof chrome !== 'undefined' &&
+          typeof chrome.storage !== 'undefined' &&
+          typeof chrome.storage.local !== 'undefined');
+}
+
+// Get a value from local browser storage. Calls callback with value.
+function getStorage(key, callback) {
+  if (isChromeApp()) {
+    // use chrome.storage.local
```
jiayliu: Capitalize the first letter and end it with '.'. Here and other places.

samples/web/content/apprtc/index.html
```diff
@@ -28,15 +29,36 @@
...
-      <div id="room-link">Waiting for someone to join this room: <a href="{{ room_link }}" target="_blank">{{ room_link }}</a></div>
+      <div id="room-link">Waiting for someone to join this room: <a id="room-link-a" href="{{ room_link }}" target="_blank">{{ room_link }}</a></div>
     </div>
     <div id="info"></div>
     <div id="status"></div>
   </footer>
 
+  <script type="text/javascript">
+  var initialParams;
+{% if not landing_page %}
```
jiayliu: I don't get how it works: when will landing_page be false?

samples/web/content/apprtc/apprtc.py
```diff
@@ -546,5 +579,7 @@ def get(self, room_id):
     ('/register/(\w+)', RegisterPage),
     # TODO(jiayl): Remove support of /room/ when all clients are updated.
     ('/room/(\w+)', RoomPage),
+    ('/room/(\w+)/status', RoomStatusPage),
```
jiayliu: We already return the state of the room in /register request. Why do we need a new one?

samples/web/content/apprtc/js/main.js
```diff
@@ -358,3 +399,142 @@ function toggleFullScreen() {
...
+      }
+    }
+    
+  };
+}
+
+function loadRoom() {
+  pushRecentRoom(roomId).then(function() {
+    // check if room is available
+    var roomCheckUri = server + '/room/' + encodeURIComponent(roomId) + '/status';
```
jiayliu: The room state might change after the /room/status response, so I don't think the room status check here adds much value. Why not just handle the "room full error" in the response of /register/?

samples/web/content/apprtc/apprtc.py
```diff
@@ -546,5 +579,7 @@ def get(self, room_id):
...
     ('/r/(\w+)', RoomPage),
+    ('/params/(\w+)', ParamsPage),
```
jiayliu: I don't think this is needed. /register/ returns the same parameters.

samples/web/content/apprtc/apprtc.py
```diff
@@ -511,13 +518,18 @@ def post(self, room_id):
...
+    
   def get(self):
-    """Redirects to a room page."""
-    room_id = generate_random(8)
-    redirect = '/r/' + room_id
-    redirect = append_url_arguments(self.request, redirect)
-    self.redirect(redirect)
-    logging.info('Redirecting visitor to base URL to ' + redirect)
+    """Render index.html."""
+    params = {
```
jiayliu: I think this should generate a new room id and does waht RoomPage does.

samples/web/content/apprtc/manifest.json
```diff
@@ -0,0 +1,26 @@
...
+    "22": "images/apprtc-22.png",
+    "32": "images/apprtc-32.png",
+    "48": "images/apprtc-48.png",
+    "128": "images/apprtc-128.png"
+  },
+  "permissions": [
+    "audioCapture",
+    "videoCapture",
+    "storage",
+    "https://1-dot-apprtc.appspot.com/",
```
jiayliu: apprtc.appspot.com?

samples/web/content/apprtc/chrome-app-build/fetch-fonts.py
```diff
@@ -0,0 +1,48 @@
...
+
+import os
+import argparse
+import sys
+import re
+import urllib2
+from urlparse import urlparse
+from os.path import basename
+
+parser = argparse.ArgumentParser(description='Parse css file and retrieve all referenced font files')
```
jiayliu: Can we split the css font related code into a separate CL? It doesn't seem a priority to me for now.

juberti:
This is a pretty large PR, so I would recommend trying to split it into multiple smaller PRs - perhaps one for the basic app-specific stuff, one for the required changes to the existing app code, and one for the changes to add things like the MRU list.

samples/web/content/apprtc/js/main.js
chuckhays: Absolutely, I was hoping we were planning to do the work you already did. Is that going to be merged soon? I suspect I will have to redo most of what I did to fit in better with the refactored code.

samples/web/content/apprtc/chrome-app-build/fetch-fonts.py
chuckhays: Sure, @juberti suggested I break up the PR into smaller pieces below, so I can do the font pieces separately as well.

chuckhays:
@juberti Are we ok with partially completed work in master? I'm happy to break it up, but to make parts of it smaller, the chrome app might not work in the first PRs. Just curious on the team philosophy around master.

samples/web/content/apprtc/index.html
chuckhays: When you access /r/<roomid> directly, you get the same page as before with params already populated, and initialize() is called immediately.

samples/web/content/apprtc/js/util.js
chuckhays: Is this referring the the comment line?

jiayliu: Right, all comment should start with capitalized and end with '.'

On Wed, Jan 7, 2015 at 1:56 PM, Chuck Hays <notifications@github.com> wrote:

> In samples/web/content/apprtc/js/util.js
> <https://github.com/GoogleChrome/webrtc/pull/345#discussion-diff-22619522>
> :
>
> > +    result.push(charSet.charAt(Math.floor(Math.random() * charSet.length)));
> > +  }
> > +  return result.join('');
> > +}
> > +
> > +// Returns true if the code is running in a packaged Chrome App.
> > +function isChromeApp() {
> > +  return (typeof chrome !== 'undefined' &&
> > +          typeof chrome.storage !== 'undefined' &&
> > +          typeof chrome.storage.local !== 'undefined');
> > +}
> > +
> > +// Get a value from local browser storage. Calls callback with value.
> > +function getStorage(key, callback) {
> > +  if (isChromeApp()) {
> > +    // use chrome.storage.local
>
> Is this referring the the comment line?
>
> —
> <https://github.com/GoogleChrome/webrtc/pull/345/files#r22619522>.
>



-- 
Jiayang Liu

samples/web/content/apprtc/apprtc.py
chuckhays: We probably don't, was concerned about side effects of calling /register when I added it, and if we wanted to be able to check the status without actually joining the room, but even if we did that, the room might become full before we try to register, so doing it in a single operation makes more sense.

juberti:
Generally, we expect master to only be blessed code. If we want to have a fully functional version, we can do it on a branch and pull it over piece by piece to master.

chuckhays:
This was split up into separate PRs.

PR #347, opened 2015-01-07
Update extension description to indicate its use with peer2peer test app...

tednakamura:
... instead of generic "my app"


KaptenJansson:
The extension was added for general use, not only peer2peer but at the moment it seems to be the only app in this github repo using so it's probably okay.

KaptenJansson:
Ted, do you agree with my comment? If so please close this ;)

tednakamura:
Ah, sorry I missed your replies. Keeping the existing text as-is is fine with me.

PR #348, opened 2015-01-07
Update capture extension instructions to reload over https.

tednakamura:
This is necessary because the extension seems to fail if the page is loaded over http.


KaptenJansson:
LGTM

PR #350, opened 2015-01-08
Logging peer connection config for loopback calls.

andresusanopinto:
This will allow to see which turn server got selected. (At the moment suspecting that when hitting a bad turn server strange delays can be seen).

### DIFF
--- a/samples/web/content/testrtc/js/call.js
+++ b/samples/web/content/testrtc/js/call.js
@@ -9,7 +9,7 @@
 
 function Call(config) {
   this.traceEvent = report.traceEventAsync('call');
-  this.traceEvent({ config : config });
+  this.traceEvent({ config: config });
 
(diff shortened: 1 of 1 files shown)


samples/web/content/testrtc/js/call.js
```diff
@@ -8,6 +8,9 @@
 'use strict';
 
 function Call(config) {
+  this.traceEvent = report.traceEventAsync('call');
+  this.traceEvent({ config : config });
```
KaptenJansson: remove space before : (should be config: config).

KaptenJansson:
LGTM with nit.

samples/web/content/testrtc/js/call.js
andresusanopinto: done.

PR #351, opened 2015-01-09
TestRTC: Uniformize bower dep format.

KaptenJansson:


samples/web/content/testrtc/bower.json
```diff
@@ -23,14 +23,14 @@
...
-    "webcomponentsjs": "Polymer/webcomponentsjs#^0.5.0",
-    "core-toolbar": "Polymer/core-toolbar#~0.5.0",
-    "core-icon-button": "Polymer/core-icon-button#~0.5.0",
-    "core-icon": "Polymer/core-icon#~0.5.0",
-    "core-collapse": "Polymer/core-collapse#~0.5.0",
-    "paper-dialog": "Polymer/paper-dialog#~0.5.0",
-    "paper-progress": "Polymer/paper-progress#~0.5.0",
-    "paper-icon-button": "Polymer/paper-icon-button#~0.5.0",
-    "paper-button": "Polymer/paper-button#~0.5.0"
+    "webcomponentsjs": "Polymer/webcomponentsjs#^0.5.2",
```
KaptenJansson: I did not even realize that they were different, by default when using -save-dependencies during bower install it uses the tilde version. To why it uses caret for webcomponents it might be due to it being deemed as critical. Thank for pointing this out. Added some info about this in the developers guide.

andresusanopinto:
Will merge since this makes the dep format uniform.
There is no clear need to actually update to 0.5.2 from 0.5.0... but since we will be testing this in IE and Safari lets update it.

PR #352, opened 2015-01-09
Add webaudio shim and fail condition to mic test

KaptenJansson:
I thought it would make sense to check if web audio is present in the specific test in order to use reportFatal to disable the test.

### DIFF
--- a/samples/web/content/testrtc/js/mictest.js
+++ b/samples/web/content/testrtc/js/mictest.js
@@ -31,7 +31,11 @@ function MicTest() {
 
 MicTest.prototype = {
   run: function() {
-    doGetUserMedia(this.constraints, this.gotStream.bind(this));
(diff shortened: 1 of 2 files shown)


samples/web/content/testrtc/js/mictest.js
```diff
@@ -31,7 +31,11 @@ function MicTest() {
 
 MicTest.prototype = {
   run: function() {
-    doGetUserMedia(this.constraints, this.gotStream.bind(this));
+    if (typeof audioContext === 'undefined') {
+      reportFatal('WebAudio is not supported, test cannot run.');
```
andresusanopinto: Can you convert the reportFatal into "reportError" and fix the testFinished() code paths?

andresusanopinto:
Thanks!

PR #353, opened 2015-01-09
WiFi periodic scan test now runs with both svrrflx and relay candidates.

andresusanopinto:
Test with only turn candidates also left in for further investigation.
Added check for min number of samples as if the page is not on the foreground the JS setTimeout only triggers once a second.

### DIFF
--- a/samples/web/content/testrtc/js/bandwidth_test.js
+++ b/samples/web/content/testrtc/js/bandwidth_test.js
@@ -215,7 +215,7 @@ function testForWiFiPeriodicScan(candidateFilter, config) {
     reportInfo('Min delay: ' + min + ' ms.');
     reportInfo('Max delay: ' + max + ' ms.');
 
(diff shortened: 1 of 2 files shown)


samples/web/content/testrtc/js/bandwidth_test.js
```diff
@@ -208,15 +211,18 @@ function testForWiFiPeriodicScan(config) {
...
     var max = arrayMax(delays);
     var min = arrayMin(delays);
-
     reportInfo('Average delay: ' + avg + ' ms.');
     reportInfo('Min delay: ' + min + ' ms.');
     reportInfo('Max delay: ' + max + ' ms.');
 
-    if (max / (min + 100) < 2.0) {
-      reportSuccess('All seems fine.');
+    if (delays.lenght < 0.8 * testDurationMs / sendIntervalMs) {
```
KaptenJansson: should be delays.length

KaptenJansson:
LGTM with nit

Issue #297, opened 2014-12-14
Make adapter.js a modern JS library (module systems aware)


ibc:
No one in the world should define a **global** function called `trace()` (and so many other `window` globals defined in `adapter.js`).

`adapter.js` seems to ignore any new design pattern and JS module system (AMD, CommonJS, etc).

Would you accept a PR removing all the global stuff and providing a library suitable to be loaded by any existing JS module loader without polluting the `window` global namespace?

ibc:
Hi, finally I've taken a different approach. It covers more than what adapter.js provides as it is a full wrapper over the RTCPeerConnection interface (along with all the other WebRTC interfaces and functions). It is loadable via Node+browserify, AMD, RequireJS, etc, and it does not pollute the global (window) namespace, even stuff like `RTCPeerConnection`, etc. Instead it provides a single entry module called "rtcninja":

https://github.com/ibc/rtcninja.js

The documentation is still missing however.

juberti:
Thanks. 2KLOC seems kind of heavy though for this shim lib. I think we probably want to have something more like willscott's webrtc-adapter.

Perhaps the documentation will make the value more clear, but I was also somewhat wary of the various timers I saw in your implementation.

juberti:
Migrated to other repo

PR #355, opened 2015-01-10
allow adapter.js to be required

fippo:
I'm aware of the discussion in #297 but the pragmatic solution is to just allow it to be require'd.
Anyone who wants to use adapter.js as a module can do it, you don't need to change anything for now, don't need to use browserify, etc.

The console.log output is a bit icky for a module though.


juberti:
Build failed...

fippo:
Fixed, jshint is a little picky sometimes.

juberti:
Can you resend this to the new repo at webrtc/adapter?

fippo:
only if you promise not to ignore it until it runs into a merge conflict ;-)
(will do)

PR #356, opened 2015-01-10
Fix misuse of class attributes for Client and Room

jiayliu:


juberti:
yep, this isn't Java :)

PR #357, opened 2015-01-12
Don't display distracting errors when WSS comm fails.

phoglund:
When there's an error communicating with the wss server, AppRTC js spams the js log with

Uncaught TypeError: Cannot read property 'toPrecision' of undefined
Uncaught TypeError: Cannot read property 'toPrecision' of undefined
Uncaught TypeError: Cannot read property 'toPrecision' of undefined

The reason is that the formatting logic does not deal when the stats values are null, like when the call fails to get up. The below is a suggestion to fix that, but I'm sure there's a better way (maybe not request stats until we know wss is ok)?


juberti:
LGTM

This seems fine for now. In the future we might want to not display any stats if endTime is null (i.e. the call isn't connected), but I think this change is a good idea in any case.

PR #359, opened 2015-01-12
Chrome app build script

chuckhays:
Sets up a grunt build script to build the chrome app from the same html file as the web app.

To build:
grunt build

To load app:
chrome://extensions, load unpackaged app, find the build/chrome-app directory.

The build steps are:
1. Delete the old build dir
2. Create the build dir
3. Copy source files (does not copy .html file)
4. Use jinja to transform index.html to appwindow.html in build dir. 
     Note: This applies index.json to index.html. Ideally index.json would be named something else, but it would require changing the grunt-jinja library. We may want to do this at some point for clarity.
5. Create a zip file from the build directory
6. Package crx file - this is not complete, as there are question we need to answer, such as:
   a. Where do we get/store the key file?
   b. Should we use the chrome executable to package the app or use one of the implementations internally, such as the keep app (more complex but has many features)

@jiayliu @juberti


Gruntfile.js
```diff
@@ -84,6 +83,32 @@ module.exports = function(grunt) {
...
       },
     },
+    
+    'grunt-chrome-build' : {
+      apprtc: {
+        options: {
+          buildDir: 'build/chrome-app',
+          zipFile: 'build/chrome-app/apprtc.zip',
+          chromeBinary: '/Applications/Google Chrome.app/Contents/MacOS/Google Chrome',
+          keyFile: 'apprtc.pem'
```
chuckhays: chromeBinary/keyFile are not currently used, they will be part of step 6 once we decide the right way to implement it.

jiayliu: I think we'll store the .pem file internally. I'd leave comments here to say that the users of the demo app should fill in the chromeBinary and keyFile path.

grunt-chrome-build/grunt-chrome-build.js
```diff
@@ -0,0 +1,103 @@
...
+          expand: true, 
+          cwd: buildDir,  
+          src: ['**/*']
+        }
+      ]
+    });
+    
+    grunt.task.run('compress:chrome-build');
+    
+    // This section does not work yet.
```
jiayliu: Can we do this step conditionally, i.e. if chromeBinary and keyFile are provided, create crx file; otherwise, skip it.

chuckhays:
Closing this in favor of a new PR (https://github.com/GoogleChrome/webrtc/pull/378)

PR #364, opened 2015-01-13
Changed Import to import

### LINKED ISSUE #344: TestRTC - Internet Explorer and Safari issues with polymer and AudioContext
It executes functions in the test case javascript files before the global variables in main.js are defined (Error SCRIPT5007 when addTest() in mictest.js (i.e. first testcase file) when trying to access length method on the global variable testSuites in main.js).

Found this while looking at providing some error message on non WebRTC browsers like internet explorer 10, safari, Chrome on iOS etc. Issue #342 brought this to my attention.

Either we fix this now by adding IE exception handling or when IE supports WebRTC alongside the fix (check if browser is WebRTC compatible else show error message to user instead of a blank page) for #342.

KaptenJansson:
Safari and internet explorer are more strict hence Import with capitol "i" causes polymer never to finish importing. Firefox and Chrome assumes lowercase and it works fine.

Partly Fixes #344.

### DIFF
--- a/samples/web/content/testrtc/index.html
+++ b/samples/web/content/testrtc/index.html
@@ -23,7 +23,7 @@
   <script src="bower_components/webcomponentsjs/webcomponents.js"></script>
   <link rel="import" href="bower_components/core-toolbar/core-toolbar.html">
   <link rel="import" href="bower_components/core-icon/core-icon.html">
-  <link rel="Import" href="bower_components/core-icons/av-icons.html">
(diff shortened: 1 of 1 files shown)


KaptenJansson:
IMHO grunt should have picked this up in addition to the browser, IE and Safari should complain when it cannot parse import links (any data) properly.

KaptenJansson:
Forgot to mention that this was figured out thanks to Andrè!

PR #366, opened 2015-01-13
Add check for webrtc components in gumhandler.js

### LINKED ISSUE #342: TestRTC to show useful message when WebRTC is not supported
Try loading test-rtc.appspot.com on Chrome for iOS.

Result: blank page.

KaptenJansson:
Added check for navigator.getusermedia if not supported an error is present to the user.
Changed the startbutton to be disabled by default in HTML.

Fixes #342.

Note: On IE , Safari and Firefox the error (all paper-dialogs essentially) dialog is presented in the middle of the screen while on Chrome it's presented at the top.

### DIFF
--- a/samples/web/content/testrtc/js/gumhandler.js
+++ b/samples/web/content/testrtc/js/gumhandler.js
@@ -16,11 +16,12 @@ function GumHandler() {
   this.gumNotSupportedDialog_ = document.getElementById('gum-not-supported-dialog');
   this.gumErrorMessage_ = document.getElementById('gum-error-message');
   this.firstUserCheck_ = null;
+  this.gumStreamSuccessCallback_ = null;
 }
 
 GumHandler.prototype = {
   start: function(callback) {
-    this.gumStreamSuccessCallback = callback;
+    this.gumStreamSuccessCallback_ = callback;
     if (typeof navigator.getUserMedia === 'undefined') {
       this.gumNotSupportedDialog_.open();
     } else {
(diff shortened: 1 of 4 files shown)


KaptenJansson:
Should we add a typeof check on MediaStreamTrack.getSources as well even if it's shimmed in adapter.js?

samples/web/content/testrtc/js/gumhandler.js
```diff
@@ -13,15 +13,20 @@
...
   this.firstUserCheck_ = null;
 }
 
 GumHandler.prototype = {
-  start: function() {
-    startButton.disabled = true;
-    this.getUserMedia_();
-    this.firstUserCheck_ = setTimeout(this.firstTimeUser_.bind(this), 300);
+  start: function(callback) {
+    this.gumStreamSuccessCallback = callback;
```
andresusanopinto: call it gumStreamSuccessCallback_ and set it to null on the constructor.

andresusanopinto:
What does shimed means? Does it means it is always there?
Or if the browser supports it in some other location it will show there?

If the later then we should check it.

samples/web/content/testrtc/js/gumhandler.js
```diff
@@ -42,7 +47,7 @@ GumHandler.prototype = {
     }
     this.gumPendingDialog_.close();
     this.gumErrorDialog_.close();
-    startButton.removeAttribute('disabled');
+    this.gumStreamSuccessCallback();
```
andresusanopinto: suffix with _ and set it to null on the constructor.. as that allows to easily see all members used by this class.

KaptenJansson: Done.

KaptenJansson:
I added a check for if MediaStreamTrack.getSources exists like we do for navigator.getUserMedia.

PR #368, opened 2015-01-13
Add input and sent FPS to camresolutiontest.js

### LINKED ISSUE #365: TestRTC to to show avg fps delivery when running supported resolutions tests
Magnus was testing some laptops and having that value printed would have be useful.

KaptenJansson:
Added googFrameRateInput and googFrameRateSent stats.
Increased the gatherStats interval to 1000ms to make sure getStats does not affect the performance.

Fixes #365 

Note:
Mac camera teardown is unpredictable and flaky on Chrome hence this test will most likely fail on Mac's. Hopefully it will be fixed soon. In the meantime, as a workaround you can have an apprtc tab open with the camera opened in HD while running TestRTC. This is not an elegant solution and it might even not give accurate stats but we do not want to create a workaround for a Chrome bug on TestRTC.

### DIFF
--- a/samples/web/content/testrtc/js/camresolutionstest.js
+++ b/samples/web/content/testrtc/js/camresolutionstest.js
@@ -91,7 +91,7 @@ CamResolutionsTest.prototype = {
     var call = new Call();
     call.pc1.addStream(stream);
     call.establishConnection();
-    call.gatherStats(call.pc1, this.analyzeStats_.bind(this), 100);
+    call.gatherStats(call.pc1, this.analyzeStats_.bind(this), 1000);
     setTimeoutWithProgressBar( function() {
       call.close();
       stream.getVideoTracks()[0].stop();
@@ -101,6 +101,8 @@ CamResolutionsTest.prototype = {
   analyzeStats_: function(stats) {
     var currentRes = this.currentResolutionForCheckEncodeTime;
     var googAvgEncodeTime = [];
+    var googAvgFrameRateInput = [];
+    var googAvgFrameRateSent = [];
 
     for (var index = 0; index < stats.length - 1; index++) {
       if (stats[index].type === 'ssrc') {
@@ -108,6 +110,8 @@ CamResolutionsTest.prototype = {
         // TODO(jansson) expand to cover audio as well.
         if (stats[index].stat('googFrameRateInput') > 0) {
           googAvgEncodeTime.push(parseInt(stats[index].stat('googAvgEncodeMs')));
+          googAvgFrameRateInput.push(parseInt(stats[index].stat('googFrameRateInput')));
+          googAvgFrameRateSent.push(parseInt(stats[index].stat('googFrameRateSent')));
         }
       }
     }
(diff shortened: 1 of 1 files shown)


KaptenJansson:
André, PTAL.

Regarding why the input fps is lower than sent fps initially might be down to how webrtc compute the stats, Magnus is going to look into that.

Also he suggested that we might actually start the timer when getStats is delivering stats indicating there is real video rather than having just a timer set at 5 sec for the whole test, WDYT?

samples/web/content/testrtc/js/camresolutionstest.js
```diff
@@ -115,16 +119,30 @@ CamResolutionsTest.prototype = {
...
+                                               minFPSSent: minFPSSent,
+                                               maxFPSSent: maxFPSSent,
+                                               avgFPSSent: avgFPSSent });
 
     if (googAvgEncodeTime.length === 0) {
       reportError('No stats collected. Check your camera.');
     } else {
       reportInfo('Encode time (ms): ' + minEncodeMs + ' min / ' + avgEncodeMs + ' avg / ' + maxEncodeMs + ' max');
+      reportInfo('Input FPS: ' + minFPSInput + ' min / ' + avgFPSInput + ' avg / ' + maxFPSInput + ' max');
+      reportInfo('Sent FPS: ' + minFPSSent + ' min / ' + avgFPSSent + ' avg / ' + maxFPSSent + ' max');
```
andresusanopinto: do you wanna add a check here check for:
 if (avgFPSSent < 5) { reportError('Very low framerate.'); } ?

samples/web/content/testrtc/js/camresolutionstest.js
```diff
@@ -91,7 +91,7 @@ CamResolutionsTest.prototype = {
     var call = new Call();
     call.pc1.addStream(stream);
     call.establishConnection();
-    call.gatherStats(call.pc1, this.analyzeStats_.bind(this), 100);
+    call.gatherStats(call.pc1, this.analyzeStats_.bind(this), 1000);
```
andresusanopinto: Btw I was thinking that due to timing one will likely only gather 4 samples during the 5 seconds... but I guess that is fine for now.

KaptenJansson: I have, but increasing the frequency seems to correlate to higher failure rate on Mac but since it's already bad anyhow maybe it's bad to sacrifice sample size on other platforms. Is 100ms to fast, maybe 300 or 500?

samples/web/content/testrtc/js/camresolutionstest.js
KaptenJansson: Sure!

KaptenJansson: Done!

samples/web/content/testrtc/js/camresolutionstest.js
KaptenJansson: Left it as is for now.

PR #380, opened 2015-01-14
Update the room state string after new client is added

jiayliu:


samples/web/content/apprtc/apprtc.py
```diff
@@ -369,6 +369,7 @@ def add_client_to_room(host, room_id, client_id, is_loopback):
       room.add_client(client_id, Client(is_initiator))
       other_client.clear_messages()
 
+    room_state = str(room)
```
juberti: room_state isn't used anywhere except the return. Should we just set it there?

(also, why wasn't this caught by unit tests?)

jiayliu:
Done.
The unit test does not catch it because it does not affect request response, but only logging.

PR #381, opened 2015-01-15
Use params returned from /register call

chuckhays:
Instead of using a params array from the initial html page, use the params array as returned by the /register/ call.

This will enable the Chrome app to get all of its parameters from the server instead of hard coding defaults, without adding any additional endpoints.

The most significant piece is that it required re-ordering the sequence of promises in call setup, because we needed the call to /register to happen first, so we had a params array to use in other calls. I did take into account the change today by @jiayliu to the ordering to maintain that change.

This step is in preparation for the landing page, and joining a room without requiring navigating to another page. When the browser requests /, the loadingParams can specify connect: false, and we will show the room selection UI. After they select a room, call setup proceeds as before. If the browser navigates directly to /r/#, loadingParams can specify connect: true and the call will be initiated immediately.

@jiayliu @juberti

### DIFF
--- a/samples/web/content/apprtc/apprtc.py
+++ b/samples/web/content/apprtc/apprtc.py
@@ -245,11 +245,6 @@ def get_room_parameters(request, room_id, client_id, is_initiator):
   username = client_id if client_id is not None else generate_random(9)
   if len(turn_base_url) > 0:
     turn_url = constants.TURN_URL_TEMPLATE % (turn_base_url, username, constants.CEOD_KEY)
-
-  room_link = None
-  if room_id is not None:
-    room_link = request.host_url + '/r/' + room_id
-    room_link = append_url_arguments(request, room_link)
     
   pc_config = make_pc_config(ice_transports)
   pc_constraints = make_pc_constraints(dtls, dscp, ipv6)
@@ -282,7 +277,10 @@ def get_room_parameters(request, room_id, client_id, is_initiator):
     'wss_url': wss_url,
     'wss_post_url': wss_post_url
   }
+
   if room_id is not None:
+    room_link = request.host_url + '/r/' + room_id
+    room_link = append_url_arguments(request, room_link)
     params['room_id'] = room_id
     params['room_link'] = room_link
   if client_id is not None:
--- a/samples/web/content/apprtc/apprtc.py
+++ b/samples/web/content/apprtc/apprtc.py
@@ -539,7 +539,11 @@ class RoomPage(webapp2.RequestHandler):
         self.write_response('full.html')
         return
     # Parse out room parameters from request.
-    params = get_room_parameters(self.request, room_id, None, None)
+    roomParams = get_room_parameters(self.request, room_id, None, None)
+    params = {
+      'error_messages': roomParams['error_messages'],
+      'roomId': room_id
+    }
     self.write_response('index.html', params)
 
 app = webapp2.WSGIApplication([
--- a/samples/web/content/apprtc/apprtc.py
+++ b/samples/web/content/apprtc/apprtc.py
@@ -246,8 +246,11 @@ def get_room_parameters(request, room_id, client_id, is_initiator):
   if len(turn_base_url) > 0:
     turn_url = constants.TURN_URL_TEMPLATE % (turn_base_url, username, constants.CEOD_KEY)
 
-  room_link = request.host_url + '/room/' + room_id
-  room_link = append_url_arguments(request, room_link)
+  room_link = None
+  if room_id is not None:
+    room_link = request.host_url + '/room/' + room_id
+    room_link = append_url_arguments(request, room_link)
+    
   pc_config = make_pc_config(ice_transports)
   pc_constraints = make_pc_constraints(dtls, dscp, ipv6)
   offer_constraints = { 'mandatory': {}, 'optional': [] }
@@ -257,8 +260,6 @@ def get_room_parameters(request, room_id, client_id, is_initiator):
   params = {
     'error_messages': error_messages,
     'is_loopback' : json.dumps(debug == 'loopback'),
-    'room_id': room_id,
-    'room_link': room_link,
     'pc_config': json.dumps(pc_config),
     'pc_constraints': json.dumps(pc_constraints),
     'offer_constraints': json.dumps(offer_constraints),
@@ -281,6 +282,9 @@ def get_room_parameters(request, room_id, client_id, is_initiator):
     'wss_url': wss_url,
     'wss_post_url': wss_post_url
   }
+  if room_id is not None:
+    params['room_id'] = room_id
+    params['room_link'] = room_link
   if client_id is not None:
     params['client_id'] = client_id
   if is_initiator is not None:
@@ -539,12 +543,15 @@ class RoomPage(webapp2.RequestHandler):
         self.write_response('full.html')
         return
     # Parse out room parameters from request.
-    roomParams = get_room_parameters(self.request, room_id, None, None)
-    params = {
-      'error_messages': roomParams['error_messages'],
-      'roomId': room_id
-    }
+    params = get_room_parameters(self.request, room_id, None, None)
     self.write_response('index.html', params)
+
+class ParamsPage(webapp2.RequestHandler):
+  def get(self):
+    # Return room independent room parameters.
+    params = get_room_parameters(self.request, None, None, None)
+    self.response.write(json.dumps(params))
+    
 
 app = webapp2.WSGIApplication([
     ('/', MainPage),
@@ -554,4 +561,5 @@ app = webapp2.WSGIApplication([
     # TODO(jiayl): Remove support of /room/ when all clients are updated.
     ('/room/(\w+)', RoomPage),
     ('/r/(\w+)', RoomPage),
+    ('/params', ParamsPage),
 ], debug=True)
--- a/samples/web/content/apprtc/apprtc.py
+++ b/samples/web/content/apprtc/apprtc.py
@@ -248,7 +248,7 @@ def get_room_parameters(request, room_id, client_id, is_initiator):
 
   room_link = None
   if room_id is not None:
-    room_link = request.host_url + '/room/' + room_id
+    room_link = request.host_url + '/r/' + room_id
     room_link = append_url_arguments(request, room_link)
     
   pc_config = make_pc_config(ice_transports)
--- a/samples/web/content/apprtc/index.html
+++ b/samples/web/content/apprtc/index.html
@@ -103,7 +103,32 @@
   <script type="text/javascript">
     var loadingParams = {
       errorMessages: {{ error_messages }},
-      roomId: '{{ roomId }}'
+      isLoopback: {{ is_loopback }},
+
+      roomId: '{{ room_id }}',
+      roomLink: '{{ room_link }}',
+
+      mediaConstraints: {{ media_constraints | safe }},
+      offerConstraints: {{ offer_constraints | safe }},
+      peerConnectionConfig: {{ pc_config | safe }},
+      peerConnectionConstraints: {{ pc_constraints | safe }},
+      turnRequestUrl: '{{ turn_url }}',
+      turnTransports: '{{ turn_transports }}',
+
+      audioSendBitrate: '{{ asbr }}',
+      audioSendCodec: '{{ audio_send_codec }}',
+      audioRecvBitrate: '{{ arbr }}',
+      audioRecvCodec: '{{ audio_receive_codec }}',
+      opusMaxPbr: '{{ opusmaxpbr }}',
+      opusFec: '{{ opusfec }}',
+      opusStereo: '{{ stereo }}',
+      videoSendBitrate: '{{ vsbr }}',
+      videoSendInitialBitrate: '{{ vsibr }}',
+      videoSendCodec: '{{ video_send_codec }}',
+      videoRecvBitrate: '{{ vrbr }}',
+      videoRecvCodec: '{{ video_receive_codec }}',
+      wssUrl: '{{ wss_url }}',
+      wssPostUrl: '{{ wss_post_url }}'
     };
 
     var appController;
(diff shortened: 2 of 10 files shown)


samples/web/content/apprtc/js/signalingchannel_test.js
```diff
@@ -98,7 +98,7 @@ SignalingChannelTest.prototype.setUp = function() {
   WebSocket = MockWebSocket;
 
   this.channel =
-      new SignalingChannel(FAKE_WSS_URL, FAKE_WSS_POST_URL);
+      new SignalingChannel();
```
jiayliu: Remove the line break

samples/web/content/apprtc/js/call.js
```diff
@@ -145,56 +142,83 @@ Call.prototype.toggleAudioMute = function() {
...
-        // TODO(tkchin): clean up response format. JSHint doesn't like it.
-        /* jshint ignore:start */
-        this.params_.clientId = roomParams.client_id;
-        this.params_.roomId = roomParams.room_id;
-        this.params_.isInitiator = roomParams.is_initiator === 'true';
-        /* jshint ignore:end */
-        this.params_.messages = roomParams.messages;
+  }.bind(this)).then(function() {
+    var mediaPromise = this.maybeGetMedia_();
+    var turnPromise = this.maybeGetTurnServers_();
```
jiayliu: Requesting media and TURN are slow, so we want to start them as soon as possible instead of waiting for 'register'. 
We could/should create the media constraints locally from the URL params, instead of getting them from the server.
But for TURN, we have to get the default turn url from the initial page load params.

samples/web/content/apprtc/index.html
```diff
@@ -101,34 +101,9 @@
...
-      opusFec: '{{ opusfec }}',
-      opusStereo: '{{ stereo }}',
-      videoSendBitrate: '{{ vsbr }}',
-      videoSendInitialBitrate: '{{ vsibr }}',
-      videoSendCodec: '{{ video_send_codec }}',
-      videoRecvBitrate: '{{ vrbr }}',
-      videoRecvCodec: '{{ video_receive_codec }}',
-      wssUrl: '{{ wss_url }}',
-      wssPostUrl: '{{ wss_post_url }}'
+      roomId: '{{ roomId }}'
```
jiayliu: This is a nice cleanup, although we need to put back the TURN url and wss urls to start TURN/WSS as soon as possible instead of waiting for 'register'.

samples/web/content/apprtc/js/call.js
chuckhays: What should we do for the chrome app? Just pick defaults? Part of the reason for this change was the desire to get all the parameters from the server for the Chrome app instead of having hard coded values. Ideas?

jiayliu: I'm fine with adding a new request handler, e.g. GET /params for TURN url
and WSS urls.
We probably want to make mobile clients follow this pattern too.

On Wed, Jan 14, 2015 at 6:11 PM, Chuck Hays <notifications@github.com>
wrote:

> In samples/web/content/apprtc/js/call.js
> <https://github.com/GoogleChrome/webrtc/pull/381#discussion-diff-22986649>
> :
>
> > -
> > -  // Asynchronously register with GAE.
> > -  var registerPromise =
> > -      this.registerWithRoomServer_().then(function(roomParams) {
> > -        // The only difference in parameters should be clientId and isInitiator,
> > -        // and the turn servers that we requested.
> > -        // TODO(tkchin): clean up response format. JSHint doesn't like it.
> > -        /* jshint ignore:start */
> > -        this.params_.clientId = roomParams.client_id;
> > -        this.params_.roomId = roomParams.room_id;
> > -        this.params_.isInitiator = roomParams.is_initiator === 'true';
> > -        /* jshint ignore:end */
> > -        this.params_.messages = roomParams.messages;
> > +  }.bind(this)).then(function() {
> > +    var mediaPromise = this.maybeGetMedia_();
> > +    var turnPromise = this.maybeGetTurnServers_();
>
> What should we do for the chrome app? Just pick defaults? Part of the
> reason for this change was the desire to get all the parameters from the
> server for the Chrome app instead of having hard coded values. Ideas?
>
> —
> <https://github.com/GoogleChrome/webrtc/pull/381/files#r22986649>.
>



-- 
Jiayang Liu

chuckhays: Do you think we should include the parameters required to request the media as well? We can then have the 2 slow calls happen as early as possible.

We can also do the feature @juberti suggested and put the users's video under the room selection UI.

jiayliu: The media constraints have nothing depending on the server actually, and
can be created completely on the client side.
I can show you where the server logic is to move it to client tomorrow.

On Wed, Jan 14, 2015 at 6:27 PM, Chuck Hays <notifications@github.com>
wrote:

> In samples/web/content/apprtc/js/call.js
> <https://github.com/GoogleChrome/webrtc/pull/381#discussion-diff-22987154>
> :
>
> > -
> > -  // Asynchronously register with GAE.
> > -  var registerPromise =
> > -      this.registerWithRoomServer_().then(function(roomParams) {
> > -        // The only difference in parameters should be clientId and isInitiator,
> > -        // and the turn servers that we requested.
> > -        // TODO(tkchin): clean up response format. JSHint doesn't like it.
> > -        /* jshint ignore:start */
> > -        this.params_.clientId = roomParams.client_id;
> > -        this.params_.roomId = roomParams.room_id;
> > -        this.params_.isInitiator = roomParams.is_initiator === 'true';
> > -        /* jshint ignore:end */
> > -        this.params_.messages = roomParams.messages;
> > +  }.bind(this)).then(function() {
> > +    var mediaPromise = this.maybeGetMedia_();
> > +    var turnPromise = this.maybeGetTurnServers_();
>
> Do you think we should include the parameters required to request the
> media as well? We can then have the 2 slow calls happen as early as
> possible.
>
> We can also do the feature @juberti <https://github.com/juberti>
> suggested and put the users's video under the room selection UI.
>
> —
> <https://github.com/GoogleChrome/webrtc/pull/381/files#r22987154>.
>



-- 
Jiayang Liu

juberti: Loading the parameters that are independent of the room (i.e. everything except initiator/clientid/room) should occur when the app starts up. For the non-app case, we should return those parameters in the pageload. Then, /register can just return the room-dependent params.

chuckhays: OK, that is mostly how I implemented it in the original PR, I'll change it back to that way of doing things.

chuckhays:
Updated to do the following:
* Web app gets params in the /r/# page load as before
* Chrome app calls /params during initialization to get params
* Call setup ordering restored to original

samples/web/content/apprtc/apprtc.py
```diff
@@ -246,8 +246,11 @@ def get_room_parameters(request, room_id, client_id, is_initiator):
   if len(turn_base_url) > 0:
     turn_url = constants.TURN_URL_TEMPLATE % (turn_base_url, username, constants.CEOD_KEY)
 
-  room_link = request.host_url + '/room/' + room_id
-  room_link = append_url_arguments(request, room_link)
+  room_link = None
+  if room_id is not None:
+    room_link = request.host_url + '/room/' + room_id
```
jiayliu: use '/r/' instead.

samples/web/content/apprtc/js/appcontroller.js
```diff
@@ -64,45 +66,65 @@ var AppController = function(params) {
...
-  if (roomErrors.length > 0) {
-    for (var i = 0; i < roomErrors.length; ++i) {
-      this.infoBox_.pushErrorMessage(roomErrors[i]);
-    }
-    return;
+  this.loadingParams_ = loadingParams;
+  var paramsPromise = Promise.resolve(JSON.stringify({}));
+  if (this.loadingParams_.paramsFunction)
+  {
+    // Fetch params from server.
```
jiayliu: Please expand the comment to make it clear that it's needed by the chrome app.

samples/web/content/apprtc/js/appcontroller.js
```diff
@@ -64,45 +66,65 @@ var AppController = function(params) {
...
-  this.call_.onlocalstreamadded = this.onLocalStreamAdded_.bind(this);
-
-  this.call_.onsignalingstatechange =
-      this.infoBox_.updateInfoDiv.bind(this.infoBox_);
-  this.call_.oniceconnectionstatechange =
-      this.infoBox_.updateInfoDiv.bind(this.infoBox_);
-  this.call_.onnewicecandidate =
-      this.infoBox_.recordIceCandidateTypes.bind(this.infoBox_);
+  Promise.resolve(paramsPromise).then(function(newParams) {
+    // Merge newly retrieved params with loadingParams
```
jiayliu: nit: comment should end with '.'

samples/web/content/apprtc/js/appcontroller.js
```diff
@@ -64,45 +66,65 @@ var AppController = function(params) {
...
+    if (newParams) {
+      Object.keys(newParams).forEach(function(key) {
+        this.loadingParams_[key] = newParams[key];
+      }.bind(this));
+    }
+  }.bind(this)).catch(function(error) {
+    trace('Initializing; error getting params from server: ' + error.message);
+  }.bind(this)).then(function() {
+    // Proceed with call set up.
```
jiayliu: I think it will be easier to read if you combine the two .then callbacks into a helper method.

samples/web/content/apprtc/js/appcontroller.js
```diff
@@ -64,45 +66,65 @@ var AppController = function(params) {
...
+    this.call_.oncallerstarted = this.displaySharingInfo_.bind(this);
+
+    this.call_.start(this.loadingParams_.roomId);
+
+    window.onbeforeunload = this.call_.hangup.bind(this.call_);
+    document.onkeypress = this.onKeyPress_.bind(this);
+    window.onmousemove = this.showIcons_.bind(this);
+  }.bind(this));
+  
   $(UI_CONSTANTS.muteAudioSvg).onclick = this.toggleAudioMute_.bind(this);
```
jiayliu: I think the rest of the method should be in the success callback of paramsFunction as well.

samples/web/content/apprtc/js/call.js
```diff
@@ -312,13 +313,13 @@ Call.prototype.startSignaling_ = function() {
 };
 
 // Registers with GAE and returns room parameters.
-Call.prototype.registerWithRoomServer_ = function() {
+Call.prototype.registerWithRoomServer_ = function(roomId) {
   return new Promise(function(resolve, reject) {
```
jiayliu: It seems unnecessary to have the roomId argument.

samples/web/content/apprtc/js/sdputils.js
```diff
@@ -8,7 +8,7 @@
 
 /* More information about these options at jshint.com/docs/options */
 
-/* globals displayError, params */
+/* globals displayError */
```
jiayliu: Not introduced by this PR, but there is no displayError any more. Could you change it to trace?

chuckhays:
Addressed your comments in latest commit.

samples/web/content/apprtc/apprtc.py
```diff
@@ -246,8 +246,11 @@ def get_room_parameters(request, room_id, client_id, is_initiator):
   if len(turn_base_url) > 0:
     turn_url = constants.TURN_URL_TEMPLATE % (turn_base_url, username, constants.CEOD_KEY)
 
-  room_link = request.host_url + '/room/' + room_id
-  room_link = append_url_arguments(request, room_link)
+  room_link = None
+  if room_id is not None:
+    room_link = request.host_url + '/r/' + room_id
+    room_link = append_url_arguments(request, room_link)
```
jiayliu: nit: move these before where they are used.

samples/web/content/apprtc/js/appwindow.js
```diff
@@ -8,63 +8,61 @@
...
+  errorMessages: [],
+  roomId: randomString(9),
+  roomServer: roomServer,
+  paramsFunction: function() {
+    return new Promise(function(resolve, reject) {
+      trace('Initializing; retrieving params from: ' + roomServer + '/params');
+      sendAsyncUrlRequest('GET', roomServer + '/params').then(function(result) {
+        var serverParams = parseJSON(result);
+        var newParams = {};
+        if (serverParams)
```
jiayliu: nit: exit early if serverParams is null to avoid extra indent.

samples/web/content/apprtc/js/call.js
```diff
@@ -45,8 +45,8 @@ Call.prototype.isInitiator = function() {
   return this.params_.isInitiator;
 };
 
-Call.prototype.start = function() {
-  this.connectToRoom_(this.maybeGetMedia_(), this.maybeGetTurnServers_());
+Call.prototype.start = function(roomId) {
+  this.connectToRoom_(roomId, this.maybeGetMedia_(), this.maybeGetTurnServers_());
```
jiayliu: connectToRoom_ takes two params and roomId seems unused.

chuckhays:
Updated. The connectToRoom_ method I mixed up with the subsequent change I've been working on/keeping up to date in another branch. It it fixed now how it will be used with the landing page.

samples/web/content/apprtc/js/appwindow.js
```diff
@@ -8,63 +8,64 @@
...
+  roomServer: roomServer,
+  paramsFunction: function() {
+    return new Promise(function(resolve, reject) {
+      trace('Initializing; retrieving params from: ' + roomServer + '/params');
+      sendAsyncUrlRequest('GET', roomServer + '/params').then(function(result) {
+        var serverParams = parseJSON(result);
+        var newParams = {};
+        if (!serverParams)
+        {
+          resolve(newParams);
```
jiayliu: Should return here

chuckhays: Instead of resolve? Wouldn't the promise we are returning never settle as a result?

jiayliu:
lgtm

chuckhays:
While testing loopback, I found another use of the global params. PTAL before I go ahead with this.

PR #382, opened 2015-01-15
Add a module for new GAE bind requests for ID registration

jiayliu:
CC: @tkchin @chuckhays


samples/web/content/apprtc/constants.py
```diff
@@ -20,4 +20,10 @@
 RESPONSE_UNKNOWN_ROOM = 'UNKNOWN_ROOM'
 RESPONSE_UNKNOWN_CLIENT = 'UNKNOWN_CLIENT'
 RESPONSE_DUPLICATE_CLIENT = 'DUPLICATE_CLIENT'
+
+RESPONSE_INVALID_ARGUMENT = 'ERR_INVALID_ARG'
```
juberti: We should be consistent in use of the ERR_ prefix. Either all errors have this, or none. I lean toward having the prefix.

samples/web/content/apprtc/gcm_register_test.py
```diff
@@ -0,0 +1,106 @@
...
+    self.assertEqual(constants.RESPONSE_SUCCESS,
+                     gcm_register.GCMRecord.verify(user_id, gcm_id, code))
+    self.assertEqual(constants.RESPONSE_NOT_FOUND,
+                     gcm_register.GCMRecord.verify(user_id, '1', code))
+    self.assertEqual(constants.RESPONSE_NOT_FOUND,
+                     gcm_register.GCMRecord.verify('1', gcm_id, code))
+    self.assertEqual(constants.RESPONSE_INVALID_STATE,
+                     gcm_register.GCMRecord.verify(user_id, gcm_id, '1'))
+
+  def testGCMRecrodRemove(self):
```
juberti: Record

samples/web/content/apprtc/gcm_register.py
```diff
@@ -0,0 +1,204 @@
...
+      for id in ids:
+        # TODO(jiayl): Only return the verified users when SMS verification is
+        # added.
+        if len(GCMRecord.get_by_user_id(id)) > 0:
+          result.append(id)
+      self.response.out.write(json.dumps(result))
+    else:
+      self.response.out.write(constants.RESPONSE_INVALID_ARG)
+
+  def get(self, cmd):
```
chuckhays: Is get() in this framework mapped directly to GET requests? It seems like new/update/del should be POST/PUT/DELETE, or at least POST. We could also use standard http responses for many of the cases. Thoughts?

jiayliu: I forgot to fix that before submit. Will fix.

jiayliu:
I'm going to move this into my own repository to avoid exposing it to the public apprtc. Will do a new PR.

PR #388, opened 2015-01-16
Manual-test: Change height to 1200px height in iframe-apprtc.html

KaptenJansson:
Currently the height is set to "90%" which does not work, need to specify in pixels instead.

### DIFF
--- a/samples/web/content/manual-test/iframe-apprtc/index.html
+++ b/samples/web/content/manual-test/iframe-apprtc/index.html
@@ -15,6 +15,6 @@
 </head>
 <body>
   AppRTC in an &lt;iframe&gt; element:<br>
-  <iframe width="90%" height="1200" src="http://apprtc.appspot.com"></iframe>
+  <iframe width="90%" height="1200px" src="http://apprtc.appspot.com"></iframe>
 </body>
 </html>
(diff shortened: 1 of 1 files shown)


KaptenJansson:
samdutton, PTAL.

samples/web/content/manual-test/iframe-apprtc/index.html
```diff
@@ -15,6 +15,6 @@
 </head>
 <body>
   AppRTC in an &lt;iframe&gt; element:<br>
-  <iframe width="90%" height="90%" src="http://apprtc.appspot.com"></iframe>
+  <iframe width="90%" height="1200px" src="http://apprtc.appspot.com"></iframe>
```
samdutton: Should probably be 1200, not 1200px (though 1200px will work – try using the Dev Tools to add a height attribute to [simpl.info/iframe](http://simpl.info/iframe)). 

Otherwise LGTM.

KaptenJansson: Actually height does not work on an iframe without the unit px.

samdutton: Hmm... That's odd (see [MDN](https://developer.mozilla.org/en/docs/Web/HTML/Element/iframe), etc.) 

Oh well – maybe percentage width with absolute height? 

Will merge.

PR #392, opened 2015-01-16
Add landing page to apprtc and chrome app.

chuckhays:
This adds a basic landing page to the web app and Chrome app, with UI similar to the existing landing page of the android app. 

We can use this as a starting point for building the new UI for the Chrome app.

@jiayliu @juberti

### DIFF
--- a/samples/web/content/apprtc/apprtc.py
+++ b/samples/web/content/apprtc/apprtc.py
@@ -515,13 +515,18 @@ class RegisterPage(webapp2.RequestHandler):
     logging.info('Room ' + room_id + ' has state ' + result['room_state'])
 
 class MainPage(webapp2.RequestHandler):
+  def write_response(self, target_page, params={}):
+    template = jinja_environment.get_template(target_page)
+    content = template.render(params)
+    self.response.out.write(content)
+    
   def get(self):
-    """Redirects to a room page."""
-    room_id = generate_random(8)
-    redirect = '/r/' + room_id
-    redirect = append_url_arguments(self.request, redirect)
-    self.redirect(redirect)
-    logging.info('Redirecting visitor to base URL to ' + redirect)
+    """Renders index.html."""
+    # Parse out parameters from request.
+    params = get_room_parameters(self.request, None, None, None)
+    params['connect'] = False
+    params['room_id'] = generate_random(8)
+    self.write_response('index.html', params)
 
 class RoomPage(webapp2.RequestHandler):
   def write_response(self, target_page, params={}):
@@ -542,6 +547,7 @@ class RoomPage(webapp2.RequestHandler):
         return
     # Parse out room parameters from request.
     params = get_room_parameters(self.request, room_id, None, None)
+    params['connect'] = True
     self.write_response('index.html', params)
 
 class ParamsPage(webapp2.RequestHandler):
--- a/samples/web/content/apprtc/apprtc.py
+++ b/samples/web/content/apprtc/apprtc.py
@@ -524,8 +524,8 @@ class MainPage(webapp2.RequestHandler):
     """Renders index.html."""
     # Parse out parameters from request.
     params = get_room_parameters(self.request, None, None, None)
-    params['connect'] = False
-    params['room_id'] = generate_random(8)
+    # room_id/room_link will not be included in the returned parameters
+    # so the client will show the landing page for room selection.
     self.write_response('index.html', params)
 
 class RoomPage(webapp2.RequestHandler):
@@ -547,7 +547,8 @@ class RoomPage(webapp2.RequestHandler):
         return
     # Parse out room parameters from request.
     params = get_room_parameters(self.request, room_id, None, None)
-    params['connect'] = True
+    # room_id/room_link will be included in the returned parameters
+    # so the client will launch the requested room.
     self.write_response('index.html', params)
 
 class ParamsPage(webapp2.RequestHandler):
@@ -555,7 +556,6 @@ class ParamsPage(webapp2.RequestHandler):
     # Return room independent room parameters.
     params = get_room_parameters(self.request, None, None, None)
     self.response.write(json.dumps(params))
-    
 
 app = webapp2.WSGIApplication([
     ('/', MainPage),
--- a/samples/web/content/apprtc/index.html
+++ b/samples/web/content/apprtc/index.html
@@ -38,6 +38,20 @@
     <video id="remote-video" autoplay></video>
     <video id="local-video" autoplay muted></video>
   </div>
+  
+  <div id="room-selection" class="hidden">
+     <h1>AppRTC</h1>
+     <p id="instructions">Please enter a room name. Room names are shared with everyone, so think of something unique and send it to a friend.</p>
+     <div>
+       <input type="text" id="room-id-input"/>
+       <button id="join-button">Join</button>
+       <button id="random-button">Random</button>
+     </div>
+     <div id="recent-rooms">
+       <p>Recently used rooms:</p>
+       <ul id="recent-rooms-list"></ul>
+     </div>
+   </div>
 
   <footer>
     <div id="sharing-div">
@@ -94,6 +108,7 @@
   <script src="/js/peerconnectionclient.js"></script>
   <script src="/js/call.js"></script>
   <script src="/js/infobox.js"></script>
+  <script src="/js/roomselection.js"></script>
   <script src="/js/appcontroller.js"></script>
 {% if chromeapp %}
   <script src="/js/appwindow.js"></script>
@@ -104,9 +119,14 @@
     var loadingParams = {
       errorMessages: {{ error_messages }},
       isLoopback: {{ is_loopback }},
-
+{% if connect %}
       roomId: '{{ room_id }}',
       roomLink: '{{ room_link }}',
+      connect: true,
+{% else %}
+      suggestedRoomId: '{{ room_id }}',
+      connect: false,
+{% endif %}
 
       mediaConstraints: {{ media_constraints | safe }},
       offerConstraints: {{ offer_constraints | safe }},
--- a/samples/web/content/apprtc/index.html
+++ b/samples/web/content/apprtc/index.html
@@ -40,18 +40,18 @@
   </div>
   
   <div id="room-selection" class="hidden">
-     <h1>AppRTC</h1>
-     <p id="instructions">Please enter a room name. Room names are shared with everyone, so think of something unique and send it to a friend.</p>
-     <div>
-       <input type="text" id="room-id-input"/>
-       <button id="join-button">Join</button>
-       <button id="random-button">Random</button>
-     </div>
-     <div id="recent-rooms">
-       <p>Recently used rooms:</p>
-       <ul id="recent-rooms-list"></ul>
-     </div>
-   </div>
+    <h1>AppRTC</h1>
+    <p id="instructions">Please enter a room name.</p>
+    <div>
+      <input type="text" id="room-id-input"/>
+      <button id="join-button">Join</button>
+      <button id="random-button">Random</button>
+    </div>
+    <div id="recent-rooms">
+      <p>Recently used rooms:</p>
+      <ul id="recent-rooms-list"></ul>
+    </div>
+  </div>
 
   <footer>
     <div id="sharing-div">
@@ -99,6 +99,7 @@
 
   <script src="/js/stats.js"></script>
   <script src="/js/util.js"></script>
+  <script src="/js/storage.js"></script>
   <script src="/js/adapter.js"></script>
 {% if not chromeapp %}
   {{ include_loopback_js }}
@@ -119,13 +120,9 @@
     var loadingParams = {
       errorMessages: {{ error_messages }},
       isLoopback: {{ is_loopback }},
-{% if connect %}
+{% if room_id %}
       roomId: '{{ room_id }}',
       roomLink: '{{ room_link }}',
-      connect: true,
-{% else %}
-      suggestedRoomId: '{{ room_id }}',
-      connect: false,
 {% endif %}
 
       mediaConstraints: {{ media_constraints | safe }},
--- a/samples/web/content/apprtc/js/appcontroller.js
+++ b/samples/web/content/apprtc/js/appcontroller.js
@@ -8,7 +8,8 @@
 
 /* More information about these options at jshint.com/docs/options */
 
-/* globals trace, InfoBox, setUpFullScreen, isFullScreen */
+/* globals trace, InfoBox, setUpFullScreen, isFullScreen,
+   RoomSelection, isChromeApp, chrome */
 /* exported AppController, remoteVideo */
 
 'use strict';
@@ -39,6 +40,11 @@ var UI_CONSTANTS = {
 
   remoteVideo: '#remote-video',
   roomLinkHref: '#room-link-href',
+  roomSelectionDiv: '#room-selection',
+  roomSelectionInput: '#room-id-input',
+  roomSelectionJoinButton: '#join-button',
+  roomSelectionRandomButton: '#random-button',
+  roomSelectionRecentList: '#recent-rooms-list',
   sharingDiv: '#sharing-div',
   statusDiv: '#status-div',
   videosDiv: '#videos',
@@ -58,6 +64,8 @@ var AppController = function(loadingParams) {
   this.remoteVideo_ = $(UI_CONSTANTS.remoteVideo);
   this.videosDiv_ = $(UI_CONSTANTS.videosDiv);
   this.roomLinkHref_ = $(UI_CONSTANTS.roomLinkHref);
+  
+  this.inCall_ = false;
 
   this.muteAudioIconSet_ = new AppController.IconSet_(
       UI_CONSTANTS.muteAudioOnSvg, UI_CONSTANTS.muteAudioOffSvg);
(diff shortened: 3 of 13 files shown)


samples/web/content/apprtc/apprtc.py
```diff
@@ -515,13 +515,18 @@ def post(self, room_id):
...
-    room_id = generate_random(8)
-    redirect = '/r/' + room_id
-    redirect = append_url_arguments(self.request, redirect)
-    self.redirect(redirect)
-    logging.info('Redirecting visitor to base URL to ' + redirect)
+    """Renders index.html."""
+    # Parse out parameters from request.
+    params = get_room_parameters(self.request, None, None, None)
+    params['connect'] = False
+    params['room_id'] = generate_random(8)
```
jiayliu: If we don't set 'room_id' here, let the client check it and generate a random id, we don't need a separate flag 'connect'.

samples/web/content/apprtc/index.html
```diff
@@ -38,6 +38,20 @@
     <video id="remote-video" autoplay></video>
     <video id="local-video" autoplay muted></video>
   </div>
+  
+  <div id="room-selection" class="hidden">
+     <h1>AppRTC</h1>
+     <p id="instructions">Please enter a room name. Room names are shared with everyone, so think of something unique and send it to a friend.</p>
```
jiayliu: "Shared with everyone" seems confusing to me. I'd rather go with simply "Please enter a room name" before we get more polished text from UI designer.

samples/web/content/apprtc/index.html
```diff
@@ -104,9 +119,14 @@
     var loadingParams = {
       errorMessages: {{ error_messages }},
       isLoopback: {{ is_loopback }},
-
+{% if connect %}
       roomId: '{{ room_id }}',
       roomLink: '{{ room_link }}',
+      connect: true,
+{% else %}
```
jiayliu: See my previous comment. I think we should simplify this.

samples/web/content/apprtc/js/appcontroller.js
```diff
@@ -119,28 +127,69 @@ var AppController = function(loadingParams) {
...
-    setUpFullScreen();
+    this.roomSelection_ = null;
+    if (this.loadingParams_.connect !== false && this.loadingParams_.roomId) {
+      // If we already have a room to connect to, connect now.
+      this.call_.start(this.loadingParams_.roomId);
+      
+      // Record this room in the recently used list.
+      var recentlyUsedList = new RoomSelection.RecentlyUsedList();
+      recentlyUsedList.pushRecentRoom(this.loadingParams_.roomId);
+
```
jiayliu: Another advantage of not generating a random room on the server is that we won't push random room to the list.

samples/web/content/apprtc/js/appcontroller.js
```diff
@@ -58,6 +64,8 @@ var AppController = function(loadingParams) {
   this.remoteVideo_ = $(UI_CONSTANTS.remoteVideo);
   this.videosDiv_ = $(UI_CONSTANTS.videosDiv);
   this.roomLinkHref_ = $(UI_CONSTANTS.roomLinkHref);
+  
+  this.inCall_ = false;
```
jiayliu: Is this necessary? Can we just check this.call_?

jiayliu: If we need to show self preview before call started, that should be moved out of Call.

samples/web/content/apprtc/js/appcontroller.js
```diff
@@ -119,28 +127,69 @@ var AppController = function(loadingParams) {
...
+      this.finishCallSetup_();
+    } else {
+      // Display the room selection UI.
+      var roomSelectionDiv = $(UI_CONSTANTS.roomSelectionDiv);
+      this.roomSelection_ = new RoomSelection(roomSelectionDiv, this.loadingParams_.suggestedRoomId);
+      this.show_(roomSelectionDiv);
+      this.roomSelection_.onRoomSelected = function(roomName) {
+        this.hide_(roomSelectionDiv);
+        this.call_.start(roomName);
+        this.finishCallSetup_();
```
jiayliu: I think it's cleaner to have all call setup done in one helper method, i.e.
if (should show landing page) {
  show landing page;
  onRoomSelected = {
      ...
      setupCall();
  }
  return;
}
setupCall();

samples/web/content/apprtc/js/appcontroller.js
```diff
@@ -321,9 +381,19 @@ AppController.prototype.toggleVideoMute_ = function() {
 
 AppController.prototype.toggleFullScreen_ = function() {
   if (isFullScreen()) {
-    document.cancelFullScreen();
+    trace('Exiting fullscreen.');
+    if (isChromeApp()) {
+      chrome.app.window.current().restore();
+    } else {
+      document.cancelFullScreen();
```
jiayliu: There is already a shim for full screen, can we move the logic there?

samples/web/content/apprtc/js/roomselection.js
```diff
@@ -0,0 +1,134 @@
...
+
+/* More information about these options at jshint.com/docs/options */
+
+/* globals randomString, UI_CONSTANTS, getStorage,
+   setStorage, parseJSON */
+/* exported RoomSelection */
+
+'use strict';
+
+var RoomSelection = function(roomSelectionDiv, randomRoomId) {
```
jiayliu: Can you add test for this class?

samples/web/content/apprtc/js/appcontroller.js
```diff
@@ -119,28 +127,69 @@ var AppController = function(loadingParams) {
...
-    document.onkeypress = this.onKeyPress_.bind(this);
-    window.onmousemove = this.showIcons_.bind(this);
-    
-    $(UI_CONSTANTS.muteAudioSvg).onclick = this.toggleAudioMute_.bind(this);
-    $(UI_CONSTANTS.muteVideoSvg).onclick = this.toggleVideoMute_.bind(this);
-    $(UI_CONSTANTS.fullscreenSvg).onclick = this.toggleFullScreen_.bind(this);
-    $(UI_CONSTANTS.hangupSvg).onclick = this.hangup_.bind(this);
-  
-    setUpFullScreen();
+    this.roomSelection_ = null;
```
jiayliu: Can you add unit test for the new logic?

samples/web/content/apprtc/js/appcontroller.js
```diff
@@ -288,10 +341,17 @@ AppController.prototype.onKeyPress_ = function(event) {
...
+  if (!isChromeApp()) {
+    window.history.pushState({'roomId': roomId, 'roomLink': roomLink }, roomId, roomLink);
+  }  
+};
+
+AppController.prototype.displaySharingInfo_ = function(roomId, roomLink) {
   this.roomLinkHref_.href = roomLink;
   this.roomLinkHref_.text = roomLink;
   this.roomLink_ = roomLink;
+  this.pushCallNavigation_(roomId, roomLink);
```
jiayliu: displaySharingInfo_ shouldn't be responsible for this. Not sure why we need it actually..

samples/web/content/apprtc/js/appcontroller.js
```diff
@@ -119,28 +127,69 @@ var AppController = function(loadingParams) {
...
+    };
+  }
+};
+
 AppController.prototype.hangup_ = function() {
   trace('Hanging up.');
   this.hide_(this.icons_);
   this.displayStatus_('Hanging up');
   this.transitionToDone_();
+  this.inCall_ = false;
```
jiayliu: Does the page go back to the landing page after hangup? I don't see how it happens.

samples/web/content/apprtc/js/roomselection.js
```diff
@@ -0,0 +1,134 @@
+/*
+ *  Copyright (c) 2014 The WebRTC project authors. All Rights Reserved.
```
jiayliu: nit: 2015

samples/web/content/apprtc/js/roomselection.js
```diff
@@ -0,0 +1,134 @@
...
+ *
+ *  Use of this source code is governed by a BSD-style license
+ *  that can be found in the LICENSE file in the root of the source
+ *  tree.
+ */
+
+/* More information about these options at jshint.com/docs/options */
+
+/* globals randomString, UI_CONSTANTS, getStorage,
```
jiayliu: Pass UI_CONSTANTS to ctor to avoid depending on global data?

samples/web/content/apprtc/index.html
chuckhays: OK, I was just copying the android UI. Will change it.

samples/web/content/apprtc/js/appcontroller.js
chuckhays: We wouldn't push the suggested roomId unless they click join, same as if it were generated client side.

samples/web/content/apprtc/js/appcontroller.js
chuckhays: I'll check if there is reasonable way to move it.

samples/web/content/apprtc/js/appcontroller.js
chuckhays: We don't have to do it, but it makes the uri bar show the uri of the current room so that if they want to bookmark it, or copy/paste the direct connection uri they can. It also allows us to use forward/back to get back to the landing page screen.

samples/web/content/apprtc/js/appcontroller.js
chuckhays: It does not currently. I think our next step as we move to the more advanced UI is to plan out/clean up how we enter/exit calls so we can do it multiple times without completely reloading the page. This will be required for the Chrome app.

samples/web/content/apprtc/js/roomselection.js
chuckhays: Sure, I like that idea.

samples/web/content/apprtc/js/appcontroller.js
chuckhays: Not strictly necessary, but we will have to do something like this as we take the next step of having the app be able to make multiple calls without reloading. I can remove it for now if we want.

samples/web/content/apprtc/apprtc.py
chuckhays: I'm ok with either way, the intent is clearer this way, but the intent can be made fairly clear with a comments.

samples/web/content/apprtc/js/appcontroller.js
chuckhays: I was thinking that myself while reviewing it but couldn't decide whether to move it. I'll move it now.

samples/web/content/apprtc/js/appcontroller.js
jiayliu: I think a better way is to unwire the keypressed callback when it's out of the call view

samples/web/content/apprtc/js/appcontroller.js
jiayliu: SG

samples/web/content/apprtc/js/appcontroller.js
chuckhays: I'll do that for now, we'll have to figure out a more central keyboard router/handler as we add the new UI/states. We can do that when we need it.

samples/web/content/apprtc/js/util.js
```diff
@@ -119,3 +124,46 @@ function randomString(strLength) {
...
+      if (callback)
+      {
+        callback(values[key]);
+      }
+    });
+  } else {
+    // use localStorage
+    var value = localStorage.getItem(key);
+    if (callback) {
+      callback(value);
```
juberti: Calling callbacks from within the actual function can lead to reentry. Consider calling this from setInterval.

samples/web/content/apprtc/js/util.js
```diff
@@ -119,3 +124,46 @@ function randomString(strLength) {
...
+
+// Returns true if the code is running in a packaged Chrome App.
+function isChromeApp() {
+  return (typeof chrome !== 'undefined' &&
+          typeof chrome.storage !== 'undefined' &&
+          typeof chrome.storage.local !== 'undefined');
+}
+
+// Get a value from local browser storage. Calls callback with value.
+function getStorage(key, callback) {
```
juberti: Consider making this a Storage object, so that we can have fewer globals.

chuckhays: Good idea, will do.

chuckhays:
Addressed all but the unit tests for AppController, I think we should add unit tests as we refactor appcontroller for the new UI/call model in the next step. I could be convinced to do it now if anyone feels strongly about it.

samples/web/content/apprtc/js/appcontroller.js
```diff
@@ -109,23 +115,60 @@ var AppController = function(loadingParams) {
...
+  
+  if (!isChromeApp()) {
+    window.onpopstate = function(event) {
+      if (!event.state) {
+        // Resetting back to room selection page not yet supported, reload
+        // the initial page instead.
+        trace('Reloading main page.');
+        location.href = location.origin;
+      } else {
+        // This could be a forward request to open a room again
```
jiayliu: nit: end with '.'

samples/web/content/apprtc/js/appcontroller.js
```diff
@@ -109,23 +115,60 @@ var AppController = function(loadingParams) {
...
+        if (event.state && event.state.roomLink) {
```
jiayliu: No need to check event.state again since it's in the else of (!event.state)

samples/web/content/apprtc/js/appcontroller.js
```diff
@@ -109,23 +115,60 @@ var AppController = function(loadingParams) {
...
+  $(UI_CONSTANTS.hangupSvg).onclick = this.hangup_.bind(this);
+
+  setUpFullScreen();
+  
+  if (!isChromeApp()) {
+    window.onpopstate = function(event) {
+      if (!event.state) {
+        // Resetting back to room selection page not yet supported, reload
+        // the initial page instead.
+        trace('Reloading main page.');
```
jiayliu: Please add a TODO (chuckhays):...if you are planning to fix it.

samples/web/content/apprtc/js/call.js
```diff
@@ -328,7 +328,9 @@ Call.prototype.registerWithRoomServer_ = function() {
         reject(Error('Error parsing response JSON.'));
         return;
       }
-      if (responseObj.result !== 'SUCCESS') {
+      if (responseObj.result === 'FULL') {
+        // TODO : handle room full state by returning to room selection state.
```
jiayliu: I think it still needs to reject(Error('...')) for now.
nit: TODO(chuckhays):

samples/web/content/apprtc/js/roomselection.js
```diff
@@ -0,0 +1,139 @@
...
+ */
+
+/* More information about these options at jshint.com/docs/options */
+
+/* globals randomString, Storage, parseJSON */
+/* exported RoomSelection */
+
+'use strict';
+
+var RoomSelection = function(roomSelectionDiv, uiConstants, recentRoomsKey) {
```
jiayliu: recentRoomKey is not used by any caller?

samples/web/content/apprtc/js/roomselection.js
```diff
@@ -0,0 +1,139 @@
...
+  this.roomSelectionDiv_ = roomSelectionDiv;
+  
+  this.uiConstants = uiConstants;
```
jiayliu: this.uiConstants seems unused outside the ctor. Remove it?

samples/web/content/apprtc/js/roomselection.js
```diff
@@ -0,0 +1,139 @@
...
+  this.recentlyUsedList_.getRecentRooms().then(this.buildRecentRoomList_.bind(this)).catch(function(error) {
+    trace('Error building recent rooms list: ' + error.message);
+  }.bind(this));
+};
+
+RoomSelection.prototype.buildRecentRoomList_ = function(recentRooms) {
+  for (var i = 0; i < recentRooms.length; ++i) {
+    // Create link in recent list
+    var li = document.createElement('li');
+    var href = document.createElement('a');
```
jiayliu: Why use 'a' if you want to avoid navigation?

samples/web/content/apprtc/js/roomselection.js
```diff
@@ -0,0 +1,139 @@
...
+    li.appendChild(href);
+    this.roomRecentList_.appendChild(li);
+  
+    // Set up click handler to avoid browser navigation.
+    href.addEventListener('click', this.makeRecentlyUsedClickHandler_(recentRooms[i]).bind(this), false);
+  }
+};
+
+RoomSelection.prototype.onRoomIdInput_ = function() {
+  // validate room id, enable/disable join button
```
jiayliu: nit: first letter should be capitalized

jiayliu: nit: end with '.'

samples/web/content/apprtc/js/roomselection.js
```diff
@@ -0,0 +1,139 @@
...
+  // The server currently accepts only the \w character class
+  var room = this.roomIdInput_.value;
+  var valid = room.length >= 5;
+  var re = /^\w+$/;
+  valid = valid && re.exec(room);
+  if (valid) {
```
jiayliu: Add a TODO to add a user hint later

samples/web/content/apprtc/js/roomselection.js
```diff
@@ -0,0 +1,139 @@
...
+    if (!roomId) {
+      resolve();
+      return;
+    }
+    
+    this.getRecentRooms().then(function(recentRooms) {
+      recentRooms = [roomId].concat(recentRooms);
+      // Remove any duplicates from the list, leaving the first occurance.
+      recentRooms = recentRooms.filter(function(value, index, self) {
+        return self.indexOf(value) === index;
```
jiayliu: Why not just "value != roomId", and move [roomId].concat(recentRoos) after the filter?

samples/web/content/apprtc/js/roomselection_test.js
```diff
@@ -0,0 +1,114 @@
...
+  targetDiv.appendChild(inputBox);
+  targetDiv.appendChild(randomButton);
+  targetDiv.appendChild(joinButton);
+  targetDiv.appendChild(recentList);
+  
+  this.roomSelection_ = new RoomSelection(targetDiv, UI_CONSTANTS, key);
+};
+
+RoomSelectionTest.prototype.tearDown = function() {
+  this.roomSelection_ = null;
```
jiayliu: Clean up the storage?

samples/web/content/apprtc/js/roomselection_test.js
```diff
@@ -0,0 +1,114 @@
...
+};
+
+RoomSelectionTest.prototype.testInputFilter = function() {
+  var validInputs = [ '123123', 'asdfs3', 'room1', '3254234523452345234523452345asdfasfdasdf'];
+  var invalidInputs = ['', ' ', 'abcd', '123', '[5afasdf', 'ñsaer3'];
+  
+  var testInput = function(roomSelection, input, expectedResult) {
+    roomSelection.roomIdInput_.value = input;
```
jiayliu: I'd avoid depending on private members of the tested class, since it makes the tests harder to maintain.

Since the test class knows all the DOM elements, can it just interact with them directly and test the public interface of RoomSelection?

samples/web/content/apprtc/js/storage.js
```diff
@@ -0,0 +1,58 @@
...
+
+'use strict';
+
+var Storage = function() {};
+
+// Get a value from local browser storage. Calls callback with value.
+// Handles variation in API between localStorage and Chrome app storage.
+Storage.prototype.getStorage = function(key, callback) {
+  if (isChromeApp()) {
+    // use chrome.storage.local
```
jiayliu: nit: Capital first letter and end with '.'. Here and other places.

samples/web/content/apprtc/js/storage.js
```diff
@@ -0,0 +1,58 @@
...
+    chrome.storage.local.get(key, function(values) {
+      // unwrap key/value pair
+      if (callback)
+      {
+        window.setTimeout(function() {
```
jiayliu: Why setTimeout?

samples/web/content/apprtc/js/roomselection.js
```diff
@@ -0,0 +1,139 @@
...
+
+RoomSelection.RecentlyUsedList = function(key) {
+  this.RECENTROOMSKEY_ = key || 'recentRooms';
+  this.storage_ = new Storage();
+};
+
+// Add a room to the recently used list and store to local storage.
+RoomSelection.RecentlyUsedList.prototype.pushRecentRoom = function(roomId) {
+  // Push recent room to top of recent list, keep max of 10 entries.
+  return new Promise(function(resolve, reject) {
```
jiayliu: It seems no one cares about the result of pushRecentRoom. Does it need to return a promise?

samples/web/content/apprtc/js/roomselection.js
chuckhays: Only used in the tests to avoid reading/writing the real key.

samples/web/content/apprtc/js/roomselection.js
chuckhays: They were originally actual navigation links in the web app. We can/should change this when we implement the upcoming UI changes.

samples/web/content/apprtc/js/roomselection.js
chuckhays: That would work too, the only advantage of this way is that it removes other duplicates as well if they happened to be in the list.

samples/web/content/apprtc/js/storage.js
chuckhays: @juberti suggested changing it from a direct call to callback to avoid possible reentrancy issues.

samples/web/content/apprtc/js/roomselection.js
chuckhays: Good catch, it was originally part of a chain, not anymore.

chuckhays:
Updated with CR feedback changes. Unit tests now use the DOM elements and avoid private members as much as possible.

chuckhays:
We could also consider polling to check if/when it is done (by checking if children have been created), but the risk is that the test might take a long time to fail.

chuckhays:
Updated per your suggestions. PTAL. 

Unit tests prove their worth again, in spite of looking at the MRU lists for days, it only had 9 entries.

juberti:
lgtm!

PR #393, opened 2015-01-16
Add polyfill for getStats to spec.

milkypostman:
This replaces the getStats function on RTCPeerConnection to match the
format in the spec: http://w3c.github.io/webrtc-pc/#statistics-model.

### DIFF
--- a/samples/web/js/adapter.js
+++ b/samples/web/js/adapter.js
@@ -1,5 +1,5 @@
 /*
- *  Copyright (c) 2014 The WebRTC project authors. All Rights Reserved.
+ *  Copyright (c) 2014-2015 The WebRTC project authors. All Rights Reserved.
  *
  *  Use of this source code is governed by a BSD-style license
  *  that can be found in the LICENSE file in the root of the source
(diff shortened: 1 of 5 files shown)


juberti:
Please fix the errors from the Travis build.
@pthatcherg can probably best comment on the getStats format changes.

juberti:
oops, looks like Travis has no problems with this

milkypostman:
It did, then I fixed it. :crocodile:

milkypostman:
should there be unit tests for this? Should I try to add them? I don't know that node actually implements the WebRTC api so maybe it doesn't make any sense or would be a headache.

samples/web/js/adapter.js
```diff
@@ -179,9 +179,35 @@ if (navigator.mozGetUserMedia) {
     };
   };
 
+  var fixChromeStats = function(response) {
```
juberti: Can you add a comment showing how the fixup works (e.g. before/after?)

juberti:
Yeah, I think unit tests would be a good habit to start, now that we have the infrastructure in place.

milkypostman:
@juberti Unittests have been added.

@phoglund please have a look at my test changes.

juberti:
Sorry for the trouble, please resend this PR to the webrtc/adapter repo

Issue #402, opened 2015-01-21
Grunt/JSCS is not running for samples


chuckhays:
Grunt-JSCS can't find it's config file specified by 'google', and is silently failing. 

This change enables it, but there are a bunch of errors to fix, or rules to adjust, before we can merge it:
https://github.com/chuckhays/webrtc/commit/7fd09eea6746f120b982c6c6600280e27e6f17e2

@samdutton @jiayliu @juberti

samdutton:
I set this up with the wrong key (or maybe it changed): it should be
preset: 'google' not config: 'google'.

See github.com/jscs-dev/grunt-jscs for more information.

As you say - this results in a number of errors in testrtc files.

I think we should change to preset: 'google' and fix the problems, and
maybe create a separate (additional) config file, or (better) use .jscsrc.

Sam

chuckhays:
I'll try the preset key, I had to dig into the node module to figure out how to make it do anything at all with the config file. Hopefully preset will be simpler.

chuckhays:
We should keep this open to track fixing the other sources. @samdutton do you have any suggestions who might be the right people for the other areas? I can take some of them too, just wasn't sure who typically does each area.

juberti:
@KaptenJansson for testrtc
@samdutton for samples

samdutton:
See #469 .

samdutton:
Fixed in #469.

PR #407, opened 2015-01-21
Enable JSCS on apprtc code #402

chuckhays:
Issue #402:

The jscs plugin for grunt wasn't enforcing any rules. This change enables it only on apprtc and fixes the apprtc violations. All changes to code should be effectively whitespace only.

We should also enable it for the other directories, but we can do the cleanup of each directory separately.

@samdutton @jiayliu @juberti


jiayliu:
lgtm

samdutton:
Thanks Chuck – and for noticing this in the first place.

PR #411, opened 2015-01-22
Fix 404 errors on index and full.html

juberti:
Also, sync meta items on full.html with index.html

@samdutton @chuckhays PTAL


samdutton:
See PR #413.

PR #415, opened 2015-01-22
Use pretty print to assist debugging.

chuckhays:
Without pretty print, the javascript is all one line, making debugging in the debug build nearly impossible. With pretty print, the javascript is still all in 1 file, but is formatted into multiple lines.

@jiayliu


jiayliu:
lgtm

PR #417, opened 2015-01-23
Rename register to join and bye to leave for GAE API

jiayliu:
I did't keep the old API since this branch will not be pushed to prod apprtc, so no compatibility issue.


jiayliu:
@tkchin @chuckhays @juberti PTAL

juberti:
I think we should maintain backcompat to ensure that both old and new clients can work with prod or the development server.

jiayliu:
Fine. Added back.

juberti:
lgtm. Let's merge this commit to trunk as well.

chuckhays:
lgtm as well.

PR #420, opened 2015-01-23
Fix a race condition with transitionToWaiting and adding remote stream

jiayliu:
The timer has been causing timing issues in the past. I wonder if we can skip resetting remoteVideo/miniVideo src all together.

### DIFF
--- a/samples/web/content/apprtc/js/appcontroller.js
+++ b/samples/web/content/apprtc/js/appcontroller.js
@@ -90,7 +90,7 @@ var AppController = function(loadingParams) {
     this.infoBox_ =
         new InfoBox($(UI_CONSTANTS.infoDiv), this.remoteVideo_, this.call_);
 
-    this.transitionToWaitingTimer_ = null;
+    this.remoteVideoResetTimer_ = null;
 
     var roomErrors = this.loadingParams_.errorMessages;
     if (roomErrors.length > 0) {
@@ -214,6 +214,11 @@ AppController.prototype.onRemoteStreamAdded_ = function(stream) {
   this.deactivate_(this.sharingDiv_);
   trace('Remote stream added.');
   attachMediaStream(this.remoteVideo_, stream);
+
+  if (this.remoteVideoResetTimer_) {
+    clearTimeout(this.remoteVideoResetTimer_);
+    this.remoteVideoResetTimer_ = null;
+  }
 };
 
 AppController.prototype.onLocalStreamAdded_ = function(stream) {
@@ -234,11 +239,6 @@ AppController.prototype.transitionToActive_ = function() {
   this.infoBox_.updateInfoDiv();
   trace('Call setup time: ' + (connectTime - this.call_.startTime).toFixed(0) +
       'ms.');
-
-  if (this.transitionToWaitingTimer_) {
-    clearTimeout(this.transitionToWaitingTimer_);
-    this.transitionToWaitingTimer_ = null;
-  }
 
   // Prepare the remote video and PIP elements.
   trace('reattachMediaStream: ' + this.localVideo_.src);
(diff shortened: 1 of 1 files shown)


juberti:
In the failure case, what's the sequence of operations?

jiayliu:
user 2 leaves
user1 transition to waiting, starting the timer
user2 rejoins
user1 added remote stream, attached to remoteVideo elem
user1 timer fires, wipes out remoteVideo
user1 and user2 ICE connected
user1 stuck in waiting for remote video


On Fri, Jan 23, 2015 at 10:40 AM, Justin Uberti <notifications@github.com>
wrote:

> In the failure case, what's the sequence of operations?
>
> —
> <https://github.com/GoogleChrome/webrtc/pull/420#issuecomment-71241832>.
>



-- 
Jiayang Liu

juberti:
When does onRemoteSdpSet get called in this flow?

Overall I think we need some sort of state variable and we can check when doing any transition whether we are in the correct state.

jiayliu:
onRemoteSdpSet is after "user1 added remote stream, attached to remoteVideo
elem", before "user1 and user2 ICE connected".

Why cannot we just leave the remoteVideo and miniVideo src? They are hidden
anyway.

On Fri, Jan 23, 2015 at 10:52 AM, Justin Uberti <notifications@github.com>
wrote:

> When does onRemoteSdpSet get called in this flow?
>
> Overall I think we need some sort of state variable and we can check when
> doing any transition whether we are in the correct state.
>
> —
> <https://github.com/GoogleChrome/webrtc/pull/420#issuecomment-71244066>.
>



-- 
Jiayang Liu

juberti:
I think it's fine to leave miniVideo set. I think we need to clear remoteVideo since seeing a stale frame from the previous caller could be a privacy issue.

jiayliu:
But remoteVideo is not visible even without resetting src, due to
this.deactivate_(this.remoteVideo_)

On Fri, Jan 23, 2015 at 11:42 AM, Justin Uberti <notifications@github.com>
wrote:

> I think it's fine to leave miniVideo set. I think we need to clear
> remoteVideo since seeing a stale frame from the previous caller could be a
> privacy issue.
>
> —
> <https://github.com/GoogleChrome/webrtc/pull/420#issuecomment-71252647>.
>



-- 
Jiayang Liu

juberti:
My concern is around what happens when the call is reconnected, we might display the video tag with a stale frame briefly. It seems like we ought to GC the old remote stream as soon as it transitions out of view. But maybe that would be better accomplished by having some wrapper object for the video tag.

samples/web/content/apprtc/js/appcontroller.js
```diff
@@ -266,9 +271,11 @@ AppController.prototype.transitionToWaiting_ = function() {
 
   this.transitionToWaitingTimer_ = setTimeout(function() {
```
juberti: One way to solve this would be to call this remoteVideoResetTimer, and always cancel it when setting remoteVideo to a stream.

I think we can get rid of the code block at line 243 with this change.

jiayliu: Makes sense. Will do that.

On Fri, Jan 23, 2015 at 1:26 PM, Justin Uberti <notifications@github.com>
wrote:

> In samples/web/content/apprtc/js/appcontroller.js
> <https://github.com/GoogleChrome/webrtc/pull/420#discussion_r23481519>:
>
> > @@ -266,9 +271,11 @@ AppController.prototype.transitionToWaiting_ = function() {
> >
> >    this.transitionToWaitingTimer_ = setTimeout(function() {
>
> One way to solve this would be to call this remoteVideoResetTimer, and
> always cancel it when setting remoteVideo to a stream.
>
> I think we can get rid of the code block at line 243 with this change.
>
> —
> <https://github.com/GoogleChrome/webrtc/pull/420/files#r23481519>.
>



-- 
Jiayang Liu

jiayliu:
PTAL

juberti:
lgtm

PR #423, opened 2015-01-23
Move json methods to shared util module

chuckhays:
Moves json methods to util module so we can use them in apprtc.py. Adds a couple of unit tests for the util methods.

Sending now as a separate PR to avoid merge conflicts with multiple people working in this code.

@tkchin @jiayliu

### DIFF
--- a/samples/web/content/apprtc/util_test.py
+++ b/samples/web/content/apprtc/util_test.py
@@ -1,0 +1,74 @@
+# Copyright 2014 Google Inc. All Rights Reserved.
+
+import unittest
+
+import util
+
+class UtilTest(unittest.TestCase):
+  def testGetMessageFromJson(self):
+    self.assertEqual(None, util.get_message_from_json(""))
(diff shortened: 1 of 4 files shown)


samples/web/content/apprtc/util_test.py
```diff
@@ -0,0 +1,74 @@
...
+    testObject = {
+      "a": False,
+      "b": "str",
+      "c": None,
+      "d": {},
+      "e": [1, 2, "3"],
+      "f": [],
+      "g": {'A': 1}
+    }
+    self.assertEqual(
```
jiayliu: Also add a test case for a key not in the dict?

jiayliu:
lgtm with nit

samples/web/content/apprtc/util_test.py
chuckhays: Added a test for missing key and passing None as a message, which could happen if the json is invalid.

chuckhays:
Added 2 test cases and made a resulting change to the method, PTAL.

PR #424, opened 2015-01-23
Navigate to landing page after hangup.

chuckhays:
This is all it would take to redirect back to the landing page after hitting the hang up button.

@juberti

### DIFF
--- a/samples/web/content/apprtc/js/appcontroller.js
+++ b/samples/web/content/apprtc/js/appcontroller.js
@@ -31,7 +31,12 @@ var UI_CONSTANTS = {
   muteAudioSvg: '#mute-audio',
   muteVideoSvg: '#mute-video',
 
+  newRoomButton: '#new-room-button',
+  newRoomLink: '#new-room-link',
   remoteVideo: '#remote-video',
+  rejoinButton: '#rejoin-button',
+  rejoinDiv: '#rejoin-div',
+  rejoinLink: '#rejoin-link',
   roomLinkHref: '#room-link-href',
   roomSelectionDiv: '#room-selection',
   roomSelectionInput: '#room-id-input',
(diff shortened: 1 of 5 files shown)


samples/web/content/apprtc/js/appcontroller.js
```diff
@@ -179,6 +179,9 @@ AppController.prototype.hangup_ = function() {
   this.transitionToDone_();
 
   this.call_.hangup();
+
+  trace('Reloading main page.');
```
juberti: I tried this, but I think we will want to retain the end-of-call page, perhaps to display a call rating widget or some such. But let's add a link in the status area to take you back to the landing page, e.g. [Click here] to rejoin, or [start a new room].

Also, can we make the rejoin-room link immediately put you back into the room (i.e. act as if you clicked Join from the landing page)?

chuckhays: Yeah, that makes sense. I'll take a look.

chuckhays:
Added a new room link like you suggested. Not sure I like the text as it is.

Maybe it should be something like:
You have left the call: [rejoin] or start a [new room].

screen shot 2015-01-26 at 8 26 25 am

chuckhays:
> Also, can we make the rejoin-room link immediately put you back into the room (i.e. act as if you clicked Join from the landing page)?

In what way? Removing the page navigation?

samdutton:
LGTM.

jiayliu:
This won't work for the Chrome App, right?

juberti:
New text SGTM.

Re: rejoin action, I think the navigation might be fine, as long as it works in the Chrome app.

chuckhays:
It won't work as is for the Chrome app, needs to have a click handler hooked up. I'll have to check how the code responds to making another call without a "reload". We'll have to take on that work eventually, I was just avoiding it to work on our other priorities.

chuckhays:
This change should work with both the Chrome App and the web app. There are fixes needed in the Chrome App, as the hang-up functionality doesn't work properly due to using sync xhr, which isn't allowed. This could be fixed at the same time as issue #437.

screen shot 2015-01-27 at 11 43 02 am

juberti:
lgtm

PR #431, opened 2015-01-26
Add Google API dependencies.

milkypostman:
@jiayliu 

This is gathered from [google-api-python-client](https://code.google.com/p/google-api-python-client/downloads/list) and is the `google-api-python-client-gae-1.2.zip` unzipped so that our API can use it.


jiayliu:
lgtm

juberti:
Since these files aren't under the top-level LICENSE, I think they should go under an apprtc/third_party directory.

milkypostman:
@juberti should I add a LICENSE file or README inside the `third_party` directory mentioning something about licensing?

juberti:
Yeah, I think a README explaining that this is third-party code (and why we need it here, as opposed to pulling it as a npm module) is an excellent idea.

milkypostman:
readme added. ptal

PR #432, opened 2015-01-26
Requires the user to confirm when joining from /r/{room} for apprtc

jiayliu:
A confirm button is added on top of the self preview asking for user's action.

A test is added.


samples/web/content/apprtc/index.html
```diff
@@ -58,6 +58,11 @@
     </div>
   </div>
 
+  <div id="confirm-join-div" class="hidden">
+    <div> Are you ready to join <span id="confirm-join-room-span"></span>?</div>
```
juberti: "Ready to join \"foo\"?" is shorter. Showing the room name in quotes helps make it obvious that this is the name of the room.

Also if the room name is one of our randomly-generated ones, I think we should simply omit it and display only "Ready to join?"

samples/web/content/apprtc/js/appcontroller.js
```diff
@@ -119,13 +119,25 @@ var AppController = function(loadingParams) {
     this.call_.oncallerstarted = this.displaySharingInfo_.bind(this);
 
     this.roomSelection_ = null;
-    // If the params has a roomId specified, we should connect to that room immediately.
-    // If not, show the room selection UI.
+    this.localStream_ = null;
+
+    // If the params has a roomId specified, we should connect to that room
+    // immediately. If not, show the room selection UI.
```
juberti: This function is getting quite long. Consider breaking out this UI setup code into its own function.

samples/web/content/apprtc/index.html
jiayliu: We cannot tell if the room name is randomly generated...

samples/web/content/apprtc/js/appcontroller.js
```diff
@@ -134,6 +146,11 @@ var AppController = function(loadingParams) {
       this.roomSelection_.onRoomSelected = function(roomName) {
         this.hide_(roomSelectionDiv);
         this.finishCallSetup_(roomName);
+
+        this.roomSelection_ = null;
+        if (this.localStream_) {
```
juberti: I don't think we want to ask for the camera if we're not going to display it right away.

samples/web/content/apprtc/js/call.js
```diff
@@ -191,11 +192,12 @@ Call.prototype.connectToRoom_ = function(roomId, mediaPromise, turnPromise) {
     // and have media and TURN. Since we send candidates as soon as the peer
     // connection generates them we need to wait for the signaling channel to be
     // ready.
-    Promise.all([turnPromise, mediaPromise]).then(function() {
-      this.startSignaling_();
-    }.bind(this)).catch(function(error) {
-      this.onError_('Failed to start signaling: ' + error.message);
-    }.bind(this));
+    Promise.all([this.getTurnServersPromise_, this.getMediaPromise_])
```
juberti: would it be easier to just call maybeGetMedia here? getMediaPromise_ doesn't seem to be used elsewhere.

samples/web/content/apprtc/js/appcontroller.js
jiayliu: Chuck is working on showing the self preview in the room selection UI. So it will be displayed right away with that change.

samples/web/content/apprtc/js/call.js
jiayliu: connectToRoom_ is called after the user confirms to join. We need to request media earlier.

samples/web/content/apprtc/index.html
juberti: If the room is all numeric, or perhaps 8 chars and all numeric, I think that would be fine for now.

The long-term solution here is to generate rooms with random words, not random numbers.

samples/web/content/apprtc/js/call.js
juberti: OK, I misunderstood the code. I still think we don't want to start the camera on the landing page until Chuck's change is in. It's strange to have the cam light on without any clear indication as to why.

jiayliu: OK, now media not requested in the room selection page.

jiayliu:
CR comments addressed. PTAL

juberti:
lgtm

jiayliu:
@phoglund We need to update the browser test again. I'll send a patch.

chuckhays:
LGTM as well.

jiayliu:
Added a flag to bypass the join confirmation for automation.
The flag can only be set from app.yaml.

@phoglund,
In the test, you should replace 
"BYPASS_JOIN_CONFIRMATION: false"
with
"BYPASS_JOIN_CONFIRMATION: true"
in app.yaml.

jiayliu:
@juberti PTAL

chuckhays:
The update LGTM.

juberti:
lgtm

PR #436, opened 2015-01-27
Modify /join to support id based calling.

chuckhays:
This change modifies /join to support id based calling. It does not yet do any of the GCM messaging. Those are marked as TODO items. 

apprtc.py is getting pretty large, we may want to consider moving some pieces of it out to another file/class.

@jiayliu @tkchin


samples/web/content/apprtc/apprtc.py
```diff
@@ -495,12 +515,109 @@ def write_response(self, result, params, messages):
...
+        'from gcmId ' + calleeId)
+    logging.info('User ' + callee.user_id + ' joined room ' + room_id)
+    logging.info('Room ' + room_id + ' has state ' + result['room_state'])
 
   def post(self, room_id):
+    # Check request body to determine what action to take.
+    msg = util.get_message_from_json(self.request.body)
+    if util.has_msg_field(msg, PARAM_ACTION, basestring):
+      action = msg[PARAM_ACTION]
+      if action == 'call':
```
jiayliu: 1. Please update the design doc for API changes like this. 
2. Move 'call'/'accept' into constant vars?

samples/web/content/apprtc/apprtc.py
```diff
@@ -333,7 +340,14 @@ def __str__(self):
 def get_memcache_key_for_room(host, room_id):
   return '%s/%s' % (host, room_id)
 
-def add_client_to_room(host, room_id, client_id, is_loopback):
+def does_room_exist(host, room_id):
+  key = get_memcache_key_for_room(host, room_id)
+  memcache_client = memcache.Client()
+  room = memcache_client.gets(key)
```
jiayliu: gets seems unnecessary, 'get' should be enough

samples/web/content/apprtc/apprtc.py
```diff
@@ -495,12 +515,109 @@ def write_response(self, result, params, messages):
...
     }))
+    
+  def report_error(self, error):
+    self.write_response(error, {}, [])
 
   def write_room_parameters(self, room_id, client_id, messages, is_initiator):
     params = get_room_parameters(self.request, room_id, client_id, is_initiator)
     self.write_response('SUCCESS', params, messages)
+    
+  def handle_call(self, msg, room_id):
```
tkchin: Does it make more sense for caller to call with no room id, in which case the server assigns and returns an unused one? (For those cases where collisions happen this avoids an additional request). Callee still needs to accept by joining that particular room though. Alternately I suppose caller can call with a uuid.

samples/web/content/apprtc/apprtc.py
```diff
@@ -495,12 +515,109 @@ def write_response(self, result, params, messages):
...
+    """Handles a new call request."""
+    if not util.has_msg_fields(msg, ((PARAM_CALLER_GCM_ID, basestring),
```
tkchin: Would it be cleaner to move all the requests into their own classes with parse methods? That way we can clean up the file(s) a bit since there won't be PARAM_* globals and these type checks. Instead we can have something like CallRequest.parse(msg).

samples/web/content/apprtc/apprtc.py
```diff
@@ -495,12 +515,109 @@ def write_response(self, result, params, messages):
...
+        (PARAM_CALLEE_ID, basestring))):
+      self.report_error(constants.RESPONSE_INVALID_ARGUMENT)
+      return
+      
+    callerId = msg[PARAM_CALLER_GCM_ID]
+    calleeId = msg[PARAM_CALLEE_ID]
+    # Look up and validate caller by gcm id.
+    callerRecords = gcmrecord.GCMRecord.get_by_gcm_id(callerId, True)
```
jiayliu: We need to pass verified=False so that we can test it out without SMS verification, with a TODO to make clear that it needs to be fixed later.

samples/web/content/apprtc/apprtc.py
```diff
@@ -495,12 +515,109 @@ def write_response(self, result, params, messages):
...
+    if len(callerRecords) < 1:
+      self.report_error(constants.RESPONSE_INVALID_CALLER)
+      return
+    caller = callerRecords[0]
+    # Look up callee by id.
+    calleeRecords = gcmrecord.GCMRecord.get_by_user_id(calleeId, True)
```
jiayliu: ditto verified=False

samples/web/content/apprtc/apprtc.py
```diff
@@ -333,7 +340,14 @@ def __str__(self):
 def get_memcache_key_for_room(host, room_id):
   return '%s/%s' % (host, room_id)
 
-def add_client_to_room(host, room_id, client_id, is_loopback):
+def does_room_exist(host, room_id):
```
tkchin: nit: if does_room_exist reads strangely (at least to me). Maybe just has_room(host, room_id)?

samples/web/content/apprtc/apprtc_test.py
```diff
@@ -96,6 +103,246 @@ def testJoinAndLeave(self):
     response = self.makePostRequest('/join/' + room_id)
     caller_id = self.verifyJoinSuccessResponse(response, True, room_id)
     self.makePostRequest('/leave/' + room_id + '/' + caller_id)
+    
+  def addTestData(self):
```
jiayliu: The test file is really long now. Probably move the new tests to a separate file.

samples/web/content/apprtc/apprtc.py
```diff
@@ -333,7 +340,14 @@ def __str__(self):
...
+  key = get_memcache_key_for_room(host, room_id)
+  memcache_client = memcache.Client()
+  room = memcache_client.gets(key)
+  return room is not None
+
+def add_client_to_room(host, room_id, client_id,
+    is_loopback, intendedCallee = None):
```
tkchin: It's not clear if intendedCallee is a user_id, a gcm_id or an object. Maybe just callee_id? It might be worth storing it as a permissions list instead, since Room is set up for multiple clients right now it looks like. So e.g maybe do Room.clients and Room.connectedClients.

samples/web/content/apprtc/apprtc.py
```diff
@@ -495,12 +515,109 @@ def write_response(self, result, params, messages):
...
     self.write_response('SUCCESS', params, messages)
+    
+  def handle_call(self, msg, room_id):
+    """Handles a new call request."""
+    if not util.has_msg_fields(msg, ((PARAM_CALLER_GCM_ID, basestring),
+        (PARAM_CALLEE_ID, basestring))):
+      self.report_error(constants.RESPONSE_INVALID_ARGUMENT)
+      return
+      
+    callerId = msg[PARAM_CALLER_GCM_ID]
```
tkchin: caller_gcm_id
also, style-wise no camel casing, here and below.

samples/web/content/apprtc/apprtc.py
```diff
@@ -495,12 +515,109 @@ def write_response(self, result, params, messages):
...
+    calleeId = msg[PARAM_CALLEE_ID]
+    # Look up and validate caller by gcm id.
+    callerRecords = gcmrecord.GCMRecord.get_by_gcm_id(callerId, True)
+    if len(callerRecords) < 1:
+      self.report_error(constants.RESPONSE_INVALID_CALLER)
+      return
+    caller = callerRecords[0]
```
tkchin: maybe caller_id = callerRecords[0].user_id instead

samples/web/content/apprtc/apprtc.py
tkchin: When are we removing the old ability to join by room id alone?

samples/web/content/apprtc/apprtc.py
```diff
@@ -495,12 +515,109 @@ def write_response(self, result, params, messages):
...
+        room_id, caller.user_id, result['messages'], result['is_initiator'])
+      
+    logging.info('User ' + caller.user_id + ' initiated call to ' + calleeId + \
+        ' from gcmId ' + callerId)
+    logging.info('User ' + caller.user_id + ' joined room ' + room_id)
+    logging.info('Room ' + room_id + ' has state ' + result['room_state'])
+
+  def handle_accept(self, msg, room_id):
+    """Handles a callee accepting a call request."""
+    if not util.has_msg_field(msg, PARAM_CALLEE_GCM_ID, basestring):
```
tkchin: ditto .parse

samples/web/content/apprtc/apprtc.py
```diff
@@ -495,12 +515,109 @@ def write_response(self, result, params, messages):
...
+      self.report_error(constants.RESPONSE_INVALID_ARGUMENT)
+      return
+
+    calleeId = msg[PARAM_CALLEE_GCM_ID]
```
tkchin: ditto callee_gcm_id

samples/web/content/apprtc/apprtc.py
chuckhays: @jiayliu did you have something in mind for how to differentiate these calls? I just went with what was simple at the time (including a action: value) because it wasn't clear to me if we wanted to be specific or just try to infer from the values passed.

samples/web/content/apprtc/apprtc.py
```diff
@@ -495,12 +515,109 @@ def write_response(self, result, params, messages):
...
+    # Look up and validate callee by gcm id.
+    calleeRecords = gcmrecord.GCMRecord.get_by_gcm_id(calleeId, True)
+    if len(calleeRecords) < 1:
+      self.report_error(constants.RESPONSE_INVALID_CALLEE)
+      return
+
+    callee = calleeRecords[0]
```
tkchin: ditto callee_id = calleeRecords[0].user_id

samples/web/content/apprtc/apprtc.py
chuckhays: I agree with this, I was thinking about collisions when working on it and had the same thought. We could also internally prefix the room name with their ID or something similar to give each user their own namespace.

samples/web/content/apprtc/apprtc.py
chuckhays: Renamed as suggested.

samples/web/content/apprtc/apprtc.py
chuckhays: I changed it to be a list of allowed ids.

samples/web/content/apprtc/apprtc.py
chuckhays: That is going to take me a bit to retrain my fingers, please keep pointing it out.

samples/web/content/apprtc/apprtc.py
chuckhays: What do you think about doing it on the other side? We could do one of two things:
1. On the /register command, we can mark the record verified
2. Accept any code in /verify

jiayliu: We are not planning to implement sending verification code from the client in the first milestone, so #2 is not appropriate.
For #1, it will leave 'incorrect' data in the datastore, which is less preferable than a simple code change in the future.

chuckhays: My concern is that we will forget to make this change. Just trying to figure out the best way to guarantee we don't overlook it.

samples/web/content/apprtc/apprtc.py
jiayliu: We are not removing the join by room id any time soon.
I think it's fine to use the action param to differentiate these requests.

samples/web/content/apprtc/apprtc.py
jiayliu: One option is to add a flag to constants.py, e.g.
ID_VERIFICATION_SUPPOORTED = False, and use this flag in such APIs calls.
So in the future we can just flip this flag when verification is added.

On Tue, Jan 27, 2015 at 2:26 PM, Chuck Hays <notifications@github.com>
wrote:

> In samples/web/content/apprtc/apprtc.py
> <https://github.com/GoogleChrome/webrtc/pull/436#discussion_r23648902>:
>
> >
> >    def write_room_parameters(self, room_id, client_id, messages, is_initiator):
> >      params = get_room_parameters(self.request, room_id, client_id, is_initiator)
> >      self.write_response('SUCCESS', params, messages)
> > +
> > +  def handle_call(self, msg, room_id):
> > +    """Handles a new call request."""
> > +    if not util.has_msg_fields(msg, ((PARAM_CALLER_GCM_ID, basestring),
> > +        (PARAM_CALLEE_ID, basestring))):
> > +      self.report_error(constants.RESPONSE_INVALID_ARGUMENT)
> > +      return
> > +
> > +    callerId = msg[PARAM_CALLER_GCM_ID]
> > +    calleeId = msg[PARAM_CALLEE_ID]
> > +    # Look up and validate caller by gcm id.
> > +    callerRecords = gcmrecord.GCMRecord.get_by_gcm_id(callerId, True)
>
> My concern is that we will forget to make this change. Just trying to
> figure out the best way to guarantee we don't overlook it.
>
> —
> <https://github.com/GoogleChrome/webrtc/pull/436/files#r23648902>.
>



-- 
Jiayang Liu

chuckhays: I put a couple of TODO lines in and added a note in the place where we will implement the SMS verification. Unlikely we will miss it. I'm just being overly cautious.

chuckhays:
The 2 comments that I haven't addressed are:
1. Moving handlers to a new class/file: I think we should do this in a separate refactoring PR
2. Possibly not accepting roomid from caller: we should discuss this before doing it.

PTAL

samples/web/content/apprtc/apprtc.py
jiayliu: Ping "Move 'call'/'accept' into constant vars?"

samples/web/content/apprtc/apprtc.py
```diff
@@ -312,14 +318,25 @@ def __str__(self):
 class Room:
   def __init__(self):
     self.clients = {}
+    self.allowed_clients = None;
```
jiayliu: It needs some documentation:
1. the initiator is also included in the list
2. anyone is allowed if it's None

samples/web/content/apprtc/apprtc.py
```diff
@@ -312,14 +318,25 @@ def __str__(self):
...
+      self.allowed_clients = []
+    if client_id not in self.allowed_clients:
+      self.allowed_clients.append(client_id)
   def remove_client(self, client_id):
     del self.clients[client_id]
   def get_occupancy(self):
     return len(self.clients)
   def has_client(self, client_id):
     return client_id in self.clients
+  def has_allowed_client(self, client_id):
```
jiayliu: nit: is_client_allowed seems more intuitive and consistent with the logic

samples/web/content/apprtc/apprtc.py
chuckhays: Good point. I wrote a test for the same reason but didn't think to add the documentation there.

samples/web/content/apprtc/apprtc.py
chuckhays: Sounds good.

samples/web/content/apprtc/apprtc.py
```diff
@@ -495,12 +526,115 @@ def write_response(self, result, params, messages):
...
+    caller_gcm_id = msg[PARAM_CALLER_GCM_ID]
+    callee_id = msg[PARAM_CALLEE_ID]
+    # Look up and validate caller by gcm id.
+    # TODO (chuckhays): Once registration is enabled, turn on restriction
+    # to only return verified records.
+    caller_records = gcmrecord.GCMRecord.get_by_gcm_id(caller_gcm_id, False)
+    if len(caller_records) < 1:
+      self.report_error(constants.RESPONSE_INVALID_CALLER)
+      return
+    caller_id = caller_records[0].user_id
```
jiayliu: I think we should use the gcm id as the client id in the room. We'll need to know the call is initiated/accepted on which device for sending GCM notificaitons in the future. It's easy to get user ID from GCM ID, but not the other way.
I'm open to other options as long as we keep the information with the room, but using GCM id as the client id seems the easiest way.

samples/web/content/apprtc/apprtc_test.py
```diff
@@ -5,7 +5,9 @@
 
 import apprtc
 import constants
+import gcmrecord
 import json
+from google.appengine.datastore import datastore_stub_util
```
jiayliu: The changes in this file can be reverted now.

samples/web/content/apprtc/apprtc.py
chuckhays: Done, just hadn't committed it yet.

samples/web/content/apprtc/gcmrecord.py
```diff
@@ -28,19 +28,27 @@ class GCMRecord(ndb.Model):
   last_modified_time = ndb.DateTimeProperty()
 
   @classmethod
-  def get_by_user_id(cls, user_id):
-    q = GCMRecord.query(ancestor=get_ancestor_key())
-    return q.filter(GCMRecord.user_id == user_id).fetch()
+  def get_by_user_id(cls, user_id, verified = False):
```
jiayliu: This, and the other APIs that take 'verified', returns all records if verified == False, and returns verified records if verified == True, which may be not what one would expect from the interface. 
Maybe rename 'verified' to 'verified_only'.

samples/web/content/apprtc/apprtc.py
chuckhays: Seems reasonable, since it now supports a list I can just add all of the callee's gcm ids (and later only verified ones)

samples/web/content/apprtc/apprtc_test.py
jiayliu: This file also contains Join test cases. Move to the new file?

samples/web/content/apprtc/join_page_test.py
```diff
@@ -0,0 +1,287 @@
...
+
+  def makeGetRequest(self, path):
+    # PhantomJS uses WebKit, so Safari is closest to the thruth.
+    return self.test_app.get(path, headers={'User-Agent': 'Safari'})
+
+  def makePostRequest(self, path, body=''):
+    return self.test_app.post(path, body, headers={'User-Agent': 'Safari'})
+    
+  def addTestData(self):
+    gcmrecord.GCMRecord.add_or_update('caller1', 'caller1gcm1', 'caller1code1')
```
jiayliu: Put the data in lists and make the calls in a loop, to make the code more concise?

samples/web/content/apprtc/join_page_test.py
```diff
@@ -0,0 +1,287 @@
...
+    
+    room_id = 'callercallee'
+    # Caller that doesn't exist.
+    body = {
+      apprtc.PARAM_ACTION: 'call',
+      apprtc.PARAM_CALLER_GCM_ID: 'foo',
+      apprtc.PARAM_CALLEE_ID: 'callee1'
+    }
+    
+    response = self.makePostRequest('/join/' + room_id, json.dumps(body))
```
jiayliu: This pattern (body = ...; makePostRequest...; expect XXX) is used over and over again. Put it in a helper method?

samples/web/content/apprtc/join_page_test.py
```diff
@@ -0,0 +1,287 @@
...
+    
+    body = {
+      apprtc.PARAM_ACTION: 'accept',
+      apprtc.PARAM_CALLEE_GCM_ID: 'callee1gcm1'
+    }
+    
+    response = self.makePostRequest('/join/' + room_id, json.dumps(body))
+    self.verifyResultCode(response, constants.RESPONSE_SUCCESS)
+
+  def testJoinAsCalleeWrongCallee(self):
```
jiayliu: Needs some documentation to differentiate this from "joinAsCalleeInvalidCallee"

samples/web/content/apprtc/apprtc_test.py
chuckhays: I thought about it, but didn't because I was trying to make the file less long. They should probably be there though, I'll move them.

samples/web/content/apprtc/apprtc.py
```diff
@@ -366,12 +390,19 @@ def add_client_to_room(host, room_id, client_id, is_loopback):
...
+          room.add_allowed_client(allowed_client)
     else:
       is_initiator = False
       other_client = room.get_other_client(client_id)
       messages = other_client.messages
       room.add_client(client_id, Client(is_initiator))
       other_client.clear_messages()
+      # Check if the callee was the callee intended by the caller.
+      if not room.has_allowed_client(client_id):
+        error = constants.RESPONSE_INVALID_ROOM
```
jiayliu: We should not allow to "accept" for a room not started with "call", which indicates a client side bug.

chuckhays: That is an interesting case. I was trying to avoid explicitly marking the rooms, but it looks like we will have to in some way. Instead of using allowed_clients = None, we will have to  have a room type, or something similar.

jiayliu: Maybe change allowed_clients to called_clients?

On Tue, Jan 27, 2015 at 4:11 PM, Chuck Hays <notifications@github.com>
wrote:

> In samples/web/content/apprtc/apprtc.py
> <https://github.com/GoogleChrome/webrtc/pull/436#discussion_r23655829>:
>
> >      else:
> >        is_initiator = False
> >        other_client = room.get_other_client(client_id)
> >        messages = other_client.messages
> >        room.add_client(client_id, Client(is_initiator))
> >        other_client.clear_messages()
> > +      # Check if the callee was the callee intended by the caller.
> > +      if not room.has_allowed_client(client_id):
> > +        error = constants.RESPONSE_INVALID_ROOM
>
> That is an interesting case. I was trying to avoid explicitly marking the
> rooms, but it looks like we will have to in some way. Instead of using
> allowed_clients = None, we will have to have a room type, or something
> similar.
>
> —
> <https://github.com/GoogleChrome/webrtc/pull/436/files#r23655829>.
>



-- 
Jiayang Liu

chuckhays:
Updated to use gcm ids as the client id in rooms, and restricted joining rooms created via web vs direct calling. PTAL.

samples/web/content/apprtc/apprtc.py
```diff
@@ -495,16 +542,136 @@ def write_response(self, result, params, messages):
...
+    
+    # TODO (chuckhays): Initiate message to verified gcm ids to ring.
+    
+    self.write_room_parameters(
+        room_id, caller_id, result['messages'], result['is_initiator'])
+      
+    logging.info('User ' + caller_id + ' initiated call to ' + callee_id + \
+        ' from gcmId ' + caller_gcm_id)
+    logging.info('User ' + caller_id + ' joined room ' + room_id)
+    logging.info('Room ' + room_id + ' has state ' + result['room_state'])
```
jiayliu: nit: consolidate these two messages, i.e.
User ... joined room ... , room_state=...

samples/web/content/apprtc/apprtc.py
```diff
@@ -495,16 +542,136 @@ def write_response(self, result, params, messages):
...
+    
+    # TODO (chuckhays): Send message to ringing clients to stop ringing.
+    
+    self.write_room_parameters(
+        room_id, callee_id, result['messages'], result['is_initiator'])
+      
+    logging.info('User ' + callee_id + ' accepted call ' + \
+        'from gcmId ' + callee_gcm_id)
+    logging.info('User ' + callee_id + ' joined room ' + room_id)
+    logging.info('Room ' + room_id + ' has state ' + result['room_state'])
```
jiayliu: ditto consolidating logging

samples/web/content/apprtc/apprtc.py
```diff
@@ -495,16 +542,136 @@ def write_response(self, result, params, messages):
...
 
   def post(self, room_id):
+    # Check request body to determine what action to take.
+    msg = util.get_message_from_json(self.request.body)
+    if util.has_msg_field(msg, PARAM_ACTION, basestring):
+      action = msg[PARAM_ACTION]
+      if action == constants.ACTION_CALL:
```
jiayliu: Sorry I didn't make it clear: I think it's better to keep the join related constants in one place, either in constants, or at the beginning of this file. Either way is fine for me.

samples/web/content/apprtc/join_page_test.py
```diff
@@ -0,0 +1,299 @@
+# Copyright 2014 Google Inc. All Rights Reserved.
```
jiayliu: nit: 2015

samples/web/content/apprtc/apprtc.py
```diff
@@ -310,16 +311,38 @@ def __str__(self):
...
+      self.allowed_clients.append(client_id)
   def remove_client(self, client_id):
     del self.clients[client_id]
   def get_occupancy(self):
     return len(self.clients)
   def has_client(self, client_id):
     return client_id in self.clients
+  def is_client_allowed(self, client_id):
+    # If allowed_clients is None, don't enforce allowed_client list.
+    if self.allowed_clients is None:
```
tkchin: nit: can do:
return self.allowed_clients is None or client_id in self.allowed_clients

samples/web/content/apprtc/apprtc.py
```diff
@@ -361,17 +392,28 @@ def add_client_to_room(host, room_id, client_id, is_loopback):
       error = constants.RESPONSE_DUPLICATE_CLIENT
       break
 
+    if room.room_type != room_type:
+      error = constants.RESPONSE_INVALID_ROOM
```
tkchin: Can we add a warning log here? This particular error is should be at least extremely uncommon and invalid room isn't descriptive enough to debug what happened. (eg not allowed also returns invalid room)

samples/web/content/apprtc/apprtc.py
```diff
@@ -361,17 +392,28 @@ def add_client_to_room(host, room_id, client_id, is_loopback):
...
+          room.add_allowed_client(allowed_client)
     else:
       is_initiator = False
       other_client = room.get_other_client(client_id)
       messages = other_client.messages
       room.add_client(client_id, Client(is_initiator))
       other_client.clear_messages()
+      # Check if the callee was the callee intended by the caller.
+      if not room.is_client_allowed(client_id):
+        error = constants.RESPONSE_INVALID_ROOM
```
tkchin: ditto warning log

samples/web/content/apprtc/apprtc.py
```diff
@@ -495,16 +537,136 @@ def write_response(self, result, params, messages):
...
+    # Look up callee by id.
+    # TODO (chuckhays): Once registration is enabled, turn on restriction
+    # to only return verified records.
+    callee_records = gcmrecord.GCMRecord.get_by_user_id(callee_id, False)
+    if len(callee_records) < 1:
+      self.report_error(constants.RESPONSE_INVALID_CALLEE)
+      return
+
+    if has_room(self.request.host_url, room_id):
+      self.report_error(constants.RESPONSE_INVALID_ROOM)
```
tkchin: ditto warning

samples/web/content/apprtc/apprtc.py
```diff
@@ -495,16 +537,136 @@ def write_response(self, result, params, messages):
...
+    # TODO (chuckhays): Once registration is enabled, turn on restriction
+    # to only return verified records.
+    callee_records = gcmrecord.GCMRecord.get_by_gcm_id(callee_gcm_id, False)
+    if len(callee_records) < 1:
+      self.report_error(constants.RESPONSE_INVALID_CALLEE)
+      return
+
+    callee_id = callee_records[0].user_id
+    
+    if not has_room(self.request.host_url, room_id):
```
tkchin: Can probably move this before callee query. Also warning log.

samples/web/content/apprtc/apprtc.py
```diff
@@ -361,17 +392,28 @@ def add_client_to_room(host, room_id, client_id, is_loopback):
...
+        for allowed_client in allowed_clients:
+          room.add_allowed_client(allowed_client)
     else:
       is_initiator = False
       other_client = room.get_other_client(client_id)
       messages = other_client.messages
       room.add_client(client_id, Client(is_initiator))
       other_client.clear_messages()
+      # Check if the callee was the callee intended by the caller.
+      if not room.is_client_allowed(client_id):
```
tkchin: Probably move the allowed check earlier. Otherwise we nuke messages for invalid client.

chuckhays:
Updated with your suggestions @tkchin.

samples/web/content/apprtc/apprtc.py
chuckhays: We discussed this, it should be ok as is because we break out of the set loop.

tkchin:
lgtm

jiayliu:
lgtm. Please squash the commits before merge.

PR #441, opened 2015-01-28
Removing the need to install webtest-master.

phoglund:
I tried uninstalling webtest, and it actually seems it may not need installing since we add it to sys.path in run_python_tests.sh. I also made it clearer if one runs the .py instead of .sh the first time, when we need to download stuff.

The problem is that installing webtest needs easy_install, whose installation process is stupid and clunky. Let's hope we can avoid it. I'm also reluctant to require devs to run stuff like

wget https://bootstrap.pypa.io/ez_setup.py -O - | sudo python

which is recommended by the easy_install webpage.

### DIFF
--- a/run_python_tests.py
+++ b/run_python_tests.py
@@ -13,7 +13,17 @@ TEST_PATH    Path to package containing test modules.
 WEBTEST_PATH Path to the webtest library."""
 
 
+def _CheckDependencyInPlace(path):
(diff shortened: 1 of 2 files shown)


phoglund:
Never mind, it doesn't work because webtest doesn't get all its dependencies if we don't install it. Meh.

PR #443, opened 2015-01-29
Move SDP manipulating url params parsing to client

jiayliu:
Previously GAE parses the url params and simply passes them back to the JS, which adds no value.

Now the SDP manipulating params are completely handled on the client side.


jiayliu:
@juberti @tkchin

samples/web/content/apprtc/js/appcontroller.js
```diff
@@ -470,6 +472,25 @@ AppController.prototype.showIcons_ = function() {
   }
 };
 
+AppController.prototype.loadUrlParams_ = function() {
+  /* jshint ignore:start */
+  var queries = urlQueryToDictionary();
```
juberti: |queries| would probably be better named |urlParams|

samples/web/content/apprtc/js/appcontroller.js
```diff
@@ -470,6 +472,25 @@ AppController.prototype.showIcons_ = function() {
   }
 };
 
+AppController.prototype.loadUrlParams_ = function() {
+  /* jshint ignore:start */
```
juberti: Why is this needed?

samples/web/content/apprtc/js/util.js
```diff
@@ -18,6 +19,18 @@ function $(selector) {
...
 }
 
+// Returns the URL query key-value pairs as a dictionary object.
+function urlQueryToDictionary() {
+  var pairs = location.search.slice(1).split('&');
+
+  var result = {};
+  pairs.forEach(function(pair) {
+    pair = pair.split('=');
+    result[pair[0]] = decodeURIComponent(pair[1] || '');
```
juberti: should the default value be |true| instead of '', similar to HTML attributes?

samples/web/content/apprtc/js/util.js
```diff
@@ -18,6 +19,18 @@ function $(selector) {
   return document.querySelector(selector);
 }
 
+// Returns the URL query key-value pairs as a dictionary object.
+function urlQueryToDictionary() {
```
juberti: queryStringToDictionary

juberti:
nice cleanup, lgtm with nits above

chuckhays:
LGTM as well, with the same name comments as juberti.

samples/web/content/apprtc/js/util.js
jiayliu: This is to preserve the original GAE logic which uses '' as the default.

samples/web/content/apprtc/js/appcontroller.js
jiayliu: Added comment to explain.

tkchin:
lgtm

juberti:
lgtm

PR #448, opened 2015-01-29
Move Room/Client/RoomPage to separate files

chuckhays:
There are no functionality changes here, only moving Client, Room and RoomPage into separate files to separate things a bit better and reduce the size of apprtc.py.

The change is split into 2 commits so the moves are clearer if the combined view is hard to follow.

@tkchin @jiayliu


chuckhays:
Updated. PTAL.

samples/web/content/apprtc/room.py
```diff
@@ -0,0 +1,422 @@
...
+    wss_url = 'ws://' + ws_host_port_pair + '/ws'
+    wss_post_url = 'http://' + ws_host_port_pair
+  else:
+    wss_url = 'wss://' + ws_host_port_pair + '/ws'
+    wss_post_url = 'https://' + ws_host_port_pair
+  return (wss_url, wss_post_url)
+
+# Returns appropriate room parameters based on query parameters in the request.
+# TODO(tkchin): move query parameter parsing to JS code.
+def get_room_parameters(request, room_id, client_id, is_initiator):
```
jiayliu: I'd keep this one and the other parameter related methods in apprtc.py, since they really have nothing to do with Room. Or put them in a separate file.

samples/web/content/apprtc/room_test.py
```diff
@@ -20,21 +21,21 @@ def tearDown(self):
     self.testbed.deactivate()
 
   def testAllowedClientList(self):
-    room = apprtc.Room(apprtc.Room.TYPE_DIRECT)
+    clientRoom = room.Room(room.Room.TYPE_DIRECT)
```
jiayliu: nit: rename clientRoom

samples/web/content/apprtc/room.py
chuckhays: OK, I'll have to move them to a separate file or we would have a circular import.

chuckhays:
Update with your suggestions. PTAL.

PR #455, opened 2015-01-30
Python tests now return an error code on fail.

milkypostman:
This change will cause the Travis builds to fail when a Python test fails. So now we should be better prepared in our PRs.

### DIFF
--- a/run_python_tests.py
+++ b/run_python_tests.py
@@ -20,7 +20,7 @@ def main(sdk_path, test_path, webtest_path):
     sys.path.append(webtest_path)
     suite = unittest.loader.TestLoader().discover(test_path,
                                                   pattern="*test.py")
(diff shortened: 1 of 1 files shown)


milkypostman:
ack. @jiayliu @tkchin @chuckhays

jiayliu:
@phoglund

chuckhays:
LGTM.

jiayliu:
This broke travis, so I reverted it.

jiayliu:
Not sure why the PR cause the failure:
https://travis-ci.org/GoogleChrome/webrtc/builds/49382739

jiayliu:
Probably related to where the third_party path is set.
Donald, could you take a look?
The analytics tests are failing on Travis.

PR #456, opened 2015-01-30
Implement /decline endpoint.

chuckhays:
Implements the /decline endpoint and moves some test code into a shared based class to reduce repetition.

@tkchin @jiayliu

### DIFF
--- a/samples/web/content/apprtc/apprtc.py
+++ b/samples/web/content/apprtc/apprtc.py
@@ -8,23 +8,26 @@ This module demonstrates the WebRTC API by implementing a simple video chat app.
 """
 
 import cgi
-import client
-import constants
-import gcmrecord
-import gcm_register
-import join_page
 import logging
 import os
 import json
 import jinja2
-import parameter_handling
-import room as room_module
 import threading
 import urllib
-import util
 import webapp2
+
 from google.appengine.api import memcache
 from google.appengine.api import urlfetch
+
+import client
+import constants
+import decline_page
+import gcm_register
+import gcmrecord
+import join_page
+import parameter_handling
+import room as room_module
+import util
 
 jinja_environment = jinja2.Environment(
     loader=jinja2.FileSystemLoader(os.path.dirname(__file__)))
@@ -119,11 +122,12 @@ class ParamsPage(webapp2.RequestHandler):
 app = webapp2.WSGIApplication([
     ('/', MainPage),
     ('/bind/(\w+)', gcm_register.BindPage),
+    ('/decline/(\w+)', decline_page.DeclinePage),
+    ('/join/(\w+)', join_page.JoinPage),
     ('/leave/(\w+)/(\w+)', LeavePage),
     ('/message/(\w+)/(\w+)', MessagePage),
-    ('/join/(\w+)', join_page.JoinPage),
-    ('/r/(\w+)', RoomPage),
     ('/params', ParamsPage),
+    ('/r/(\w+)', RoomPage),
     # TODO(jiayl): Remove support of the old APIs when all clients are updated.
     ('/room/(\w+)', RoomPage),
     ('/register/(\w+)', join_page.JoinPage),
--- a/samples/web/content/apprtc/constants.py
+++ b/samples/web/content/apprtc/constants.py
@@ -7,6 +7,7 @@
 This module contains the constants used in AppRTC Python modules.
 """
 ROOM_MEMCACHE_EXPIRATION_SEC = 60 * 60 * 24
+ROOM_MEMCACHE_RETRY_LIMIT = 100
 
 LOOPBACK_CLIENT_ID = 'LOOPBACK_CLIENT_ID'
 
--- a/samples/web/content/apprtc/gcm_register_test.py
+++ b/samples/web/content/apprtc/gcm_register_test.py
@@ -1,46 +1,20 @@
 # Copyright 2014 Google Inc. All Rights Reserved.
 
+import json
 import unittest
 import webtest
+
+from google.appengine.datastore import datastore_stub_util
+from google.appengine.ext import ndb
+from google.appengine.ext import testbed
 
 import apprtc
 import constants
 import gcm_register
 import gcmrecord
-import json
-from google.appengine.datastore import datastore_stub_util
-from google.appengine.ext import ndb
-from google.appengine.ext import testbed
+import test_utilities
 
-class BindPageHandlerTest(unittest.TestCase):
-  def setUp(self):
-    # First, create an instance of the Testbed class.
-    self.testbed = testbed.Testbed()
-    # Then activate the testbed, which prepares the service stubs for use.
-    self.testbed.activate()
-
-    # Create a consistency policy that will simulate the High Replication
-    # consistency model.
-    self.policy = datastore_stub_util.PseudoRandomHRConsistencyPolicy()
-    # Initialize the datastore stub with this policy.
-    self.testbed.init_datastore_v3_stub(consistency_policy=self.policy)
-    self.testbed.init_memcache_stub()
-
-    self.test_app = webtest.TestApp(apprtc.app)
-
-  def tearDown(self):
-    self.testbed.deactivate()
-
-  def makePostRequest(self, path, body=''):
-    return self.test_app.post(path, body, headers={'User-Agent': 'Safari'})
-
-  def checkInvalidRequests(self, path, params):
-    body = {x: '' for x in params}
-    while len(body) > 0:
-      response = self.makePostRequest(path, json.dumps(body))
-      self.assertEqual(constants.RESPONSE_INVALID_ARGUMENT, response.body)
-      body.popitem()
-
+class BindPageHandlerTest(test_utilities.BasePageHandlerTest):
   def testBindNew(self):
     body = {
       gcm_register.PARAM_USER_ID: 'foo',
--- a/samples/web/content/apprtc/room.py
+++ b/samples/web/content/apprtc/room.py
@@ -7,29 +7,38 @@
 This module implements the room data model.
 """
 
-import client
-import constants
 import json
 import logging
+
+from google.appengine.api import memcache
+
+import client
+import constants
 import parameter_handling
 import util
-from google.appengine.api import memcache
 
 class Room:
   # Room used for simple named rooms, can be joined by knowing the room name.
   TYPE_OPEN = 1
   # Room used for direct calling, can only be joined by allowed clients.
   TYPE_DIRECT = 2
+
+  STATE_EMPTY = 0
+  STATE_WAITING = 1
+  STATE_FULL = 2
   
   def __init__(self, room_type):
     self.clients = {}
     # The list of allowed clients will include the initiator and other
     # clients allowed to join the room. If no clients are added, no
     # list is enforced, and is_client_allowed will always return true.
-    self.allowed_clients = None;
+    self.allowed_clients = None
     if room_type != Room.TYPE_OPEN and room_type != Room.TYPE_DIRECT:
       room_type = Room.TYPE_OPEN
     self.room_type = room_type
+  def reset(self):
+    self.clients = {}
+    self.allowed_clients = None
   def add_client(self, client_id, client):
     self.clients[client_id] = client
   def add_allowed_client(self, client_id):
@@ -53,6 +62,13 @@ class Room:
       if key is not client_id:
         return client
     return None
+  def get_room_state(self):
+    occupancy = self.get_occupancy()
+    if occupancy == 0:
+      return Room.STATE_EMPTY
+    elif occupancy == 1:
+      return Room.STATE_WAITING
+    return Room.STATE_FULL
   def __str__(self):
     return str(self.clients.keys())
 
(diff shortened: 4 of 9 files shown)


samples/web/content/apprtc/room.py
```diff
@@ -65,6 +65,27 @@ def has_room(host, room_id):
   room = memcache_client.get(key)
   return room is not None
 
+def remove_room(host, room_id, client_id):
+  memcache_client = memcache.Client()
```
jiayliu: This needs to use cas to avoid race.

samples/web/content/apprtc/room.py
```diff
@@ -65,6 +65,27 @@ def has_room(host, room_id):
...
+  
+  if room is None:
+    logging.warning('Can\'t remove room ' + room_id +
+        ' because it doesn\'t exist, client_id: ' + client_id)
+    return constants.RESPONSE_INVALID_ROOM
+
+  if not room.is_client_allowed(client_id):
+    logging.warning('Can\'t remove room ' + room_id +
+        ' because room does not allow client ' + client_id)
+    return constants.RESPONSE_INVALID_CALLEE
```
jiayliu: I think we need to also check room type to be direct.

samples/web/content/apprtc/test_utilities.py
```diff
@@ -0,0 +1,100 @@
...
+import apprtc
+import constants
+import gcm_register
+import gcmrecord
+import json
+from google.appengine.datastore import datastore_stub_util
+from google.appengine.ext import ndb
+from google.appengine.ext import testbed
+
+class BasePageHandlerTest(unittest.TestCase):
```
jiayliu: Nice improvement :)

samples/web/content/apprtc/room.py
jiayliu: Also need to check that the room does not already have two clients, i.e. call already accepted.

samples/web/content/apprtc/apprtc.py
```diff
@@ -10,6 +10,7 @@
 import cgi
 import client
 import constants
+import decline_page
```
jiayliu: Will it be better to group the local module together, separated from the built-in/third_party modules?

chuckhays: I couldn't figure out if we had a set style for this. I was doing alphabetical in absence of another standard. Happy to do it however we think is appropriate.

samples/web/content/apprtc/room.py
chuckhays: Would we use the cas method to set the value to None, or is there a version of delete that is aware of the cas type update?

samples/web/content/apprtc/room.py
chuckhays: I thought this seemed too simple. Good catch.

samples/web/content/apprtc/room.py
jiayliu: There is no compare-and-delete, so we can only set None, like the existing delete_client_from_room.

chuckhays: There is actually an interesting race condition either way.

Client A calls add_client_to_room
Before A calls .cas, B calls remove_room and deletes the value (either through delete or cas set to None)
Client A fails .cas, retrieves the key again, room is None, so it recreates the room

Maybe we should mark the room as deleted on the room object, and either clean them out periodically or just let memcache evict them.

jiayliu: Is A the caller B the callee?

B's decline should fail if the room is None or has no caller.

Or do you mean A and B are both callee?

Then A cannot recreate the room as a callee, its "accept" should fail.


On Fri, Jan 30, 2015 at 12:53 PM, Chuck Hays <notifications@github.com>
wrote:

> In samples/web/content/apprtc/room.py
> <https://github.com/GoogleChrome/webrtc/pull/456#discussion_r23871510>:
>
> > @@ -65,6 +65,27 @@ def has_room(host, room_id):
> >    room = memcache_client.get(key)
> >    return room is not None
> >
> > +def remove_room(host, room_id, client_id):
> > +  memcache_client = memcache.Client()
>
> There is actually an interesting race condition either way.
>
> Client A calls add_client_to_room
> Before A calls .cas, B calls remove_room and deletes the value (either
> through delete or cas set to None)
> Client A fails .cas, retrieves the key again, room is None, so it
> recreates the room
>
> Maybe we should mark the room as deleted on the room object, and either
> clean them out periodically or just let memcache evict them.
>
> —
> <https://github.com/GoogleChrome/webrtc/pull/456/files#r23871510>.
>



-- 
Jiayang Liu

chuckhays: I hadn't thought about it in terms of the actual caller/callee, just the race condition in the code.

I think A and B would both be callees. A is from accept, B is from decline.

I'll think this through a little more. It might make sense to move the Room creation to a more explicit call instead of being a side effect of the add_client update process.

jiayliu: A should not create a new room if A is a callee.

On Fri, Jan 30, 2015 at 1:12 PM, Chuck Hays <notifications@github.com>
wrote:

> In samples/web/content/apprtc/room.py
> <https://github.com/GoogleChrome/webrtc/pull/456#discussion_r23872714>:
>
> > @@ -65,6 +65,27 @@ def has_room(host, room_id):
> >    room = memcache_client.get(key)
> >    return room is not None
> >
> > +def remove_room(host, room_id, client_id):
> > +  memcache_client = memcache.Client()
>
> I hadn't thought about it in terms of the actual caller/callee, just the
> race condition in the code.
>
> I think A and B would both be callees. A is from accept, B is from decline.
>
> I'll think this through a little more. It might make sense to move the
> Room creation to a more explicit call instead of being a side effect of the
> add_client update process.
>
> —
> <https://github.com/GoogleChrome/webrtc/pull/456/files#r23872714>.
>



-- 
Jiayang Liu

chuckhays: I agree.

chuckhays: How can you use cas to do delete? If my understanding is correct, calling gets adds a timestamp like id to the object returned that is then used in the cas call to be sure the underlying record hasn't changed, but passing None as the value won't have that timestamp on it.

It seems like the only way we can use cas is to do a soft delete, with a deleted flag. 

Am I missing something?

jiayliu: I thought it's done through some sort of = operator overriding, so that value=None will keep the timestamp.
But honestly I'm not quite sure.
According to http://neopythonic.blogspot.com/2011/08/compare-and-set-in-memcache.html, it works for value+1, which indicates + operator overriding (?). But it could be different for '='.
Anyway, it might be a good idea to invalidate the room, to be safe, e.g. set the room id to None, and let it be evicted when expired.

chuckhays: Interesting, assigning None to the value and using the variable works with cas. Maybe there is a special case for that path.

samples/web/content/apprtc/apprtc.py
```diff
@@ -120,6 +121,7 @@ def get(self):
     ('/', MainPage),
     ('/bind/(\w+)', gcm_register.BindPage),
     ('/leave/(\w+)/(\w+)', LeavePage),
+    ('/decline/(\w+)', decline_page.DeclinePage),
```
tkchin: nit: maybe make these alphabetical sans the old APIs

samples/web/content/apprtc/apprtc.py
tkchin: I think it makes sense to follow the guide's import order.

samples/web/content/apprtc/gcm_register_test.py
```diff
@@ -8,39 +8,12 @@
...
-    return self.test_app.post(path, body, headers={'User-Agent': 'Safari'})
-
-  def checkInvalidRequests(self, path, params):
-    body = {x: '' for x in params}
-    while len(body) > 0:
-      response = self.makePostRequest(path, json.dumps(body))
-      self.assertEqual(constants.RESPONSE_INVALID_ARGUMENT, response.body)
-      body.popitem()
-
+class BindPageHandlerTest(test_utilities.BasePageHandlerTest):
```
tkchin: Just out of curiosity - what's the difference between unittest.TestCase and test_utilities.BasePageHandlerTest?

tkchin: Nevermind, got to the bottom of the review :)

samples/web/content/apprtc/test_utilities.py
tkchin: +1

chuckhays:
Updated and merged. PTAL.

samples/web/content/apprtc/room.py
```diff
@@ -65,8 +81,63 @@ def has_room(host, room_id):
...
+  memcache_client = memcache.Client()
+  return memcache_client.get(key)
+
+def get_room_state(host, room_id):
+  room = get_room(host, room_id)
+  if room is None:
+    return None
+  return room.get_room_state()
+
+def remove_room(host, room_id, client_id):
```
jiayliu: nit: The name can be more descriptive, e.g. remove_room_for_declined_call, and s/client_id/callee_gcm_id

samples/web/content/apprtc/room.py
```diff
@@ -65,8 +81,63 @@ def has_room(host, room_id):
...
+  retries = 0
+  while retries < constants.ROOM_MEMCACHE_RETRY_LIMIT:
+    room = memcache_client.gets(key)
+    
+    if room is None:
+      logging.warning('Can\'t remove room ' + room_id +
+          ' because it doesn\'t exist, client_id: ' + client_id)
+      return constants.RESPONSE_INVALID_ROOM
+  
+    if not room.is_client_allowed(client_id):
```
jiayliu: I think we should also check client_id is not the caller

samples/web/content/apprtc/constants.py
```diff
@@ -7,6 +7,7 @@
 This module contains the constants used in AppRTC Python modules.
 """
 ROOM_MEMCACHE_EXPIRATION_SEC = 60 * 60 * 24
+ROOM_MEMCACHE_RETRY_LIMIT = 5
```
jiayliu: 5 feels too small. How about 100?

jiayliu: And should this apply to other cas loops?

chuckhays:
Updated per suggestions.

jiayliu:
lgtm. Thx!

samples/web/content/apprtc/room.py
```diff
@@ -65,8 +81,70 @@ def has_room(host, room_id):
...
+  if room is None:
+    return None
+  return room.get_room_state()
+
+def remove_room_for_declined_call(host, room_id, callee_gcm_id):
+  memcache_client = memcache.Client()
+  key = get_memcache_key_for_room(host, room_id)
+  
+  retries = 0
+  while retries < constants.ROOM_MEMCACHE_RETRY_LIMIT:
```
tkchin: nit: can avoid having to manually increment by doing:
for retries in xrange(constants.ROOM_MEMCACHE_RETRY_LIMIT):

samples/web/content/apprtc/room.py
```diff
@@ -74,18 +152,24 @@ def add_client_to_room(host, room_id, client_id,
   retries = 0
   room = None
   # Compare and set retry loop.
-  while True:
+  while retries < constants.ROOM_MEMCACHE_RETRY_LIMIT:
```
tkchin: ditto nit

samples/web/content/apprtc/room.py
```diff
@@ -138,7 +222,7 @@ def remove_client_from_room(host, room_id, client_id):
   memcache_client = memcache.Client()
   retries = 0
   # Compare and set retry loop.
-  while True:
```
tkchin: ditto nit here and below

tkchin:
lgtm

samples/web/content/apprtc/room.py
```diff
@@ -65,8 +81,70 @@ def has_room(host, room_id):
...
+    if memcache_client.cas(key, room, constants.ROOM_MEMCACHE_EXPIRATION_SEC):
+      logging.info('Reset room %s to base state in remove_room by client %s, retries = %d' \
+          %(room_id, callee_gcm_id, retries))
+      return constants.RESPONSE_SUCCESS
+    else:
+      retries = retries + 1
+    
+    logging.warning('Failed to remove room ' + room_id + ' after retry limit ' +
+        ' client: ' + callee_gcm_id)
+    return constants.RESPONSE_INTERNAL_ERROR
```
tkchin: Indentation is hard to read on CR tool - is this supposed to be two left outside the loop?

chuckhays: Good eye. I also realized I should add return statements to a couple of the loops that were previously infinite loops.

samples/web/content/apprtc/apprtc.py
```diff
@@ -8,24 +8,27 @@
...
-import util
 import webapp2
+
 from google.appengine.api import memcache
 from google.appengine.api import urlfetch
 
+import client
+import constants
+import decline_page
+import gcmrecord
```
tkchin: >>> sorted(['gcmrecord', 'gcm_register'])
['gcm_register', 'gcmrecord']

tkchin:
lgtm. Wish we could unit test cas loops though, but I don't know how we'd do that.

PR #460, opened 2015-02-03
Remove deprecated GAE API

jiayliu:


jiayliu:
@juberti @chuckhays

juberti:
lgtm

Issue #457, opened 2015-01-30
Toolbar buttons don't show after rejoining room


juberti:
If you leave and then rejoin a room via the 'rejoin' button, the toolbar buttons never show up.

chuckhays:
This was the result merging two features working in the same area. PR https://github.com/GoogleChrome/webrtc/pull/432 refactored code in a way that no longer worked in the way assumed by https://github.com/GoogleChrome/webrtc/pull/424.

PR #462, opened 2015-02-03
Fix for #457, rejoin is broken.

chuckhays:
Previously calling .start() triggered the media promise to be run, but it no longer does, so restart needs to do it. All of the media attach events happen as a result of that promise executing.

We probably could skip doing the turn servers promise, but I was just trying to restore the previous functionality and not optimizing it.
@jiayliu @juberti

### DIFF
--- a/samples/web/content/apprtc/js/call.js
+++ b/samples/web/content/apprtc/js/call.js
@@ -40,12 +40,18 @@ var Call = function(params) {
   this.onsignalingstatechange = null;
   this.onstatusmessage = null;
 
-  this.getMediaPromise_ = this.maybeGetMedia_();
-  this.getTurnServersPromise_ = this.maybeGetTurnServers_();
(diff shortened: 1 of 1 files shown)


jiayliu:
Unit test?

chuckhays:
I'm happy to add a test, what do you think I should test here? That restart calls the init method?

jiayliu:
I think we should test that what was broken is now fixed, i.e. restart
should re-request user media.

On Tue, Feb 3, 2015 at 11:17 AM, Chuck Hays <notifications@github.com>
wrote:

> I'm happy to add a test, what do you think I should test here? That
> restart calls the init method?
>
> —
> <https://github.com/GoogleChrome/webrtc/pull/462#issuecomment-72715046>.
>



-- 
Jiayang Liu

juberti:
Assuming obsoleted by repo move

PR #466, opened 2015-02-03
Fix failing Python tests.

milkypostman:
This is what caused #464. Looks like b806466 broke this but since Python tests don't error Travis we didn't see it.


milkypostman:
IT WORKS! You should be able to merge this and then merge #465 and Python tests will now cause PRs to fail on travis.

juberti:
assuming this is obsoleted by repo move

PR #468, opened 2015-02-04
Clean out samples repo, fix links

juberti:
@jiayl, @chuckhays PTAL


jiayliu:
Clean up package.json and .gitignore?

juberti:
Cleaned package.json and .gitignore

KaptenJansson:
I removed package.json, there is not point for the samples to use this right?

KaptenJansson:
Also not that a lot of files that are excluded for JSCS but I guess that's due to the files not adhering to the styleguide yet, will file a separate issue for that.

jiayliu:
The benefit of package.json I think is that one can simply 'npm install' to install all the dependencies, without worrying about which exact libraries are needed.
That's why I kept it for apprtc.

juberti:
Agree with Jiayang. package.json restored, Travis passes.

juberti:
PTAL

PR #478, opened 2015-02-17
icetransports is a configuration, not a constraint

fippo:
Recently stumbled over this, thanks @marwahvikas for pointing out the right way to do it.


marwahvikas:
Thanks Philipp..LGTM.

PR #481, opened 2015-02-22
add filetransfer demo

fippo:
I was surprised by the lack of a filetransfer demo... let's provide one so people do not use examples that still use rtp data channels...

Performance seems to be between 15 and 20mbit/sec which is is somewhat lower than the 100mbps quoted in https://code.google.com/p/webrtc/issues/detail?id=2276 but it's a start.
This is actually quite interesting to watch with the new datachannel stats. Don't switch away during the filetransfer however. And remove the closeDataChannels call or get someone to merge https://code.google.com/p/chromium/issues/detail?id=416876 :-))

@jesup: TYMLTK -- works in firefox as well (setting .binaryType to arraybuffer)


fippo:
looking at those progress bars in chrome and firefox is quite interesting actually. Firefox is sending way faster than Chrome (which is almost sync)

samdutton:
Excellent – thank you! Will take a look.

src/content/datachannel/filetransfer/index.html
```diff
@@ -0,0 +1,58 @@
...
+  <link rel="stylesheet" href="css/main.css" />
+
+</head>
+
+<body>
+
+  <div id="container">
+
+    <h1><a href="https://googlechrome.github.io/webrtc/" title="WebRTC samples homepage">WebRTC samples</a> <span>RTCDataChannel Filetransfer</span></h1>
+
```
samdutton: file transfer

src/content/datachannel/filetransfer/index.html
```diff
@@ -0,0 +1,58 @@
...
+  <link rel="stylesheet" href="../../../css/main.css">
+  <link rel="stylesheet" href="css/main.css" />
+
+</head>
+
+<body>
+
+  <div id="container">
+
+    <h1><a href="https://googlechrome.github.io/webrtc/" title="WebRTC samples homepage">WebRTC samples</a> <span>RTCDataChannel Filetransfer</span></h1>
```
samdutton: Link should be to https://webrtc.github.io/samples/

(Just fixed this elsewhere.)

src/content/datachannel/filetransfer/index.html
```diff
@@ -0,0 +1,58 @@
...
+
+    <form id="fileInfo">
+      <input type="file" id="fileInput" name="files"/>
+    </form>
+
+    Send progress: <progress id="sendProgress" max="0" value="0"></progress>
```
samdutton: Could you add a little more space (margin: 0 0 1em 0) between each or these lines? If you can be bothered, maybe put the progress elements on the same line as the label text, like http://webrtc.github.io/samples/src/content/getusermedia/volume.

src/content/datachannel/filetransfer/js/main.js
```diff
@@ -0,0 +1,171 @@
...
+ *
+ *  Use of this source code is governed by a BSD-style license
+ *  that can be found in the LICENSE file in the root of the source
+ *  tree.
+ */
+
+'use strict';
+
+var localConnection, remoteConnection, sendChannel, receiveChannel,
+  pcConstraint;
```
samdutton: We've just started using JSCS validation with `"preset": "google"`. This adds a few validation rules such as only allowing one var declaration per line.

Could you add this and make the changes? There are less than 10, and they're all pretty trivial. 

FWIW there's a JSCS plugin for Sublime Text, if that helps.

samdutton:
LGTM % nits.

Thanks again for doing this.

Will also ask Benjamin Schwartz to take a look.

bemasc:
:+1:

fippo:
All done... grunt seemed happy, but might be ignoring the file?

Now I just need to figure out what's the best way to merge the move of the basic sample with your changes. Probably rebase.

fippo:
amazing, merged without problems. And grunt started being unhappy but that was easy to fix.

samdutton:
LGTM

fippo:
awesome.

Would be good to showcase the recently added datachannel stats there when https://c
ode.google.com/p/webrtc/issues/detail?id=4321 lands -- even though in the meantime the connection bitrate should be sufficient. I'll see if can do that as well, shouldn't be hard.

juberti:
lgtm, cool!

PR #483, opened 2015-02-25
filetransfer: show bitrate stats

fippo:
followup for #481 @samdutton

Bitrate calculation code borrowed from the constraints-and-stats demo (which looks like it needs a cleanup...)

### DIFF
--- a/src/content/datachannel/filetransfer/index.html
+++ b/src/content/datachannel/filetransfer/index.html
@@ -57,6 +57,7 @@
         <div class="label">Receive progress: </div>
         <progress id="receiveProgress" max="0" value="0"></progress>
       </div>
+
       <div id="bitrate"></div>
 
       <a id="received"></a>
--- a/src/content/datachannel/filetransfer/index.html
+++ b/src/content/datachannel/filetransfer/index.html
@@ -43,6 +43,7 @@
       <div class="label">Receive progress: </div>
       <progress id="receiveProgress" max="0" value="0"></progress>
     </div>
+    <div id="bitrate"></div>
 
     <a id="received"></a>
 
--- a/src/content/datachannel/filetransfer/index.html
+++ b/src/content/datachannel/filetransfer/index.html
@@ -35,7 +35,7 @@
 
       <p>To accomplish this in an interoperable way, the file is split into chunks which are then transferred via the datachannel. The datachannel is reliable and ordered by default which is well-suited to filetransfers.</p>
 
-      <p>Send and receive progress is monitored using HTML5 <i>progress</i> elements.
+      <p>Send and receive progress is monitored using HTML5 <i>progress</i> elements.</p>
       
       <p>At the receiver, the file is reassembled using the Blob API and made available for download.</p>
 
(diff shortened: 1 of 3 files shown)


samdutton:
Nice one! 

I noticed that the bitrate won't display for files that transfer in less than a second (the setInterval interval). 

Reducing the interval to 100ms or less means that the bitrate gets displayed, but only for as long as the file takes to transfer. 

Might be worth displaying bitrate in a separate div, and displaying the average (or something) once the transfer has completed.

Also noticed that sometimes (with a shorter interval at least) bitrate is NaN – not sure why. As a workaround, could check for that.

fippo:
How about average bitrate calculated as file.size / time-it-takes-to-transfer? That also works in Firefox (which doesn't support sent/recv for datachannels currently -- https://bugzilla.mozilla.org/show_bug.cgi?id=1136832)

The NaN might happen when getStats internally did not update yet since the calculation is based on the stats reports timestamp. Since bytesReceived wasn't updated I'm just skipping when that happens.

You made me wonder how hard it would be to rip out the bitrate drawing code webrtc-internals uses. Even though I prefer highcharts for drawing that

samdutton:
>
> How about average bitrate calculated as file.size /
> time-it-takes-to-transfer?


Perfect.

BTW - do you want to add a sentence or two that explains what's happening
in the demo?

 bitrate drawing code webrtc-internals uses


Tempting, though it will make the code a bit more complex.

fippo:
done, WDYT?

If I had time I would write something about the different choices of signaling the file metadata... for small files, in-band signaling makes sense, for larger files you want to use the signaling channel etc. Looks like I need to find a conference that lets me talk about this :-)

Also found a way to display the bitrate for Firefox. Because demos that work in all browsers are the web we want :-)

src/content/datachannel/filetransfer/index.html
```diff
@@ -29,28 +29,46 @@
...
+      <div class="progress">
+        <div class="label">Send progress: </div>
+        <progress id="sendProgress" max="0" value="0"></progress>
+      </div>
+
+      <div class="progress">
+        <div class="label">Receive progress: </div>
+        <progress id="receiveProgress" max="0" value="0"></progress>
+      </div>
+      <div id="bitrate"></div>
```
samdutton: Add a line space above this in the code – and with CSS for the layout :).

samdutton:
Nice.

LGTM aside from nitpick.

Could maybe display the bitrate every 100ms - I only saw it flash up once
with a ~6MB file.

fippo:
done. 100ms looked too aggressive to me. Changed to 500ms and and added additional display steps at 100ms and 300ms for small files :-)

juberti:
This would probably work great with my long-suffering PR for drawing arbitrary stats graphs.

fippo:
http://fippo.github.io/webrtc/src/content/datachannel/filetransfer/ :-)

ashumeow:
@fippo Looks awesome! :+1:

samdutton:
Fantastic! Merged.

samdutton:
[webrtc.github.io/samples/src/content/datachannel/filetransfer](http://webrtc.github.io/samples/src/content/datachannel/filetransfer/)

Duly tweeted, G+ posted, and included in forthcoming blog post about the samples repo.

fippo:
And linked in a comment at https://bloggeek.me/send-file-webrtc-data-api/ -- which I suspect is the reason why people still attempt to use RTP datachannels, ignoring that the advice there is from 2013 :-)

PR #484, opened 2015-02-26
Add polyfill for getStats in adapter.js.

milkypostman:
The Chrome version of getStats produces a cryptic data structure. This
wraps the getStats call so that it comes out in the W3C standard
structure.

### DIFF
--- a/src/js/adapter.js
+++ b/src/js/adapter.js
@@ -1,5 +1,5 @@
 /*
- *  Copyright (c) 2014 The WebRTC project authors. All Rights Reserved.
+ *  Copyright (c) 2014-2015 The WebRTC project authors. All Rights Reserved.
  *
  *  Use of this source code is governed by a BSD-style license
  *  that can be found in the LICENSE file in the root of the source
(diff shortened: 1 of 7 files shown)


src/js/adapter_test.js
```diff
@@ -0,0 +1,82 @@
...
+ *
+ *  Use of this source code is governed by a BSD-style license
+ *  that can be found in the LICENSE file in the root of the source
+ *  tree.
+ */
+
+var AdapterTest = new TestCase('AdapterTest');
+
+AdapterTest.prototype.setUp = function() {
+  window.console.log("Setting stuff up!");
```
juberti: I assume you meant to remove this?

src/js/test_preload.js
```diff
@@ -0,0 +1,21 @@
+/*
+ *  Copyright (c) 2015 The WebRTC project authors. All Rights Reserved.
+ *
+ *  Use of this source code is governed by a BSD-style license
+ *  that can be found in the LICENSE file in the root of the source
+ *  tree.
+ */
+
+/* This file is used to simply extract the Chrome bits somewhere we can test. */
```
juberti: Please dumb this down to I can understand what is happening here

juberti: Please dumb this down so I can understand what is happening here

src/js/adapter.js
```diff
@@ -181,7 +181,33 @@ if (navigator.mozGetUserMedia) {
 
   // The RTCPeerConnection object.
   RTCPeerConnection = function(pcConfig, pcConstraints) {
-    return new webkitRTCPeerConnection(pcConfig, pcConstraints);
+    var pc = new webkitRTCPeerConnection(pcConfig, pcConstraints);
+    var origGetStats = pc.getStats.bind(pc);
+    pc.getStats = function(selector, successCallback) {
```
juberti: To avoid breaking all stats users, should we just leave the stats untouched if the old argument order was used?

milkypostman: Hrm. That's a thought. It's going to ugly up the code a bit because we'll have to do type checking on the arguments to see if one is a function or not. Another option would be to keep the old function under a different name.

milkypostman:
@juberti have addressed comments and reworked the testing setup. This is now specialized to Chrome but allows us to add Firefox tests in the future. Each needs it own *setup* because we are essentially faking out the differing parts.

KaptenJansson:
milkypostman@ when do you think you can move this to the adapter repo?

milkypostman:
i just need to carve out time to do it again. i'll try to get this going.

milkypostman:
opened new pr in webrtc/adapter#46

PR #486, opened 2015-02-27
filetransfer: blow away all previous state and restart

fippo:
as suggested in https://github.com/webrtc/samples/issues/485#issuecomment-76434489


fippo:
wait a second, I should probably do what I am talking about re revokeObjectURL...

PR #487, opened 2015-03-03
remove rtp data channels from samples

fippo:
I think the samples should not show rtp based data channels anymore. There is also a lack of Firefox detection in the sample so removing this is the easiest fix.

also removes code for chrome < 31 since adapter.js doesnt care about versions that old either


samdutton:
LGTM.

@juberti @alvestrand OK?

alvestrand:
LGTM. Exposing the RTP datachannel through the datachannel interface was a get-something-out hack.

juberti:
lgtm

PR #488, opened 2015-03-05
candidate gathering demo: try to detect authentication failures and unreachable servers

fippo:
that demo is actually pretty useful when helping people to debug their TURN servers. This makes it slightly more useful by doing a little bit of cause analysis. Or at least easier to explain.

@juberti does it make your eyes hurt enough that I get the turn server event listener faster? Oh wait, this actually solves ~50% of the issue at least so no hurry.

### DIFF
--- a/src/content/peerconnection/trickle-ice/js/main.js
+++ b/src/content/peerconnection/trickle-ice/js/main.js
@@ -25,6 +25,7 @@ removeButton.onclick = removeServer;
 
 var begin;
 var pc;
+var candidates;
 
 function selectServer(event) {
   var option = event.target;
@@ -105,6 +106,7 @@ function start() {
 
 function gotDescription(desc) {
   begin = window.performance.now();
+  candidates = [];
   pc.setLocalDescription(desc);
 }
 
@@ -150,6 +152,46 @@ function appendCell(row, val, span) {
   row.appendChild(cell);
 }
 
+// Try to determine authentication failures and unreachable TURN
+// servers by using heuristics on the candidate types gathered.
+function getFinalResult() {
+  var result = 'Done';
+
+  // if more than one server is used, it can not be determined
+  // which server failed.
+  if (servers.length === 1) {
+    var server = JSON.parse(servers[0].value);
+
+    // get the candidates types (host, srflx, relay)
+    var types = candidates.map(function(cand) {
+      return cand.type;
+    });
+
+    // If the server is a TURN server we should have a relay candidate.
+    // If we did not get a relay candidate but a srflx candidate
+    // authentication might have failed.
+    // If we did not get  a relay candidate or a srflx candidate
+    // we could not reach the TURN server. Either it is not running at
+    // the target address or the clients access to the port is blocked.
+    //
+    // This only works for TURN/UDP since we do not get
+    // srflx candidates from TURN/TCP.
+    if (server.url.indexOf('turn:') === 0 &&
+        server.url.indexOf('?transport=tcp') === -1) {
+      if (types.indexOf('relay') === -1) {
+        if (types.indexOf('srflx') > -1) {
+          // a binding response but no relay candidate suggests auth failure.
+          result = 'Authentication failed?';
+        } else {
+          // either the TURN server is down or the clients access is blocked.
+          result = 'Not reachable?';
+        }
+      }
+    }
+  }
+  return result;
+}
+
 function iceCallback(event) {
   var elapsed = ((window.performance.now() - begin) / 1000).toFixed(3);
   var row = document.createElement('tr');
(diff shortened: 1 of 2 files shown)


fippo:
hrm... based on some feedback I just got I wonder whether there should also be detection of relay candidates with private ips (TURN server on ec2) :-)

src/content/peerconnection/trickle-ice/js/main.js
```diff
@@ -164,7 +164,33 @@ function iceCallback(event) {
     appendCell(row, c.port);
     appendCell(row, formatPriority(c.priority));
   } else {
-    appendCell(row, 'Done', 7);
+    var result = 'Done';
+    if (servers.length === 1) {
+      var server = JSON.parse(servers[0].value);
+
+      // figure out the types of candidates (host, srflx, relay)
+      var types = pc.localDescription.sdp.split('\r\n').filter(
```
juberti: This is pretty gnarly. I recommend breaking this out into its own getFinalResult function, as well as storing the candidate objects in a global list so that you don't have to process SDP.

src/content/peerconnection/trickle-ice/js/main.js
```diff
@@ -164,7 +164,33 @@ function iceCallback(event) {
     appendCell(row, c.port);
     appendCell(row, formatPriority(c.priority));
   } else {
-    appendCell(row, 'Done', 7);
+    var result = 'Done';
+    if (servers.length === 1) {
```
juberti: Explain why we need to limit to 1 server

src/content/peerconnection/trickle-ice/js/main.js
```diff
@@ -164,7 +164,33 @@ function iceCallback(event) {
...
+      var types = pc.localDescription.sdp.split('\r\n').filter(
+        function(line) {
+          return line.indexOf('a=candidate:') === 0;
+        }
+      ).map(function(cand) {
+        return parseCandidate(cand).type;
+      });
+
+      // this trick only works for TURN/UDP
+      if (server.url.indexOf('turn:') === 0 &&
```
juberti: Should still be able to detect some sort of failure for TURN/TCP, right?

src/content/peerconnection/trickle-ice/js/main.js
fippo: if used with more than one server, it can not be determined which server failed

juberti: true, but worth mentioning in a comment

fippo: good point. done in working copy.

src/content/peerconnection/trickle-ice/js/main.js
fippo: that's easier even and avoids double-parsing. thanks.

src/content/peerconnection/trickle-ice/js/main.js
fippo: only that no candidate could be gathered. Since we don't get STUN candidates from a TURN/TCP server it's not possible to decide whether that is due to a unreachable port or  an authentication failure.

Typically I would expect people to test authentication using TURN/UDP and then just reuse those credentials on TURN/TCP.

fippo:
I think that documents the heuristics better. Thanks!

fippo:
Interesting... this doesn't work in Firefox since it doesn't send binding requests to the TURN server.

KaptenJansson:
fippo@ what's the plan here? Make it work with FF?

fippo:
@KaptenJansson I typically tell people to debug TURN server issues using this demo and if the demo shows an authentication failure nicely that makes their life easier. Chrome-only feature, but it's not essential so...

KaptenJansson:
I just tried this with a turn server (provided by https://computeengineondemand.appspot.com) with the wrong password and I get nothing, just some time and Priority Done.

It might be the turn server I'm using does not provide a response at all if the credentials are incorrect. Does other turn servers generally provide a response if credentials are incorrect or some general auth error occurs?

Basically I just want to test the code with a real TURN server but cannot find a scenario where I see the errors? And no, I do not want to do a synthetic scenario, I want to try it out for real.

fippo:
Did you remove the stun server? See screenshot.
fail
Should be independent from the turn server, tested with ceod turn servers from apprtc (coturn) and talky (restund)

KaptenJansson:
Yeah I did remove it but could not get it to work. I might have left two servers in the list since I first added a legit server, then double clicked on it, modified it and added it back, in that case it does not fail.

So it seems that this only works with one server at a time?

fippo:
https://github.com/webrtc/samples/pull/488/files#diff-df48c0c3771b5ce959a279d283fcf9a7R162 -- by design -- if there is more than one server, there is ambiguity... which might be resolvable if you correlate ip addresses of the relay candidates to the relay candidate, but that doesn't work for hostnames. So a very specific test (which should probably be documented)

KaptenJansson:
Then we should update the instructions, the best would be if the UI would take care of this without having to instruct the user, I mean if you have a list, users will most likely assume you can populate it with several servers. 

Maybe only gather candidates for the selected server in the list? It is currently select-able. Need to make it clear though.

fippo:
Slightly updated instructions to explain the auth failure case at least.
This tool will always be for advanced users so let's see when the next chance to explain it on discuss-webrtc happens :-)

KaptenJansson:
LGTM

jncenterprises:
Remember guys It ain't broken don't fix it Just a heads upBless,Regards,President of board/CEOJnC.Enterprises+46764118516At Jun 4, 2015, 09:48:33, Christoffer Jansson wrote:Merged #488.

Issue #490, opened 2015-03-12
Favicon URL is wrong


KaptenJansson:
Seen on the following page but it most likely affects most sample pages with a favicon.
http://webrtc.github.io/samples/src/content/peerconnection/pc1/

http://webrtc.github.io/favicon.ico Failed to load resource: the server responded with a status of 404 (Not Found)

samdutton:
I can't see any files now using favicon, but I've done a pull request that updates the meta and link elements: #492.

KaptenJansson:
Fixed in  #492.

PR #492, opened 2015-03-12
Standardised all meta and link elements

samdutton:
See #490.

### DIFF
--- a/index.html
+++ b/index.html
@@ -11,7 +11,7 @@
 
   <meta charset="utf-8">
   <meta name="description" content="WebRTC Javascript code samples">
-  <meta name="viewport" content="width=device-width, user-scalable=no, initial-scale=1, maximum-scale=1">
+  <meta name="viewport" content="width=device-width, user-scalable=yes, initial-scale=1, maximum-scale=1">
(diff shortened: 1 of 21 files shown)


juberti:
I guess HTML imports aren't yet ready?

samdutton:
HTML Imports [aren't implemented yet in Firefox, IE or Safari](http://caniuse.com/#feat=imports) so I guess we should avoid this approach. 

Other approaches I've seen are quite complex:
* [Grunt templating](http://gruntjs.com/api/grunt.template).
* Other kinds of templating such as [Swig](https://www.npmjs.com/package/grunt-swig-templates).
* Simple [text replacement](https://github.com/yoniholmes/grunt-text-replace).

Templating means we'd by sticking to the DRY principle, but I'm not sure it's worth it. Once the meta elements have been standardised (in this PR) they shouldn't need to change much, and neither should other boilerplate content, since we don't add version numbers or other dynamic content. 

If we do want to go for templating, I think grunt.template should be able to do the job.

juberti:
This seems fine for now. If we ever decide to touch this again, the grunt template seems like a good idea.

Issue #494, opened 2015-03-17
Links on https://webrtc.github.io/samples/ redirects to HTTP


KaptenJansson:

samdutton:
More information about HTTPS and GitHub Pages [here](https://konklone.com/post/github-pages-now-sorta-supports-https-so-use-it).

KaptenJansson:
LGTM!

samdutton:
Changed, but links on [webrtc.github.io/samples](https://webrtc.github.io/samples/) still redirect to HTTPS URLs :/. 

Just saw comments [here](https://konklone.com/post/github-pages-now-sorta-supports-https-so-use-it) – looks like GitHub Pages doesn't support HTTPS for pages other than index.html.

You can enter HTTPS URLs manually, however.

KaptenJansson:
Closing as per last comment.

PR #495, opened 2015-03-17
Changed links to HTTPS

samdutton:
Attempt to force HTTPS: see #494. @KaptenJansson

### DIFF
--- a/index.html
+++ b/index.html
@@ -75,63 +75,63 @@
 
       <h3 id="getusermedia">getUserMedia</h3>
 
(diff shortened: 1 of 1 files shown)


samdutton:
Changed, but links on [webrtc.github.io/samples](https://webrtc.github.io/samples/) still redirect to HTTPS URLs :/.

juberti:
I noticed that this breaks index.html when run locally - it always points to github.io. Given that it doesn't seem to improve the situation, I suggest rolling it back.

juberti:
See #505

Issue #496, opened 2015-03-29
file:// URL warning needs to be updated


fippo:
from https://webrtc.github.io/samples/:
```
Please note that all samples that use getUserMedia() must be run from a server.
Calling getUserMedia() from a file:// URL will result in a PERMISSION_DENIED
NavigatorUserMediaError.
```
This doesn't seem to be true any longer. In fact, this just does nothing these days (@marwahvikas?). In any case, that text is probably obsolete, also given the ancient error name. A better way would be to warn inside adapter.

samdutton:
You're right – good spot. See PR #497.

KaptenJansson:
Fixed in #497

PR #497, opened 2015-03-30
#496: Updated text re file:// URL

samdutton:
See #496.

### DIFF
--- a/index.html
+++ b/index.html
@@ -59,7 +59,7 @@
 
       <p>Most of the samples use <a href="//github.com/webrtc/adapter">adapter.js</a>, a shim to insulate apps from spec changes and prefix differences. (In fact, the standards and protocols used for WebRTC implementations are highly stable, and there are only a few prefixed names. For full interop information, see <a href="//www.webrtc.org/web-apis/interop">webrtc.org/web-apis/interop</a>.)</p>
 
-      <p>Please note that all samples that use <code>getUserMedia()</code> must be run from a server. Calling <code>getUserMedia()</code> from a file:// URL won't work.</p>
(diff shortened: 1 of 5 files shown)


index.html
```diff
@@ -59,7 +59,7 @@
 
       <p>Most of the samples use <a href="//github.com/webrtc/adapter">adapter.js</a>, a shim to insulate apps from spec changes and prefix differences. (In fact, the standards and protocols used for WebRTC implementations are highly stable, and there are only a few prefixed names. For full interop information, see <a href="//www.webrtc.org/web-apis/interop">webrtc.org/web-apis/interop</a>.)</p>
 
-      <p>Please note that all samples that use <code>getUserMedia()</code> must be run from a server. Calling <code>getUserMedia()</code> from a file:// URL will result in a PERMISSION_DENIED NavigatorUserMediaError.</p>
+      <p>Please note that all samples that use <code>getUserMedia()</code> must be run from a server. Calling <code>getUserMedia()</code> from a file:// URL won't work.</p>
```
juberti: is an exception thrown, or does it fail silently?

samdutton: Fails silently in Chrome and Opera.

Works in Firefox – don't think it used to. I've updated the text in #496.

samdutton:
Added support for grunt-githooks, which forces grunt to run our validity checks before a git commit is executed (see my email to webrtc-eng).

@phoglund @juberti PTAL.

samdutton:
@phoglund @juberti ping. 

We should probably add grunt-githooks to the other repos as well.

KaptenJansson:
LGTM

PR #498, opened 2015-04-02
audio demo: add bitrate graph

fippo:
This adds a bitrate display to the audio-only demo. This is actually pretty interesting as it allows to visually  compare the audio bitrates of several codecs and the effects of silence.

The graph code is taken from webrtc-internals (removing the use of $ in two places and lots  of changes to make jscs and jshint happy) so that should not be a license problem. It doesn't really look as pretty as highcharts but it's ok.
Maybe it should go to src/js/lib so it can be reused by other demos?

@juberti @samdutton @jiayliu

### DIFF
--- a/src/content/peerconnection/audio/js/graph.js
+++ b/src/content/peerconnection/audio/js/graph.js
@@ -1,0 +1,667 @@
+// Copyright (c) 2013 The Chromium Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
+
+// taken from chrome://webrtc-internals
+
+// The maximum number of data points bufferred for each stats. Old data points
+// will be shifted out when the buffer is full.
+var MAX_STATS_DATA_POINT_BUFFER_SIZE = 1000;
+
+var TimelineDataSeries = (function() {
+  'use strict';
+
+  /**
+   * @constructor
+   */
+  function TimelineDataSeries() {
+    // List of DataPoints in chronological order.
+    this.dataPoints_ = [];
+
+    // Default color.  Should always be overridden prior to display.
+    this.color_ = 'red';
+    // Whether or not the data series should be drawn.
+    this.isVisible_ = true;
+
+    this.cacheStartTime_ = null;
+    this.cacheStepSize_ = 0;
+    this.cacheValues_ = [];
+  }
+
+  TimelineDataSeries.prototype = {
+    /**
+     * @override
+     */
+    toJSON: function() {
+      if (this.dataPoints_.length < 1)
+        return {};
+
+      var values = [];
+      for (var i = 0; i < this.dataPoints_.length; ++i) {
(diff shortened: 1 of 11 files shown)


samdutton:
Looks good – minor detail: could you sort out the code formatting problem that's causing the Travis failure? Thanks.

fippo:
done, should also fix the issue in #496

fippo:
disabling VAD suggested by @juberti -- this moves the PCMU bitrate back to where I would expect it.

Do we want this to be a "disable VAD" checkbox that is on by default? I'd really like to see a blog post explaining this.

src/content/peerconnection/audio/js/graph.js
```diff
@@ -0,0 +1,690 @@
+// Copyright (c) 2013 The Chromium Authors. All rights reserved.
```
juberti: Sam, how should we deal with this copyright, which is slightly different than the webrtc-samples copyright?

fippo:
Reduced canvas size to the same size used in webrtc-internals, this looked blurry.
Also added a graph for the number of packets which should be interesting when ptime can be tweaked.

KaptenJansson:
Ping samdutton@ regarding juberti@'s comment about copyright.

samdutton:
Oops – sorry. I looked into this, but forgot to follow it up. Just checking
now.

samdutton:
From the open-sourcing team at Google: 'please use the same WebRTC project authors heading throughout. If you need to, you can amend the AUTHORS files to also list "The Chromium Authors".'

In general they recommend putting all code like this in a third_party folder.

KaptenJansson:
@fippo ping regarding @samdutton's last comment.

fippo:
@samdutton how is that?

samdutton:
I followed up on this with open source team here (but forgot to update
here!)

They say:

No [don't use two licence notices], please use the same WebRTC project
> authors heading throughout. If you need to, you can amend the AUTHORS files
> to also list "The Chromium Authors".

fippo:
assuming that there will be an AUTHORS file after merge :-)

samdutton:
@fippo please sanity check the AUTHORS file I just added :).

fippo:
LGTM. I assume that me and @jan-ivar count as "webrtc project authors" since we signed the CLA (and heck, we have https://github.com/webrtc/samples/graphs/contributors anyway)

samdutton:
@juberti

Is 'WebRTC Project Authors' equivalent to repo contributors?

KaptenJansson:
I cannot find any other suitable "author" group unless we make up a new one like "WebRTC Project Contributors" which I think it's too close to "WebRTC Project Authors" to bother. But I'm no expert in that area.

KaptenJansson:
Ping, I would like to merge this.

samdutton:
I think we should stick with 'WebRTC Project Authors' – can change if
necessary.

PR #501, opened 2015-04-10
Add a variant of the file transfer demo that generates data instead.

phoglund:
The main purpose is to be able to pull the demo into a telemetry page and measure CPU and memory usage while transferring a large amount of data. I initially planned to just give the original filetransfer demo a generated file, but that turned out to be difficult on CrOS and Android. This variant just generates a random 16k string it sends over and over.

The demo is a reworked and somewhat stripped down variant of Philip Hancke's file transfer demo.

### DIFF
--- a/src/content/getusermedia/desktopcapture/extension/content-script.js
+++ b/src/content/getusermedia/desktopcapture/extension/content-script.js
@@ -17,7 +17,7 @@ port.onMessage.addListener(function(msg) {
 });
 
 window.addEventListener('message', function(event) {
-	// We only accept messages from ourselves
+  // We only accept messages from ourselves
   if (event.source !== window) {
     return;
   }
--- a/src/content/datachannel/datatransfer/css/main.css
+++ b/src/content/datachannel/datatransfer/css/main.css
@@ -1,0 +1,25 @@
+/*
+ *  Copyright (c) 2014 The WebRTC project authors. All Rights Reserved.
+ *
+ *  Use of this source code is governed by a BSD-style license
+ *  that can be found in the LICENSE file in the root of the source
+ *  tree.
+ */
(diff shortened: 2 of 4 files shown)


phoglund:
Sam, can you review this new demo?

src/content/getusermedia/desktopcapture/extension/content-script.js
```diff
@@ -17,7 +17,7 @@ port.onMessage.addListener(function(msg) {
 });
 
 window.addEventListener('message', function(event) {
-	// We only accept messages from ourselves
```
phoglund: I have no idea why Travis complained to me about this since I didn't change this file, but I though I might as well fix this while I was at it.

fippo: this was fixed in #498. I can cherry-pick from there if necessary.

samdutton: It's odd – also seems like Travis is complaining even though there's no complaint from JSCS locally. I'll look into it.

samdutton:
Nice one! LGTM.

src/content/getusermedia/desktopcapture/extension/content-script.js
fippo: check if your jscs version is different. I've seen that issue with an old version of jshint locally.

fippo:
awesome. nit: can you add a "megabytes" label after the megsToSend input?

samdutton:
Added to [samples home page](https://webrtc.github.io/samples/) and [tweeted](https://twitter.com/sw12/status/587618260952207360).

phoglund:
Wow, that was quick! Thanks.

PR #502, opened 2015-04-13
Cleaned up some dead code, reduced default to 128 megs, added Megs unit.

phoglund:
Addressing Phillip's comment on the last PR.

### DIFF
--- a/src/content/datachannel/datatransfer/index.html
+++ b/src/content/datachannel/datatransfer/index.html
@@ -46,7 +46,7 @@
 
     <section>
       <div>
-        <input type="number" id="megsToSend" min="1" name="megs" value="1048"/>
+        <input type="number" id="megsToSend" min="1" name="megs" value="128"/> Megs
         <button id="sendTheData" type="button">Generate and send data</button>
       </div>
       <div class="progress">
--- a/src/content/datachannel/datatransfer/index.html
+++ b/src/content/datachannel/datatransfer/index.html
@@ -45,9 +45,16 @@
     </section>
 
     <section>
-      <div>
-        <input type="number" id="megsToSend" min="1" name="megs" value="128"/> Megs
+      <div id="button">
         <button id="sendTheData" type="button">Generate and send data</button>
+      </div>
+      <div class="input">
+        <input type="number" id="megsToSend" min="1" name="megs" value="128"/>
+        <label for="megsToSend">MB</label>
+      </div>
+      <div class="input">
+        <input type="checkbox" id="ordered" checked>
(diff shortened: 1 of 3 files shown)


src/content/datachannel/datatransfer/index.html
```diff
@@ -46,7 +46,7 @@
 
     <section>
       <div>
-        <input type="number" id="megsToSend" min="1" name="megs" value="1048"/>
+        <input type="number" id="megsToSend" min="1" name="megs" value="128"/> Megs
```
samdutton: I think this should be MB, and should be in a label element. (Can't remember why the progress labels are in divs – had a good reason which I've now forgotten...)

Also (nitpick) I think the button should probably be left aligned above the input and progress element, to match demos like [this](http://webrtc.github.io/samples/src/content/datachannel/basic/). 

So:

    <div id="button">
        <button>...    
    </div>
    <div class="label">Data size: </div>
    <input ...>
    <label for="megsToSend">megabytes</label>

(The input/label content will be displayed on a single line.)

In css/main.css: 

    div#button {
        margin: 0 0 1em 0;
    }

(You can also get rid of the form selector in that file.)

src/content/datachannel/datatransfer/index.html
```diff
@@ -46,7 +46,7 @@
...
         <button id="sendTheData" type="button">Generate and send data</button>
```
juberti: Suggest adding a checkbox for ordered/unordered. We don't have any unordered demos, and we had a crasher in this area recently. The rest of the demo should work fine as-is in either case.

src/content/datachannel/datatransfer/index.html
phoglund: Done.

src/content/datachannel/datatransfer/index.html
phoglund: Ok, sure. How do I know it actually worked to go into unordered mode? I'm passing {ordered: true/false} into createDataChannel now based on the value of the checkbox.

phoglund:
PTAL

samdutton:
LGTM except that I get an error when the transfer completes:

    Uncaught InvalidStateError: Failed to execute 'send' on 'RTCDataChannel':
    RTCDataChannel.readyState is not 'open'main.js:83 generateData

There's also a minor code formatting glitch: there should be curly brackets for the if statement on main.js line 41.

phoglund:
Fixed nits and the invalid state error.

samdutton:
LGTM.

PR #508, opened 2015-04-18
change chunkSize to 32K

ashumeow:
The maximum chunksize is <code>128K</code>. But, it stopped in the middle and also it didn't work with <code>64K</code>. But, it works well with <code>32K</code> and faster than <code>16384</code>.


ashumeow:
I'll send better pull request in future. (^_^)

PR #509, opened 2015-04-18
changed to 32K

ashumeow:
With respect to https://github.com/webrtc/samples/pull/508, i made changes in <code>gh-pages</code>. Give it a try! =)

### DIFF
--- a/src/content/datachannel/datatransfer/js/main.js
+++ b/src/content/datachannel/datatransfer/js/main.js
@@ -72,7 +72,7 @@ function randomAsciiString(length) {
 function sendGeneratedData() {
   sendProgress.max = bytesToSend;
   receiveProgress.max = sendProgress.max;
-  var chunkSize = 16384;
+  var chunkSize = 32000;
   var stringToSendRepeatedly = randomAsciiString(chunkSize);
   var generateData = function(offset) {
     sendChannel.send(stringToSendRepeatedly);


juberti:
Suggest making this a separate input control, defaulting to 16k. (16k is the safe default for inter-browser use, iirc)

ashumeow:
I don't think so. For IIRC, if the chunkSize exceeds 256K [1], then it affects the performance. 32K chunkSize [2] isn't dangerous and it doesn't create RAID issues either. So, it will not cause any inter-browser problems too. I personally found that 16K is kinda less.

[1] https://code.google.com/p/chromium/issues/detail?id=294367#c21
[2] http://www.pikus.net/~pikus/plug_softraid/page20.html

juberti:
What does any of that have to do with webrtc data channels? The issue is with Firefox, which does (or used to) fragment chunks > 16KB.

See also http://www.html5rocks.com/en/tutorials/webrtc/datachannels/

ashumeow:
Sorry, i gave explanation in https://github.com/webrtc/samples/pull/509#issuecomment-94209366 to https://github.com/webrtc/samples/pull/509#issuecomment-94205614 that it won't create issues.

Coming to your https://github.com/webrtc/samples/pull/509#issuecomment-94241672 in relation to webrtc data channels, I've already read that article few months back... In that <a href="http://www.html5rocks.com/en/tutorials/webrtc/datachannels/">link</a>, there was a paragraph,
<code>Chunking is required, since there is currently a send size limit (though this will be fixed in a future version of data channels). The current recommendation for maximum chunk size is 16KB.</code>
This sentence clearly explains that "the size limit can also be further extended in future". Why not Google recommend it? =)

ashumeow:
I'll send better pull request in future. (^_^)

PR #493, opened 2015-03-14
Modern constraints samples using polyfill in adapter.js

jan-ivar:
I noticed the samples were promoting old constraint-syntax, so I updated the shim and two of the samples. Translation is logged to web console. Tested in Chrome 41 & 43 and Firefox 36-39. No-op on Firefox 38+.

PR #514, opened 2015-06-01
Update samples to use spec-compliant constraints.

jan-ivar:
A rebase of #493. @alvestrand do I need any automated tests here?

Note: I've also updated the /samples/js/adapter.js to the most recent one from /adapter (but I'm unsure if this is the correct procedure here).

To test manually, try [getusermedia/resolution/index.html](http://htmlpreview.github.io/?https://raw.githubusercontent.com/jan-ivar/samples/ecc0fcb98d24db1e393fbf5ad4acf94557794d3a/src/content/getusermedia/resolution/index.html) and [peerconnection/constraints.html](http://htmlpreview.github.io/?https://raw.githubusercontent.com/jan-ivar/samples/ecc0fcb98d24db1e393fbf5ad4acf94557794d3a/src/content/peerconnection/constraints/index.html).

### DIFF
--- a/src/content/getusermedia/resolution/index.html
+++ b/src/content/getusermedia/resolution/index.html
@@ -85,6 +85,7 @@
     <a href="https://github.com/webrtc/samples/tree/gh-pages/src/content/getusermedia/resolution" title="View source for this page on GitHub" id="viewSource">View source on GitHub</a>
   </div>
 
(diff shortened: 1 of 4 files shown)


KaptenJansson:
LGTM. FTR looking at removing adapter.js from the samples repo and fetching it via npm instead so that we do not need to update it in several places.

Issue #178, opened 2014-10-28
pr-answer sample is broken


smurugu:
The sample does not seem to display video on Party B (callee) on both Canary and GA versions

40.0.2176.0 canary (64-bit)
38.0.2125.104

http://googlechrome.github.io/webrtc/samples/web/content/peerconnection/pr-answer/

KaptenJansson:
webrtc bug: https://code.google.com/p/webrtc/issues/detail?id=3349.

SteveCuiPanPan:
Issue - I still see this error in the js console:Failed to set local  sdp: Session error code: ERROR_TRANSPORT. Session error description: Couldn't set up DTLS-SRTP on RTP channel..

Who can tell me how to solve this problem? Thanks a lot.

KaptenJansson:
This works in >= M48, please try it out.

samdutton:
BTW: This is now at https://webrtc.github.io/samples/src/content/peerconnection/pr-answer. 

Also — can't remember why this was never ported to the new sample format. Raised issue #704.

KaptenJansson:
Closing due to age.

mustadev:
example in  https://webrtc.github.io/samples/src/content/peerconnection/pr-answer. does not work.
does not show remote stream.

PR #519, opened 2015-06-08
Pranswer demo - Add error callbacks and move JS to separate file

KaptenJansson:
* Separated the JS to main.js
* Moved script src to end file to ensure DOM is loaded prior fetching buttons and video elements.
* Added error callbacks for setLocalDesription and createAnwser. Also added the stop() method to all the error callbacks which results in the user not having to reload the page.
* Updated copyright year.

This does not fix the demo per see (WebRTC is most likely to blame, not the demo page) however it surfaces the error we are currently seeing as mentioned in #178 and https://code.google.com/p/webrtc/issues/detail?id=3349.


samdutton:
LGTM.

PR #520, opened 2015-06-08
Added an update_dependencies.sh script.

alvestrand:
This will pull in the latest version of adapter.js and push it
into the js include directory.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -10,6 +10,8 @@ If appropriate, write a unit test which demonstrates that your code functions as
 
 To request a change or addition, you must [submit a pull request](https://help.github.com/categories/collaborating/).
 
+If you need to update the adapter.js version, run "npm install" and submit a CL with the resulting change - this will update the version in src/js/adapter.js.
+
 WebRTC developers monitor outstanding pull requests. They may request changes to the pull request before accepting. They will also verify that a CLA has been signed.
 
 The [Developer's Guide](https://bit.ly/webrtcdevguide) for this repo has more detailed information about code style, structure and validation.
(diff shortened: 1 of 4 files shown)


alvestrand:
@juberti @KaptenJansson this is a very pedestrian method of importing adapter.js. It allows us to easily pin to a version if desired, and encodes where we pull it from - but it does not break anything automatically on adapter.js updates, which is good.

Thoughts?

KaptenJansson:
Should we assume everyone is running in a shell environment? I think a python script is a more universal approach.

alvestrand:
I think we can safely assume that anyone who updates the dependencies has access to a shell environment. I don't want to call the NPM command line from python - I don't know how portable that command line is.

KaptenJansson:
Also, we would not need to have adapter.js in the master repo? Or do you want it there for consistency between master and gh-pages?

If the latter it would be nice if npm install would run the script automagically.

KaptenJansson:
npm scripts: https://docs.npmjs.com/misc/scripts

alvestrand:
The reasons for doing a copy and checking in the result:
- Allows us to serve out of gh-pages
- Allows anyone who checks out the demos to run them without running npm install
But there's no real reason for doing npm install without also copying into src/js/adapter.js.

New version, fully npm-scriptified. Anyone who does npm install will also be updating adapter.js.

KaptenJansson:
Perfect, I think we should also update README.md instructing users to run npm install.

KaptenJansson:
On second thought it's only necessary if you want the latest which potentially could be a different version.   Maybe we do not have to instruct the users about, rather just inform about it.

KaptenJansson:
LGTM with nit in previous comment.

alvestrand:
A note in CONTRIBUTING.js seems appropriate.

KaptenJansson:
LGTM

KaptenJansson:
@alvestrand Should we perhaps do this for apprtc as well?

PR #521, opened 2015-06-08
Pranswer demo - Add mandatory successCallbacks

KaptenJansson:
After updating to the latest adapter.js I noticed the a promise in adapter.js started to fail. Turned out I had nullified non nullable successCallbacks according to the spec. This fixes it.


samdutton:
LGTM.

Issue #472, opened 2015-02-05
Pull adapter.js as a dependency


juberti:
Instead of including our own copy.

KaptenJansson:
I'm looking into this, I spoke to @alvestrand and we came to the conclusion that we should use npm. WDYT? 

The downside is that we have to include the "node_modules" folder when we push to gh-pages branch. No manual step (other than npm install ofc) if we keep all plugins in the folder but if we want to clean it out prior pushing to gh-pages each npm plugin installed has to be removed.

alvestrand:
Either npm or bower should work; npm seems a bit more popular. @flippo says we can call the npm module "webrtc" if we want to.
I'm not sure whether using npm requires use of browserify when installing JS to the server. I'll experiment.
(package publishing also needs tests. I'll add some to adapter.js.)

KaptenJansson:
Assigning to @alvestrand for now.

piranna:
The node_modules is not necesary for gh-pages at all... why do you think it's needed?

piranna:
If you think about using them to publish de samples, I think is better to use bower, since it's focused for browsers Javascript.

KaptenJansson:
It is needed if we install adapter.js via npm.

alvestrand:
Check out PR #520 and see what you think. This makes the update manual (but scripted), which might be a Good Thing.

alvestrand:
Marking as resolved - after PR #520 doing "npm install" will auto-update the adapter.js script.

PR #524, opened 2015-06-10
createIceServer was removed

### LINKED ISSUE #523: 'Uncaught ReferenceError: CreateIceServer is not defined' is seen while trying to add the ICE Servers
1. Go to the page https://webrtc.github.io/samples/src/content/peerconnection/trickle-ice/
2. Enter the STUN or TURN URI detail
3. Click on Add Server
4. Expected Result:
5. The Server should be added

Actual Result: Instead, seeing the error message as follows:
'Uncaught ReferenceError: CreatIceServer is not defined'

filed on behalf of @selavazhagan, just moved the issue from https://github.com/webrtc/testrtc/issues/92 to here.

fippo:
fixes #523 @KaptenJansson (and makes me wonder why I did not notice that it broke my favorite demo)

### DIFF
--- a/src/content/peerconnection/trickle-ice/js/main.js
+++ b/src/content/peerconnection/trickle-ice/js/main.js
@@ -44,8 +44,11 @@ function addServer() {
 
   // Store the ICE server as a stringified JSON object in option.value.
   var option = document.createElement('option');
-  var iceServer = createIceServer(urlInput.value, usernameInput.value,
-      passwordInput.value);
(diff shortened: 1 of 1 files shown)


KaptenJansson:
Most likely since you tested with an older adapter.js from the samples repo? It was updated after you updated the demo (when we added it as a npm dependency in #472).

We should start adding tests for demo pages we care most about (preferably all). Will start looking into that.

LGTM

PR #525, opened 2015-06-11
candidate gathering demo: use urls everywhere

fippo:
@KaptenJansson looks like I missed a couple of issues in #524 :-/


KaptenJansson:
LGTM

PR #527, opened 2015-06-15
make canvas width/height equal video width/height

fippo:
mostly for cameras where the default resolution (and aspect ratio) is different from the default one


KaptenJansson:
LGTM!

PR #530, opened 2015-06-18
add bitrate like filetransfer demo

kawa-:
I have added bitrate in datatransfer demo, following filetransfer demo.

### DIFF
--- a/src/content/datachannel/datatransfer/index.html
+++ b/src/content/datachannel/datatransfer/index.html
@@ -64,7 +64,14 @@
         <div class="label">Receive progress: </div>
         <progress id="receiveProgress" max="0" value="0"></progress>
       </div>
-      <div id="bitrate"></div>
+      <div class="bitrate">
+        <div class="label">Current bitrate: </div>
+	<span id="currentBitrate"></span>
+      </div>
+      <div class="bitrate">
+	<div class="label">Average bitrate: </div>
+        <span id="bitrate"></span>
+      </div>
     </section>
 
     <section>
--- a/src/content/datachannel/datatransfer/index.html
+++ b/src/content/datachannel/datatransfer/index.html
@@ -60,11 +60,11 @@
         <div class="label">Send progress: </div>
         <progress id="sendProgress" max="0" value="0"></progress>
       </div>
-
       <div class="progress">
         <div class="label">Receive progress: </div>
         <progress id="receiveProgress" max="0" value="0"></progress>
       </div>
+      <div id="bitrate"></div>
     </section>
 
     <section>
--- a/src/content/datachannel/datatransfer/js/main.js
+++ b/src/content/datachannel/datatransfer/js/main.js
@@ -18,7 +18,8 @@ var sendButton = document.querySelector('button#sendTheData');
 var orderedCheckbox = document.querySelector('input#ordered');
 var sendProgress = document.querySelector('progress#sendProgress');
 var receiveProgress = document.querySelector('progress#receiveProgress');
-var bitrateDiv = document.querySelector('div#bitrate');
+var currentBitrate = document.querySelector('span#currentBitrate');
+var averagebitrate = document.querySelector('span#bitrate');
 
 var receivedSize = 0;
 var bytesToSend = 0;
@@ -89,10 +90,6 @@ function sendGeneratedData() {
   var chunkSize = 16384;
   var stringToSendRepeatedly = randomAsciiString(chunkSize);
   var sendAllData = function() {
-    // Try to queue up a bunch of data and back off when the channel starts to
-    // fill up. We don't setTimeout after each send since this lowers our
-    // throughput quite a bit (setTimeout(fn, 0) can take hundreds of milli-
-    // seconds to execute).
     while (sendProgress.value < sendProgress.max) {
       if (sendChannel.bufferedAmount > 5 * chunkSize) {
         setTimeout(sendAllData, 250);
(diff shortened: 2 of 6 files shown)


kawa-:
I signed it!

src/content/datachannel/datatransfer/js/main.js
```diff
@@ -168,6 +175,14 @@ function onReceiveMessageCallback(event) {
   receiveProgress.value = receivedSize;
 
   if (receivedSize === bytesToSend) {
+    var bitrate = Math.round(receivedSize * 8 /
+        ((new Date()).getTime() - timestampStart));
```
samdutton: You might want to use `window.performance.now()` for this (more information [here](http://www.html5rocks.com/en/tutorials/webperformance/usertiming/)).

src/content/datachannel/datatransfer/index.html
```diff
@@ -60,11 +60,11 @@
         <div class="label">Send progress: </div>
         <progress id="sendProgress" max="0" value="0"></progress>
       </div>
-
       <div class="progress">
         <div class="label">Receive progress: </div>
         <progress id="receiveProgress" max="0" value="0"></progress>
       </div>
+      <div id="bitrate"></div>
```
samdutton: To get consistent layout, probably best to parallel the progress divs, something like this:

    <div>
        <div class="label">Current bitrate: </div>
        <div id="bitrate"></div>
     </div>

samdutton:
Nice one!

Couple of nitpick comments inline.

Also noticed that if you change the transfer size and click **Generate and send data**, a negative bitrate is displayed initially.

kawa-:
@samdutton Thanks for your suggestions. I'm going to fix datatransfer and filetransfer sources.

kawa-:
- use User Timing API
- fix div for bitrate
- remove unnecessary comments

These are done in datatransfer. 

and also done in filetransfer:

- use User Timing API
- fix div for bitrate

>Also noticed that if you change the transfer size and click Generate and send data, a negative bitrate is displayed initially.

@samdutton I could'nt reproduce it. In new version (using User Timing API (thanks!)), it shows 0 at first. How is it in your machine now?

kawa-:
>continuous-integration/travis-ci/pr — The Travis CI build failed 

I'll fix them.

samdutton:
>
> I could'nt reproduce it. In new version (using User Timing API (thanks!)),
> it shows 0 at first. How is it in your machine now?


Thanks – can't reproduce now.

One other minor glitch: I get Nan kibts/sec when the send and receive
completes.
[image: Inline images 1]

Same part way through sending 2000MB:

[image: Inline images 4]

Also noticed this:

[image: Inline images 2]

And for 2000MB, the transfer never seems to complete:

[image: Inline images 3]

BTW: these demos completely freeze up the UI while the transfer is in
progress. @phoglund

kawa-:
>I get Nan kibts/sec when the send and receive completes

Yes, sometimes it happens. Need to debug.
And looks like the images are broken. Cloud you upload again?

samdutton:
> And looks like the images are broken. Cloud you upload again?

Oops – added images to comment, though only links to images seem to be working :/.

kawa-:
@samdutton in my examination about invalid current/max bitrate, it occurs from Chrome's setInterval's behavior. 

~~~
function onReceiveChannelStateChange() {
  var readyState = receiveChannel.readyState;
  trace('Receive channel state is: ' + readyState);
  if (readyState === 'open') {
    timestampStart = window.performance.now();
    timestampPrev = timestampStart;
    statsInterval = window.setInterval(displayStats, 500);
    window.setTimeout(displayStats, 100);
    window.setTimeout(displayStats, 300);
  }
}
~~~

In the code above, the line of `statsInterval = window.setInterval(displayStats, 500);` does not work in every 500 millisecond in chrome. Is there any tips to work properly?

Looks like it works properly in Firefox.

KaptenJansson:
>BTW: these demos completely freeze up the UI while the transfer is in progress.
I wonder if using web workers would fix this as it would not starve the main thread (only way to get multi thread support in JS AFAIK). I want to look into it but not sure I will have time for this for a while.

samdutton:
setInterval() and setTimeout() are not guaranteed to schedule accurately

You might be able to use window.requestAnimationFrame()
<https://developer.mozilla.org/en-US/docs/Web/API/window/requestAnimationFrame>,
then check the time, something like this:

var start = null;

function step(timestamp) {
  if (!start) start = timestamp;
  var progress = timestamp - start;
  //  if progress indicates 500ms or more has elapsed, display update
  window.requestAnimationFrame(step);
}

window.requestAnimationFrame(step);

kawa-:
@samdutton thanks, I tried requestAnimationFrame in this sample, but in chrome it does not work every 500 millisecond :( . It works at very first, but soon stops like the setInterval usage.

phoglund:
It's a surprisingly tricky problem to have both reliable UI updates and max throughput. I didn't look into it too much when I wrote the demo, since I just cared about throughput. I do think it's solvable though. Perhaps if you play around with how we're filling up the buffer and either fill it up more or less before yielding? The new bitrate counter will tell you how you're doing on throughput. You don't have to do that in this patch though, land this first (after addressing any remaining comments) and then do it in a second patch.

kawa-:
@phoglund 

>Perhaps if you play around with how we're filling up the buffer and either fill it up more or less before yielding? The new bitrate counter will tell you how you're doing on throughput. You don't have to do that in this patch

I wanted to measure the throughput of WebRTC DataChannel in various machines, browsers and network environments (wired, wireless, 3G/4G). Thanks to your code, I could do that (using User Timing API, also thanks to @samdutton). In my measurement code (not this PR), there's no UI when sending/receiving data (just showing plain text "now measuring..."). I thought it's good that there exists bitrate in data transfer demo and wrote this PR. But, 

>It's a surprisingly tricky problem to have both reliable UI updates and max throughput.

I can't resolve this problem, I close this PR. Thanks all, it's fun to have this constructive talk :)

PR #531, opened 2015-06-21
use navigator.mediaDevices.enumerateDevices

fippo:
it seems that the check for MediaStreamTrack is broken in Firefox 38 (which is a function there)... and we have this cool spec-compliant thing anyway :-)


KaptenJansson:
LGTM

PR #532, opened 2015-06-22
use getTracks to get all tracks

fippo:
from @alvestrand's comment in https://github.com/webrtc/adapter/issues/58#issuecomment-114026619 -- also uses attachMediaStream with a null stream which should already work (kind of) in the current adapter.js


KaptenJansson:
LGTM

PR #533, opened 2015-06-22
make filters work in firefox; add none filter

fippo:
Unfortunately it seems we can't remove the -webkit filters yet (@samdutton?) but adding standard ones makes this work in Firefox (cc @jan-ivar)

### DIFF
--- a/src/content/getusermedia/filter/index.html
+++ b/src/content/getusermedia/filter/index.html
@@ -27,17 +27,25 @@
   <link rel="stylesheet" href="../../../css/main.css">
 
   <style>
+    .none {
+      -webkit-filter: none;
+      filter: none;
+    }
     .blur {
       -webkit-filter: blur(3px);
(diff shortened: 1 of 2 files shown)


samdutton:
Good one – thanks Philipp.

LGTM.

> we can't remove the -webkit filters yet 

Correct – see [simpl.info/cssfilters](http://simpl.info/cssfilters/): doesn't work in Firefox as is, doesn't work in Chrome without prefix :^/. Discussion of Intent to Ship unprefixed [here](https://groups.google.com/a/chromium.org/forum/#!topic/blink-dev/T9evTlqPqTY).

jan-ivar:
@fippo - I know nothing about css unfortunately, but way to go anyway!

KaptenJansson:
This LGTM, anything blocking this from being merged @samdutton ?

samdutton:
LGTM. Merged.

PR #539, opened 2015-07-01
Add output device selection to source demo

KaptenJansson:
* Add output device selection
* Add "attachSinkId" to the local copy of adapter.js in the Samples repo, I do intend of landing the change in the adapter.js repo but for the sake of having a working pull request it's currently only here, will create a pull request in the adapter.js repo if this approach is approved.

To test this properly you need to host this page over HTTPS, this can be accomplished using Stunnel configured to point to python -m SimpleHTTPServer.

Use latest M45 Chrome and start it with the following command line flags:
--enable-blink-features=EnumerateDevices,AudioOutputDevices

Note: I've on purpose not changed the HTML layout, I would get some input on that if it should be changed or if it's acceptable just leaving it as is with 3 drop downs.

### DIFF
--- a/src/content/getusermedia/source/index.html
+++ b/src/content/getusermedia/source/index.html
@@ -43,10 +43,10 @@
 
     <h1><a href="//webrtc.github.io/samples/" title="WebRTC samples homepage">WebRTC samples</a> <span>Select audio &amp; video sources</span></h1>
 
-    <p>Get available audio, video sources and audio output devices* from <code>mediaDevices.enumerateDevices()</code> then set the source for <code>getUserMedia()</code> using a <code>sourceId</code> constraint.
+    <p>Get available audio, video sources and audio output devices* from <code>mediaDevices.enumerateDevices()</code> then set the source for <code>getUserMedia()</code> using a <code>sourceId</code> constraint.</p>
 
-    *Experimental support in Chrome 45.0.2441.x or later using command line flag "--enable-blink-features=EnumerateDevices,AudioOutputDevices". Use the <code>setSinkID()</code> method on the video element and provide a MediaStream and a sinkId (sourceId for where <code>sourceInfos.kind === 'audiooutput'</code>) as arguments.
-    </p>
+    <p>*Experimental support in Chrome 45.0.2441.x or later using command line flag "--enable-blink-features=EnumerateDevices,AudioOutputDevices". Use the <code>setSinkID()</code> method on the video element and provide a MediaStream and a sinkId (sourceId for where <code>sourceInfos.kind === 'audiooutput'</code>) as arguments.</p>
+
 
 
     <div class="select">
--- a/src/content/getusermedia/source/index.html
+++ b/src/content/getusermedia/source/index.html
@@ -1,6 +1,6 @@
 <!DOCTYPE html>
 <!--
- *  Copyright (c) 2014 The WebRTC project authors. All Rights Reserved.
+ *  Copyright (c) 2015 The WebRTC project authors. All Rights Reserved.
  *
  *  Use of this source code is governed by a BSD-style license
  *  that can be found in the LICENSE file in the root of the source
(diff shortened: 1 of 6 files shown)


KaptenJansson:
Since this can be a bit cumbersome to test I've pushed it to my gh-pages branch: http://kaptenjansson.github.io/samples/src/content/getusermedia/source/

KaptenJansson:
Also do note of an issue in Chrome https://code.google.com/p/chromium/issues/detail?id=506124. Maybe we should unmute the video element until that is fixed? ;)

samdutton:
Oh dear... I just built a sample [here](https://samdutton.github.io/webrtc/src/content/getusermedia/output-device/). This is only oriented to output (not input) selection. Looks like @KaptenJansson wasn't on the email thread about this :(...

Not sure if we should have one sample for input and one for output or if it's better to combine the two, as per this PR. 

Pro two: possibly less confusing, each demonstrates a single feature.
Pro one: both input and output selection is about using the devices got from `enumerateDevices()`, so it's actually informative to combine input and output selection in the same sample.

KaptenJansson:
I vote for "Pro one" as we are testing essentially testing enumeration and actual device selection, i.e. 
* Enumeration via enumerateDevices()
* Device selection via getUserMedia() and setSinkID() using sourceId.

Basically this is what this page has done all along, just adding another method for device selection, specifically for output.

I think we could still create a separate demo that focuses only on output device selection but includes the different element types and sources, i.e. 
* audio tag with a mp3 (just an example format/container) file as src 
* video tag with a mp4 (just an example format/container) file as src
* audio tag with an Audio MediaStream as src
* video tag with an Audio and Video MediaStream as src
* A device list with sourceId's and labels listed

samdutton:
>
> I think we could still create a separate demo that focuses only on output
> device selection but includes the different element types and sources
>

Good idea – I can/will modify mine to fit.

KaptenJansson:
@samdutton do you have any comments about this PR?

src/content/getusermedia/source/index.html
```diff
@@ -43,10 +43,18 @@
 
     <h1><a href="//webrtc.github.io/samples/" title="WebRTC samples homepage">WebRTC samples</a> <span>Select audio &amp; video sources</span></h1>
```
samdutton: Maybe change heading to `<span>Select sources &amp; outputs</span>`.

src/content/getusermedia/source/index.html
```diff
@@ -43,10 +43,18 @@
...
 
-    <p>Get available audio and video sources from <code>MediaStream.getSources()</code> then set the source for <code>getUserMedia()</code> using a <code>sourceId</code> constraint.</p>
+    <p>Get available audio, video sources and audio output devices* from <code>mediaDevices.enumerateDevices()</code> then set the source for <code>getUserMedia()</code> using a <code>sourceId</code> constraint.</p>
+
+    <p>*Experimental support in Chrome 45.0.2441.x or later using command line flag "--enable-blink-features=EnumerateDevices,AudioOutputDevices". Use the <code>setSinkID()</code> method on the video element and provide a MediaStream and a sinkId (sourceId for where <code>sourceInfos.kind === 'audiooutput'</code>) as arguments.</p>
```
samdutton: Maybe 'Enable experimental support in Chrome 45.0.2441.x or later by selecting <a href="chrome://flags/#enable-experimental-web-platform-features">Enable experimental Web Platform features</a> or by using command line flag ...'

samdutton:
LGTM apart from minor wording suggestions.

BTW: for PRs like this, putting a working demo on GitHub Pages is very helpful (thank you!) We should do this by default. @juberti: do you agree?

KaptenJansson:
>BTW: for PRs like this, putting a working demo on GitHub Pages is very helpful (thank you!) We should do this by default. @juberti: do you agree?

We could add that to the developers "handbook". 

On a general note I find that it would be really useful for all PR's but unless we have a "serve on the web"  button somewhere it's not plausible.

KaptenJansson:
I will make a separate pull request in the adapter.js repo with tests and land that first.

samdutton:
> unless we have a "serve on the web" button somewhere it's not plausible.

I was thinking that PRs should include a link to the associated page on the
contributor's GitHub Pages: the contributor pushes to their gh-pages branch
as well as master.

KaptenJansson:
SGTM

KaptenJansson:
Spoke to phoglund and we came to the conclusion the attachSinkId function is not needed in adapter.js. Will move it to the test page instead.

KaptenJansson:
I've pushed the latest to http://kaptenjansson.github.io/samples/src/content/getusermedia/source/

samdutton:
One extra nit – could you make the labels the same width, so the select
elements line up? Add this to the style element in the page:

label {
  width: 12em;
  display: inline-block;
}


Also (really pedantic!) add a full stop after 'must be served over HTTPS'.

KaptenJansson:
Latest pushed http://kaptenjansson.github.io/samples/src/content/getusermedia/source/

samdutton:
Magnificent :^).

LGTM.

samdutton:
Can you push to gh-pages?

KaptenJansson:
Done

samdutton:
Thanks.

Just noticed one other minor glitch: when a long name is associated with a
device, the select pushes the label down :(.

Probably best just to remove the CSS for select in index.html, i.e. remove
this:

select {
      width: 170px;

            }

samdutton:
We should probably put this sample at a different URL now that it's not
just about sources.

I think @juberti suggested a /devices directory — maybe this sample could
be at devices/input-output/ and the my one could be /devices/multi-input/
(or something)?

PR #541, opened 2015-07-08
Use spec deviceId constraint (makes selection work in Firefox).

jan-ivar:
Also made the buttons a bit wider to show the labels better (w/persistent permissions).


KaptenJansson:
LGTM

KaptenJansson:
Will push gh-pages shortly

PR #543, opened 2015-07-08
add redirect + fix broken links

### LINKED ISSUE #544: Seeing '404 file not found' for the requested file in Glimmer on the link mentioned
Steps:
1. Go to the link  https://webrtc.github.io/samples/src/content/getusermedia/source and hit 'Enter'

Expected Result:
There should be the page with the source selection option for the media.

Actual Result: 
Instead seeing the '404 File not Found' Error.  Checked in the samples page of the github page and from there also i am seeing the same error.   The screenshot is atached.
screenshot 2015-07-08 at 12 16 33 pm

KaptenJansson:
Addresses comments in https://github.com/webrtc/samples/pull/542#issuecomment-119630827

@jan-ivar WDYT?

Fixes #544

### DIFF
--- a/src/content/devices/input-output/index.html
+++ b/src/content/devices/input-output/index.html
@@ -49,7 +49,7 @@
 
     <p>Get available audio, video sources and audio output devices<b>*</b> from <code>mediaDevices.enumerateDevices()</code> then set the source for <code>getUserMedia()</code> using a <code>deviceId</code> constraint.</p>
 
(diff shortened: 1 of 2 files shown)


jan-ivar:
Not sure what `enhancement

KaptenJansson:
>Sorry, I didn't mean to alter the text. I meant: maybe not show this whole piece of text at all unless webrtcDetectedBrowser == 'chrome' in the code :)

I see ;). I've removed the text and enhancement label, was in a bit of a hurry yesterday and did not think it through.

Regarding the text I think we should always display the text otherwise users might wonder which browser output selection can be tested in. Also since the page now describes it can select outputs, having conditional text or capabilities that removes said capability will confuse users.

samdutton:
Thanks for sorting this! 

LGTM, though I'll look into doing redirects via Jekyll.

KaptenJansson:
I've pushed this to gh-pages branch and verified that the redirect works.

PR #548, opened 2015-07-13
use navigator.mediaDevices.getUserMedia

fippo:
instead of getUserMedia variable from adapter.js -- a result of https://github.com/webrtc/adapter/pull/75

### DIFF
--- a/src/content/peerconnection/audio/js/main.js
+++ b/src/content/peerconnection/audio/js/main.js
@@ -82,8 +82,7 @@ function call() {
   pc2.onicecandidate = iceCallback2;
   pc2.onaddstream = gotRemoteStream;
   trace('Requesting local stream');
-  // Call into getUserMedia via the polyfill (adapter.js).
-  getUserMedia({
+  navigator.getUserMedia({
       audio: true,
       video: false
     },
--- a/src/content/peerconnection/audio/js/main.js
+++ b/src/content/peerconnection/audio/js/main.js
@@ -82,13 +82,13 @@ function call() {
   pc2.onicecandidate = iceCallback2;
   pc2.onaddstream = gotRemoteStream;
   trace('Requesting local stream');
-  navigator.getUserMedia({
-      audio: true,
-      video: false
-    },
-    gotStream, function(e) {
-      alert('getUserMedia() error: ' + e.name);
-    });
+  navigator.mediaDevices.getUserMedia({
+    audio: true,
+    video: false
+  }).then(gotStream)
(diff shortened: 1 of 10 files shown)


fippo:
hrm. Actually after a quick chat with @jan-ivar I think we should go straight for navigator.mediaDevices.getUserMedia -- unless anyone objects (@samdutton?)

jan-ivar:
> I think we should go straight for navigator.mediaDevices.getUserMedia

+1

KaptenJansson:
>I think we should go straight for navigator.mediaDevices.getUserMedia

As per my comment in #549, please update this PR to use avigator.mediaDevices.getUserMedia instead.

Any objections @samdutton ?

samdutton:
No objections. Agree we should use this. Will also arrange for Web Fundamentals Update explaining the change.

fippo:
all done. Somewhat tempted to also inline the gotStream callbacks... thoughts?

jan-ivar:
Promises facilitate continuous forward code flow, so inlining `gotStream` fits my palate, but ymmv.

samdutton:
> Promises facilitate continuous forward code flow

Agree, but `gotStream()` is 10 lines or more in most of these, so I think it's clearer to keep it split out.

I think the split-out style is quite readable: `doSomething().then(handleResult).catch(handleError);`. 

Could be persuaded though — feel free to inline if you'd prefer :).

fippo:
Let's merge this and fight over in-lining later (over a beer) (-:

src/content/peerconnection/audio/js/main.js
```diff
@@ -82,14 +82,13 @@ function call() {
...
-      audio: true,
-      video: false
-    },
-    gotStream, function(e) {
-      alert('getUserMedia() error: ' + e.name);
-    });
+  navigator.mediaDevices.getUserMedia({
+    audio: true,
+    video: false
+  }).then(gotStream)
```
KaptenJansson: Cannot we have `.then` on a new line like in https://github.com/webrtc/samples/pull/548/files#diff-e28becd7475fb7e59b9088adaa6eebacR60 ?

samdutton:
LGTM with @KaptenJansson's suggestion.

src/content/peerconnection/audio/js/main.js
fippo: this is in several files. I have no opinion so will update

KaptenJansson:
LGTM

PR #549, opened 2015-07-13
add navigator.mediaDevices.getUserMedia demo

fippo:
since as @jan-ivar pointed out we currently don't have one. Simple copy of the gum sample.

Downside is that this doubles the maintenance effort...

### DIFF
--- a/index.html
+++ b/index.html
@@ -77,6 +77,8 @@
 
       <p><a href="src/content/getusermedia/gum">Basic getUserMedia demo</a></p>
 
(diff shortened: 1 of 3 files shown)


fippo:
might be unnecessary if all demos get updated to use the promise-based version (see discussion in #548)

KaptenJansson:
> might be unnecessary if all demos get updated to use the promise-based version (see discussion in #548)

Yes I would think so.

Issue #526, opened 2015-06-12
stats are not showing up on the constraints page in firefox 38


milkypostman:
think it's built to use the google stats collection. guess I will update when I complete webrtc/adapter#46

KaptenJansson:
Fixed in #550.

PR #550, opened 2015-07-14
support firefox in constraints demo

fippo:
when createOffer is called with only one argument, firefox interprets this a offerOptions and returns a promise so the callback is never called.

I'll also try to fix getStats while i'm in that file :-)

### DIFF
--- a/src/content/peerconnection/constraints/index.html
+++ b/src/content/peerconnection/constraints/index.html
@@ -85,6 +85,7 @@
         <video autoplay muted></video>
         <div></div>
         <div id="bitrate"></div>
+        <div id="peer"></div>
       </div>
     </section>
 
--- a/src/content/peerconnection/constraints/index.html
+++ b/src/content/peerconnection/constraints/index.html
@@ -42,8 +42,11 @@
       <p>The lefthand video shows the output of <code>getUserMedia()</code>; on the right is the video after being passed through the peer connection. The transmission bitrate is displayed below the righthand video.</p>
     </section>
 
-    <button id="getMedia">Get media</button>
-    <button id="connect" disabled>Connect</button>
+    <div>
+      <button id="getMedia">Get media</button>
+      <button id="connect" disabled>Connect</button>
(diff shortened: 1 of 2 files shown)


fippo:
uhm... since https://github.com/webrtc/samples/blob/master/src/content/peerconnection/constraints/js/main.js#L222
is working around code which (I think) is from Chrome 25 almost two years ago... any objections against completly rewriting this demo in a spec-compliant way?

fippo:
https://fippo.github.io/webrtc/src/content/peerconnection/constraints/
-- almost done. Just need to figure out how to get the peer ip and port in Firefox. 
For chrome I know how to do that (without making @juberti's eyes hurt even).

@milkypostman I think that's going to fix #526 :-)

KaptenJansson:
>https://fippo.github.io/webrtc/src/content/peerconnection/constraints/
-- almost done. Just need to figure out how to get the peer ip and port in Firefox. 
For chrome I know how to do that (without making @juberti's eyes hurt even).


Nice work! I noticed the constraints page allows the user to start arbitrary peerconnections by spamming the connect button ;). Created issue #552 for this. 

Do you want to add anything more to this pull request before merge?

fippo:
Ended up rewriting the demo. The only thing missing is figuring out the peer's ip address. I want a separate PR for this... actually tempted to demo it in https://github.com/webrtc/apprtc/pull/99 instead :-)

fippo:
meta question: should this be renamed/repurposed to be only about stats? we have the resolution demo for resolutions which afaics only lacks a framerate setting

jan-ivar:
I'm guessing the reason constraints cane be changed here is to test various bitrate loads, which presumably is still useful.

fippo:
and the peer ip is back. Looking for the remotecandidate after figuring out the active pair didn't require any backward compat hacks thankfully!

KaptenJansson:
LGTM, @samdutton WDYT?

samdutton:
Nice one. LGTM.

samdutton:
Pushed to gh-pages.

PR #551, opened 2015-07-14
support for firefox getStats

fippo:
this makes the graph in the pc audio demo work in Firefox, too, thanks to the shim of getStats in adapter.js

note: it took me a bit to figure out that adapter doesn't shim the type=outboundrtp on the stats report. There are probably many pitfalls here...

### DIFF
--- a/src/content/peerconnection/audio/js/main.js
+++ b/src/content/peerconnection/audio/js/main.js
@@ -244,38 +244,39 @@ function setDefaultCodec(mLine, payload) {
 }
 
 // query getStats every second
-if (webrtcDetectedBrowser === 'chrome') {
-  window.setInterval(function() {
-    if (!window.pc1) {
-      return;
-    }
-    window.pc1.getStats(function(res) {
-      res.result().forEach(function(report) {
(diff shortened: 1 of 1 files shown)


src/content/peerconnection/audio/js/main.js
```diff
@@ -244,38 +244,40 @@ function setDefaultCodec(mLine, payload) {
...
-      });
+        lastBytes = bytes;
+        lastPackets = packets;
+        lastTime = now;
+      }
     });
-  }, 1000);
-}
+  }, function(err) {
+    console.log(err);
```
KaptenJansson: Not sure how severe it is if the error callback is triggered but I'm leaning more to use console.error() or console.warn() for errors to separate info and errors, sure this error could be considered as info but still. WDYT?

fippo: I think this is not even hooked up in any browser yet... but +1

KaptenJansson:
LGTM

KaptenJansson:
Will do a gh-pages push later today.

PR #558, opened 2015-07-14
Use container based build system

KaptenJansson:
Reasoning here http://docs.travis-ci.com/user/migrating-from-legacy/?utm_source=legacy-notice&utm_medium=banner&utm_campaign=legacy-upgrade

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,4 @@
+sudo: false
 language: node_js
 node_js:
   - 0.10


samdutton:
Hadn't heard of this, but SGTM.

KaptenJansson:
I noticed that the apprtc repo was using it and Travis reported it was using the "deprecated version" of the build setup.

Issue #564, opened 2015-07-16
PermissionDeniedError is seen for the peerconnection/constraints page


selavazhagan:
1. Use the link https://webrtc.github.io/samples/src/content/peerconnection/constraints on the device Paine(45.0.2454.4/7262.3.0 dev) 
2. Click on 'Get Media'

Expected Result: 
The media should be seen in the window below

Actual Result:
The error message is seen as s
![screenshot 2015-07-16 at 2 29 57 pm](https://cloud.githubusercontent.com/assets/9001606/8735697/addfb7cc-2bc7-11e5-896a-d4f4f42634cf.png)
een in the screenshot attached

fippo:
ah... 45 adds navigator.mediaDevices.getUserMedia so it's no longer shimmed in https://github.com/webrtc/adapter
However, the non-shimmed version doesn't accept the spec-constraints. @KaptenJansson will make a PR in adapter

KaptenJansson:
To give background to why this slipped by my testing, I was testing using 45.0.2441.3 while this was introduced in 45.0.2453.0 it seems.

Note to self, always test on latest and greatest.

PR #567, opened 2015-07-17
add basic selenium test

fippo:
this adds a basic selenium test to the gum sample. What the test does is that the video has a videoWidth and videoHeight set which imo is the right high-level assertion that things work.

Ideally, we can get this thing to run on travis for every PR as well as nightly on every new canary build so you notice things like #564 without someone reporting it, but there is some work left. 
Currently requires manual installation of [selenium-webdriver](https://www.npmjs.com/package/selenium-webdriver) from npm, also $BROWSER must be set (and requires chromedriver for running in chrome)

### DIFF
--- a/src/content/getusermedia/gum/js/test.js
+++ b/src/content/getusermedia/gum/js/test.js
@@ -63,5 +63,10 @@ test('video width and video height are set on GUM sample', function(t) {
     t.pass('got video dimensions ' + dimensions.join('x'));
     driver.quit();
     t.end();
+  })
+  .then(null, function (err) {
+    t.fail(err);
+    driver.quit();
+    t.end();
   });
 });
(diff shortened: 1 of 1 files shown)


KaptenJansson:
To clarify the steps required to run a test (this worked for me):

1. Chrome installed already on your system.
2. (sudo) npm install -g selenium-webdriver
3. (sudo) npm install -g chromedriver (other browsers require standalone selenium)
4. BROWSER=chrome node src/content/getusermedia/gum/js/test.js

Results:
TAP version 13
# video width and video height are set on GUM sample
ok 1 page loaded
ok 2 found video element
ok 3 got videoWidth 640
ok 4 got videoHeight 480
ok 5 got video dimensions 640x480

1..5
# tests 5
# pass  5

# ok

PR #571, opened 2015-07-21
Fippo selenium all the way

KaptenJansson:
This PR is based on on top of @fippo's work from branch https://github.com/fippo/webrtc/tree/selenium-all-the-way

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,6 +11,7 @@ env:
     - BROWSER=firefox BVER=stable
     - BROWSER=firefox BVER=beta
     - BROWSER=firefox BVER=unstable
+    - BROWSER=firefox BVER=esr
 
 matrix:
   fast_finish: true
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,9 +3,29 @@ language: node_js
 node_js:
   - 0.10
 
-install:
+env:
+  matrix:
+    - BROWSER=chrome  BVER=stable
+    - BROWSER=chrome  BVER=beta
+    - BROWSER=chrome  BVER=unstable
+    - BROWSER=firefox BVER=stable
+    - BROWSER=firefox BVER=beta
+    - BROWSER=firefox BVER=unstable
+
+matrix:
+  fast_finish: true
+
+  allow_failures:
+    - env: BROWSER=chrome  BVER=unstable
+    - env: BROWSER=firefox BVER=unstable
+
+before_script:
   - npm install
+  - export DISPLAY=:99.0
+  - sh -e /etc/init.d/xvfb start
 
 script:
-  - grunt
   - npm test
+
+after_failure:
+  - for file in *.log; do echo $file; echo "======================"; cat $file; done
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,3 +8,4 @@ install:
 
 script:
   - grunt
+  - npm test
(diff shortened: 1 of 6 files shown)


.travis.yml
```diff
@@ -3,8 +3,29 @@ language: node_js
...
 
-install:
+env:
+  matrix:
+    - BROWSER=chrome  BVER=stable
+    - BROWSER=chrome  BVER=beta
+    - BROWSER=chrome  BVER=unstable
+    - BROWSER=firefox BVER=stable
+    - BROWSER=firefox BVER=beta
+    - BROWSER=firefox BVER=unstable
```
fippo: add BVER=esr

fippo:
i can add a piece of code that polls for videoWidth every 500ms in a bit.

KaptenJansson:
Firefox ESR does not honor media.navigator.streams.fake = true settings hence why it's failing.

jan-ivar:
@KaptenJansson - Looks like `media.navigator.streams.fake` [was added in 39](https://bugzilla.mozilla.org/show_bug.cgi?id=1110973). Hmm, the workaround would be to modify the constraints. e.g. `{video:true, fake:true}`. Is that doable somehow?

fippo:
@jan-ivar  that explains why it works in adapter -- we're explicitly setting fake:true there. Let's skip ESR for now? Injecting something into localStorage which makes the GUM calls use fake:true might be difficult for any tests that call GUM immediately

jan-ivar:
You mean skip ESR in the samples test? Sure. We'd still test ESR in adapter?

KaptenJansson:
> You mean skip ESR in the samples test? Sure. We'd still test ESR in adapter?
Yes and Yes. We can add it back once it catches up to 39.

jan-ivar:
> We can add it back once it catches up to 39.

Note that the [next ESR](https://www.mozilla.org/en-US/firefox/organizations/faq) will be Firefox 45.

jan-ivar:
Do these tests by any chance run only on linux? If so there may be another way.

KaptenJansson:
Currently only Linux yes but we could if enough people wanted to add the possibility to run on any Windows and Mac as well using the currently installed browsers locally.

fippo:
@tkchin maybe you can help @googlebot with finding the right CLA for my email? The one from https://codereview.webrtc.org/1173403008 should cover that (and github can associate the commits with my github username so...)

tkchin:
AUTHORS doesn't have anything to do with finding CLA. You should check that the git commit has the appropriate email address. I haven't checked, but my guess is that somewhere in the chain of CLs there is an unapproved email address. If you still run into problems @KaptenJansson should be able to  manually override it.

KaptenJansson:
@fippo I just checked and fippo@andynet.net has not been used to sign the CLA while your gmail account has.

KaptenJansson:
But your last PR used fippo@andynet.net and that was approved. This is strange. I think this because I made a PR with your commits, i.e. I'm not the initial author and googlebot@ do not approve this automatically.

KaptenJansson:
Overriding label

tkchin:
Should it be andyet.net?

fippo:
I think googlebot uses my github username usually. The corporate CLA is valid for my andyet.net email but that is not associated with my github username which might be the reason the check fails. But a CLA covering this exists so... (-:

samdutton:
As Zeke suggested, is this a problem with fippo@andy*n*et.net versus
fippo@andyet.net <fippo@andynet.net>?

KaptenJansson:
Also remember you need to sign the CLA per email address you use for the googlebot@ to be happy, at least that's what I did and how I interpret the documentation.

PR #572, opened 2015-07-21
wait for one second to make sure attachMediaStream worked

fippo:
also move common selenium stuff to a module

@KaptenJansson

### DIFF
--- a/src/content/getusermedia/gum/js/test.js
+++ b/src/content/getusermedia/gum/js/test.js
@@ -12,43 +12,21 @@
 /* jshint node: true */
 var test = require('tape');
 
-// https://code.google.com/p/selenium/wiki/WebDriverJs
 var webdriver = require('selenium-webdriver');
-var chrome = require('selenium-webdriver/chrome');
-var firefox = require('selenium-webdriver/firefox');
-
-var profile = null;
-var firefoxOptions = null;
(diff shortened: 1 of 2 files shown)


src/content/getusermedia/gum/js/test.js
```diff
@@ -12,44 +12,22 @@
...
+  var driver = seleniumHelpers.buildDriver();
 
   driver.get('file://' + process.cwd() +
       '/src/content/getusermedia/gum/index.html')
   .then(function() {
     t.pass('Page loaded');
   })
+  .then(function() {
+    // sleep for one second
+    return driver.sleep(1000);
```
KaptenJansson: Too bad we to use sleep here but we do not have a choice for now.

jan-ivar: Why exactly is this needed?

fippo: but we can't hook onloadedmetadata from selenium unfortunately

jan-ivar: In that case is 1 second enough? With debug builds + camera hardware state, I've used 4 seconds in some of my tests to be sure, but that was with real hardware, rather than fake, so maybe that's ok.

PR #574, opened 2015-07-22
add munge-sdp and trickle-ice selenium tests

fippo:
tests for my favorite demos. Comments inline...

PR #576, opened 2015-07-22
add selenium tests for munge-sdp and trickle ice pages

fippo:
basically #574 without all the commits trying to figure out what went wrong.

* fixes the shared profile issue created by moving stuff to the lib
* adds tests for munge-sdp and trickle-ice
* makes the gum demo not wait for one second but until window.stream is defined (i think this polls every 500ms)

### DIFF
--- a/src/content/getusermedia/gum/js/test.js
+++ b/src/content/getusermedia/gum/js/test.js
@@ -25,8 +25,9 @@ test('Video width and video height are set on GUM sample', function(t) {
     t.pass('Page loaded');
   })
   .then(function() {
-    // sleep for one second
-    return driver.sleep(1000);
+    return driver.wait(function() {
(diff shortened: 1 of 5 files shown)


fippo:
the test seems flaky. Also spams alot of Firefox windows locally which I think is related overriding the profile created with "integrationtest" in https://github.com/DamonOehlman/travis-multirunner/blob/master/start-firefox.sh#L25

fippo:
the "fix" in the last commit is odd. Why is this only a problem with the last test run?!

KaptenJansson:
Also why is firefox-unstable hanging on the first test? I traced it to the `driver.get('file://' + process.cwd() +
      '/src/content/getusermedia/gum/index.html')` step in the gum test. The .then Promise method is never triggered it seems.

KaptenJansson:
LGTM

PR #579, opened 2015-07-22
update extension name. check in the zip file.

guoweis:


guoweis:
No need to approve this one as more change is probably coming.

PR #585, opened 2015-07-23
add basic datachannel and pc1 tests

fippo:
I wish there was a "slowdown" mode for selenium where it delays actions for 1 second so one can watch this... :-)


KaptenJansson:
LGTM

PR #589, opened 2015-07-24
Adding the domain and trailing slash solves HTTPS issue

### LINKED ISSUE #586: How do I link to HTTPS samples?
Maybe I'm missing something fundamental, but I'm having trouble linking people to the HTTPS version of https://webrtc.github.io/samples/src/content/devices/input-output, e.g. from [here](http://stackoverflow.com/questions/31586889/how-to-detect-microphone-type/31601758#31601758), which is sad, because the HTTP version doesn't show labels, which is hard to explain.

Is there a way to fix or work around this from our end?

KaptenJansson:
So just adding a trailing slash does not solve it however adding the URL as well does.

Fixes #586

### DIFF
--- a/index.html
+++ b/index.html
@@ -75,63 +75,63 @@
 
       <h3 id="getusermedia">getUserMedia</h3>
 
-      <p><a href="src/content/getusermedia/gum/">Basic getUserMedia demo</a></p>
+      <p><a href="//webrtc.github.io/samples/src/content/getusermedia/gum/">Basic getUserMedia demo</a></p>
 
-      <p><a href="src/content/getusermedia/canvas/">Use getUserMedia with canvas</a></p>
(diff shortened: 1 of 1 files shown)


KaptenJansson:
Actually I'm going to merge this and TBR samdutton as I'm curios to see if it works on github. FTR I did modify the existing page served on github in Chrome and got it to work so it should work.

samdutton:
AFAICT the trailing slash works for links with relative URLs on a page on GitHub if (and only if) the page has a HTTPS URL. 

Using a protocol relative URL (i.e. with `//`) in the link means it will use either HTTP or HTTPS, depending on the scheme of the page the link is on.

Hope that makes sense :).

KaptenJansson:
Good to know, but since we link to https://webrtc.github.io/samples/  on the main repo [page](https://github.com/webrtc/samples) we cover most cases with this approach. I can live with that caveat.

PR #593, opened 2015-07-26
Input output test

KaptenJansson:
This adds a simple device selection test. This has a potential of being expanded if we add more fake devices. Details below.

Also FTR I created a firefox [bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1187738) asking if we can add fake device labels.

WDYT about filing feature request for adding a 2nd fake video device so we can swap between devices (chrome already has 2 fake audio devices)? 

I will also file a bug about the fake video device name inconsistency in Chrome compared to the fake audio devices (`fake_device_0` vs `Fake Audio 1`), could just mentioned that in the potential feature request mentioned in previous paragraph. 

@samdutton @fippo @jan-ivar PTAL.


src/content/devices/input-output/js/test.js
```diff
@@ -0,0 +1,135 @@
...
+  })
+  .then(function() {
+    // Making sure we can select the 1st video device.
+    // TODO: Select more devices if Firefox adds a 2nd fake A/V device and
+    // Chrome adds another fake video device.
+    t.pass('Selecting 1st video device');
+    var option = driver.findElement(webdriver.By.xpath(
+        'id("videoSource")/option[1]'));
+    option.click();
+    return option.getAttribute('value');
```
fippo: how about chaining this? i.e. 
```return driver.findElement(...)
.click()
.getAttribute('value')
```
Also above. I have a slight preference for that style since it doesn't appear to be synchronous while it actually isn't. See https://code.google.com/p/selenium/wiki/WebDriverJs#Value_Propagation_and_Chaining

fippo:
i'm getting a
```
ok 33 Fake audio device found with label: Fake Audio 1
not ok 34 Fake video device found with label: fake_device_0 (0bda:5752)
```
when trying to run this in Chrome 44 on Linux. Travis works however.

I'll see if I can fix the Firefox stuff. It seems to be leaking windows/profiles which causes the test to fail.

src/content/devices/input-output/js/test.js
```diff
@@ -0,0 +1,131 @@
...
+        }
+      });
+    });
+  })
+  .then(function(dimensions) {
+    t.pass('Got video dimensions ' + dimensions.join('x'));
+  })
+  .then(function() {
+    t.pass('shutting down success');
+    // Chrome does not shutdown unless close() is called due to starting it via
```
fippo: obsolete with #594 -- thankfully

KaptenJansson:
> ok 33 Fake audio device found with label: Fake Audio 1
not ok 34 Fake video device found with label: fake_device_0 (0bda:5752)

Looks like a USB id has been added to the fake device name, this could be potentially really flaky :(

src/content/devices/input-output/js/test.js
KaptenJansson: The problem is that the element object does not return by xpath() hence no methods can be used (undefined error for click()), however after assigning it to a variable, it works fine.

I tried a few variants to see if I could get to work as you suggested but I failed.

src/content/devices/input-output/js/test.js
KaptenJansson: Removed it.

src/content/devices/input-output/js/test.js
fippo: how about
.By.css('#videoSource>option')
(see the trickle test)

KaptenJansson: Done

src/content/devices/input-output/js/test.js
```diff
@@ -0,0 +1,133 @@
...
+    return new webdriver.ActionSequence(driver).
+        doubleClick(element).perform();
+  })
+  // Check enumerateDevices has returned an id.
+  .then(function() {
+    return driver.findElement(webdriver.By.css(
+        '#videoSource>option')).getAttribute('value');
+  })
+  .then(function(deviceId) {
+    t.ok(deviceId !== '', 'Device/source id: ' + deviceId);
```
jan-ivar: This test will succeed if deviceId is `undefined`. How about `t.ok(deviceId, 'Device/source id: ' + deviceId);` ?

src/content/devices/input-output/js/test.js
```diff
@@ -0,0 +1,133 @@
...
+  })
+  .then(function() {
+    // Make sure the stream is ready.
+    return driver.wait(function() {
+      return driver.executeScript('return window.stream !== undefined;');
+    }, 30 * 1000);
+  })
+   // Check for a fake audio device label (Chrome only).
```
jan-ivar: Are we testing the sample or the browser here? If we're testing the logic in the sample, would using use `deviceId` instead help?

KaptenJansson: The problem is that the deviceId cannot be tied to the stream, sure you can test that after selecting a device that you get an active stream with appropriate tracks.

I want to verify that the device we select is the one used for the stream, having said that I see a flaw in my test, I do not select device based on the label, rather deviceId, this is due to it not being populated in the drop down, this could ofc be fixed by making sure real labels are populated in the drop down prior selecting them (I've seen fake labels being populated in Chrome while in selenium even though it's not served over HTTPS).

src/content/devices/input-output/js/test.js
KaptenJansson: That is true, I assumed adapter.js would at least deliver an empty string but that's a bad assumption as that is something we want to catch ofc.

Will do as you suggest.

src/content/devices/input-output/js/test.js
jan-ivar: `deviceId` is also guaranteed to be unique, whereas `label` is not afaik.

KaptenJansson: Still not sure what to do here, the id on a `mediaStreamTrack` is a unique string not related to `deviceId`. So at best I can only make sure we get a `deviceId` and a `mediaStream` with live media. If we had a promise to get labels from `enumerateDevices` we could match those to the labels on `mediaStreamTrack`.

Not sure what `deviceId` uniqueness provides?

jan-ivar: I was confused. Hopefully with https://github.com/webrtc/samples/pull/590 we'll have labels to compare.

jan-ivar: In firefox, if we set the pref `media.navigator.permission.disabled` to `true`, that should make labels appear in the test fwiw.

KaptenJansson: When I try that I get labels from enumerateDevices but not on the streams still.

labels

But we can update that later if that changes in FF.

jan-ivar: I've verified that mst.label is not yet implemented in firefox.

jan-ivar: I've filed https://bugzil.la/1188755

KaptenJansson: Good to know, I guess we can keep track of https://bugzilla.mozilla.org/show_bug.cgi?id=1187738 to know when it's implemented.

KaptenJansson:
> ok 33 Fake audio device found with label: Fake Audio 1
not ok 34 Fake video device found with label: fake_device_0 (0bda:5752)
Looks like a USB id has been added to the fake device name, this could be potentially really flaky :(

I'm not going to add substring comparison here because I do not think we should hide this issue (USB id in the label for fake devices in Chrome). It runs fine on Travis by the looks of it. We can ofc change this later if it becomes a real hassle when running locally.

src/content/devices/input-output/js/test.js
KaptenJansson: Aha, then you can dupe my bug to that one https://bugzilla.mozilla.org/show_bug.cgi?id=1187738.

jan-ivar: I duped mine to yours.

PR #594, opened 2015-07-27
use a single webdriver instance

fippo:
Since we don't use localStorage using a single driver instance seems ok to me. Also, tests run quite a bit faster.

@KaptenJansson might result in a merge conflict with #593. But should solve the Firefox failure.


KaptenJansson:
LGTM

Issue #555, opened 2015-07-14
video.currentSrc in pc1 demo


fippo:
https://github.com/webrtc/samples/blob/master/src/content/peerconnection/pc1/js/main.js#L25

i'm not sure what .currentSrc on a video is even but since this doesn't work in Firefox I assume it's non-standard. I'd just remove that and only log the video sizes.

 @samdutton?

KaptenJansson:
What about this:

    localVideo.addEventListener('loadedmetadata', function(event) {
      var src = this.src || this.mozSrcObject.id;
      trace('Local video src: ' + src + ', videoWidth: ' + this.videoWidth +
          'px,  videoHeight: ' + this.videoHeight + 'px');
    });

    remoteVideo.addEventListener('loadedmetadata', function() {
      var src = this.src || this.mozSrcObject.id;
      trace('Remote video src: ' + src +', videoWidth: ' + this.videoWidth +
          'px,  videoHeight: ' + this.videoHeight + 'px');
    });

samdutton:
`currentSrc` gives the current source of a video or audio element. Used here just to show the (blob) src of the video, but makes most sense when JavaScript is used to set the src or with the sources element: [simpl.info/v](https://simpl.info/v). More info [here](https://developer.mozilla.org/en-US/docs/Web/API/HTMLMediaElement).

See what you mean though, that it doesn't work here for Firefox :/ (though it does for [simpl.info/v](https://simpl.info/v).

KaptenJansson:
Firefox supports video.src for media files, not mediaStreams.

jan-ivar:
What should `video.currentSrc` read when source was set using `video.srcObject`?

samdutton:
Not sure where/if this is specified, but Chrome shows the blob URL.

jan-ivar:
> Not sure where/if this is specified, but Chrome shows the blob URL.

@samdutton but looking at https://github.com/webrtc/adapter/blob/master/adapter.js#L326 :

    if (typeof element.srcObject !== 'undefined') {
      element.srcObject = stream;
    } else if (typeof element.src !== 'undefined') {
      element.src = URL.createObjectURL(stream);

there is no blob url in the `srcObject` case, is there?

jan-ivar:
From https://html.spec.whatwg.org/multipage/embedded-content.html#concept-media-load-algorithm :
> If mode is object
>        ⌛ Set the currentSrc attribute to the empty string.

samdutton:
[...]

PR #596, opened 2015-07-28
remove currentSrc per #555

fippo:
remove currentSrc per discussion in #555


samdutton:
LGTM — but could you fix the JSCS nits while you're in this file :)?

KaptenJansson:
Merging this as the nits were fixed in #599

PR #599, opened 2015-07-28
Fix grunt-jscs 2.0 errors

KaptenJansson:
Same story as in https://github.com/webrtc/adapter/pull/101.

I do not like adding inliners like that on each one but we do not have have choice for our prefixed constructors, it's better than turning off the requireCapitalizedConstructors check globally.


samdutton:
LGTM.

PR #600, opened 2015-07-28
Add zip file of desktopCaptureExtension and grunt compress

KaptenJansson:
* Create zip file of desktopcapturextension for easier download.
* Add grunt-contrib-compress and a `grunt compress` task to compress desktopCaptureExtension
* Add instructions for the compress step in the extension README.md file
* Cleaned up the order of the package.json file (actually installing plugins using the --save-dev flag does this for you;))


samdutton:
LGTM, just waiting for Travis.

PR #602, opened 2015-07-28
add filetransfer test

fippo:
currently transfers index.html. Should this maybe also transfer some larger files from
https://github.com/webrtc/samples/tree/master/src/content/devices/multi/video ?

### DIFF
--- a/src/content/datachannel/filetransfer/js/test.js
+++ b/src/content/datachannel/filetransfer/js/test.js
@@ -20,9 +20,8 @@ function sendFile(t, path) {
   })
   .then(function() {
     // Wait for the received element to be displayed.
-    return driver.wait(function() {
-      return driver.findElement(webdriver.By.id('received')).isDisplayed();
-    }, 30 * 1000);
+    return driver.wait(webdriver.until.elementLocated(
+        webdriver.By.id('received')));
   })
   .then(function() {
     t.end();
--- a/src/content/datachannel/filetransfer/js/test.js
+++ b/src/content/datachannel/filetransfer/js/test.js
@@ -7,7 +7,7 @@ var test = require('tape');
 var webdriver = require('selenium-webdriver');
 var seleniumHelpers = require('../../../../../test/selenium-lib');
 
-test('Filetransfer via Datachannels', function(t) {
+function sendFile(t, path) {
   var driver = seleniumHelpers.buildDriver();
 
   driver.get('file://' + process.cwd() +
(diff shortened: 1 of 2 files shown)


src/content/datachannel/filetransfer/js/test.js
```diff
@@ -0,0 +1,34 @@
...
+      '/src/content/datachannel/filetransfer/index.html')
+  .then(function() {
+    t.pass('page loaded');
+    // Based on https://saucelabs.com/resources/articles/selenium-file-upload
+    return driver.findElement(webdriver.By.id('fileInput'))
+       .sendKeys(process.cwd() + '/index.html');
+  })
+  .then(function() {
+    // Wait for the received element to be displayed.
+    return driver.wait(function() {
```
KaptenJansson: Checkout the [variant](https://github.com/webrtc/samples/commit/3ec2294cadc7e5323f4bb83abc956c3da1ab84e2) I used to fix the flake in the input-output test, yes it does not have a timeout but it does work more cases it seems than using a timeout (seems to trigger on false positives sometimes.).

src/content/datachannel/filetransfer/js/test.js
```diff
@@ -0,0 +1,34 @@
...
+test('Filetransfer via Datachannels', function(t) {
+  var driver = seleniumHelpers.buildDriver();
+
+  driver.get('file://' + process.cwd() +
+      '/src/content/datachannel/filetransfer/index.html')
+  .then(function() {
+    t.pass('page loaded');
+    // Based on https://saucelabs.com/resources/articles/selenium-file-upload
+    return driver.findElement(webdriver.By.id('fileInput'))
+       .sendKeys(process.cwd() + '/index.html');
```
KaptenJansson: WDYT about using a larger file? Or is that out of the scope as we are just testing that the page can transfer a file?

KaptenJansson:
Note a merge with master is required since I pushed a couple of PR's causing conflict.

src/content/datachannel/filetransfer/js/test.js
fippo: ah, testing four files of various sizes costs nothing virtually. And this tests that the 16k fragmentation works.

src/content/datachannel/filetransfer/js/test.js
fippo: I think this needs to check for visibility... let me see...

KaptenJansson: You can use the `elementVisible()` method on `webdriver.until` as well, more details [here](http://selenium.googlecode.com/git/docs/api/javascript/namespace_webdriver_until.html).

I just think the combo of `wait()` and `until()` is the way to go in general for elements.

KaptenJansson:
LGTM!

src/content/datachannel/filetransfer/js/test.js
fippo: elementIsVisible doesn't seem to work for unfortunately. And the current thing doesn't check for visibility, causing things to end early.

PR #605, opened 2015-07-29
Use wait+promise for videoelements

KaptenJansson:
Lets see if this approach fixes the flake for real.

It's funny that it only flakes out after I merge it to the master branch.........

### DIFF
--- a/src/content/devices/input-output/js/test.js
+++ b/src/content/devices/input-output/js/test.js
@@ -35,7 +35,7 @@ test('Fake device selection and check video tag dimensions in ' +
     // Chrome adds another fake video device.
     t.pass('Selecting 1st audio device');
     return driver.wait(driver.findElement(
-        webdriver.By.css('#audioSource>option')), 30 * 1000)
+        webdriver.By.css('#audioSource>option')), 60 * 1000)
     .then(function(element) {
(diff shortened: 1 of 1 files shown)


KaptenJansson:
Just realized it was already at 30 seconds and I bumped to 1 minute. Does it really take that long to enumerate devices?

KaptenJansson:
Trying to merge this and pray to the selenium and firefox gods ;).

KaptenJansson:
Yay it's green!

jan-ivar:
> Just realized it was already at 30 seconds and I bumped to 1 minute. Does it really take that long to enumerate devices?

That seems way excessive. Have you looked at using [MutationObserver](https://developer.mozilla.org/en-US/docs/Web/API/MutationObserver)? (I haven't tried it).

KaptenJansson:
Yes it was and it did not work, it does not seem like the timeout is actually enforced as it failed before the timeout was reached. Anyway I've removed the timeout altogether and use `wait()` and `until()` together instead.

samdutton:
FWIW I built a MutationObserver sample here: simpl.info/mo.

jan-ivar:
sweet!

PR #606, opened 2015-07-29
filetransfer: use mostly-spec getStats and fix test

fippo:
@KaptenJansson also shows how to use elementIsVisible. I had assumed one could call it directly from By.id() but you need to wrap it apparently.

Some of the field names used in getStats are not spec but removing the .stat calls is a step in the right direction.


KaptenJansson:
LGTM

PR #610, opened 2015-07-31
Use onbufferedamountlow, if available, in the data transfer demo

bemasc:
This functionality was just added to Chromium: https://crbug.com/496700

### DIFF
--- a/src/content/datachannel/datatransfer/js/main.js
+++ b/src/content/datachannel/datatransfer/js/main.js
@@ -27,6 +27,7 @@ var bitrateMax = 0;
 sendButton.onclick = createConnection;
 
 function createConnection() {
+  sendButton.disabled = true;
   var servers = null;
   pcConstraint = null;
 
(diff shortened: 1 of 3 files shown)


KaptenJansson:
Could you add a test for this? You can use https://github.com/webrtc/samples/blob/master/src/content/datachannel/filetransfer/js/test.js as a guide.

KaptenJansson:
@bemasc ping

KaptenJansson:
Thanks for the PR btw ;)

bemasc:
I think I'm more inclined to move this change to the filetransfer demo in that case.

KaptenJansson:
Sure, that's fine.

KaptenJansson:
Also test framework introduction can be found at https://github.com/webrtc/samples/tree/master/test.

bemasc:
OK, I've added a test for the datatransfer demo.

KaptenJansson:
LGTM with one nit, I would have preferred to have the test contained in the "test" function rather than calling an outside function (sendData()) however that can be changed if more tests are added.

PR #611, opened 2015-08-01
add getStats helper

fippo:
Things like https://github.com/webrtc/samples/pull/504 are tricky to test since getStats (to allow querying the codec name used) is async and we can only execute sync calls.

This adds a helper function which allows doing the following:
```
.then(function() {
  return seleniumHelpers.getStats(driver, 'name of peerconnection variable to inspect');
})
.then(function(stats) {
  console.log(stats);
})
```

### DIFF
--- a/test/selenium-lib.js
+++ b/test/selenium-lib.js
@@ -54,7 +54,7 @@ function getStats(driver, peerConnection) {
   // Execute getStats on peerconnection named `peerConnection`
   driver.executeScript(
       'window.' + tmp + ' = null;' +
-      peerConnection + '.getStats(null, function(report) {' +
+      peerConnection + '.getStats(null).then(function(report) {' +
       '  window.' + tmp + ' = report;' +
       '});');
   // Wait for result.
--- a/test/selenium-lib.js
+++ b/test/selenium-lib.js
@@ -49,7 +49,6 @@ function buildDriver() {
 // A helper function to query stats from a PeerConnection.
 function getStats(driver, peerConnection) {
   // Execute getStats on peerconnection named `peerConnection`.
-  console.log(driver.manage().timeouts());
   driver.manage().timeouts().setScriptTimeout(1000);
   return driver.executeAsyncScript(
       'var callback = arguments[arguments.length - 1];' +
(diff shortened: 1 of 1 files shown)


fippo:
needs an adapter.js with https://github.com/webrtc/adapter/pull/108 before merging. Now also works in Firefox

KaptenJansson:
I will push adapter 0.2.0 to NPM tomorrow and make a pull request here merging it.

jan-ivar:
I'm no selenium expert, so I may be missing something, but why is this needed? If selenium can synchronously get us back a promise from getStats, isn't that sufficient?

fippo:
@jan-ivar we need the result from the getStats call, not just the promise (which I doubt can be serialized in a meaningful way)

So we need to execute code in an async way and only have sync calls.

fippo:
need to re-evaluate the usage of executeAsyncScript after some discussion with jib

fippo:
I tried
```
  return driver.executeAsyncScript(
      'var callback = arguments[arguments.length - 1];' +
      peerConnection + '.getStats(null).then(function(report) {' +
      '  callback(report);' + 
      '});');
```
but that seems rather flaky and errors quite often since the default timeout seems to be 0 seconds?

test/selenium-lib.js
```diff
@@ -46,6 +46,18 @@ function buildDriver() {
...
 }
 
+// A helper function to query stats from a PeerConnection.
+function getStats(driver, peerConnection) {
+  // Execute getStats on peerconnection named `peerConnection`.
+  driver.manage().timeouts().setScriptTimeout(1000);
+  return driver.executeAsyncScript(
+      'var callback = arguments[arguments.length - 1];' +
+      peerConnection + '.getStats(null).then(function(report) {' +
+      '  callback(report);' +
```
KaptenJansson: Nit picky :( Remove space.

KaptenJansson:
@fippo anything else needed here?

test/selenium-lib.js
fippo: I don't see it. Which is probably due to the nature of whitespace, even with a green background :-)

fippo:
no, I have another test as followup which I could add here but lets keep things separate.

KaptenJansson:
LGTM

PR #612, opened 2015-08-03
use video.srcObject

fippo:
probably breaks for chrome < 43 but... @samdutton

also bumps adapter to 0.2.1 and removes the grunt githooks call because of https://github.com/webrtc/adapter/pull/113 (even though samples are not a npm package)

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,7 +20,7 @@
     "url": "https://github.com/webrtc/samples.git"
   },
   "scripts": {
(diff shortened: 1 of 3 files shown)


KaptenJansson:
LGTM.

KaptenJansson:
what are your thoughts @samdutton ?

KaptenJansson:
Merging this to bump adapter. Will see how it affects M43.

PR #614, opened 2015-08-11
fix wrong check, dont use webkitAudioContext

fippo:
I am not sure what 'wacx' means... also this doesn't work in Firefox at all, not sure what the intent was/is?

### DIFF
--- a/src/content/peerconnection/create-offer/js/main.js
+++ b/src/content/peerconnection/create-offer/js/main.js
@@ -26,8 +26,7 @@ numAudioTracksInput.onchange = function() {
 };
 
 var pc = new RTCPeerConnection(null);
-/* global webkitAudioContext */
-var wacx = new webkitAudioContext(); // jscs:ignore requireCapitalizedConstructors
+var wacx = new AudioContext();
 
 function createOffer() {
   var numRequestedAudioTracks = numAudioTracksInput.value;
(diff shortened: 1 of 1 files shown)


src/content/peerconnection/create-offer/js/main.js
```diff
@@ -49,7 +48,8 @@ function createOffer() {
     }]
   };
   // These constraints confuse Firefox, even if declared as optional.
-  if (webrtcDetectedBrowser !== 'Firefox') {
+  // See https://bugzilla.mozilla.org/show_bug.cgi?id=1184712
```
jan-ivar: I think you want https://bugzilla.mozilla.org/show_bug.cgi?id=1098015

jan-ivar: On second thought, this is createOffer, so there shouldn't be a problem. Are we sure this is needed?

src/content/peerconnection/create-offer/js/main.js
```diff
@@ -49,7 +48,8 @@ function createOffer() {
...
+  if (webrtcDetectedBrowser !== 'firefox') {
     offerConstraints.optional.push({
```
jan-ivar: At some point we should stop referring to RTOfferOptions as "constraints".

KaptenJansson:
wacx means webkit audio context and this demo was originally created before  "other browsers than chrome" were considered I think. Old versions of Chrome (before blink I believe) and Safari used this (not sure if they have fixed it yet). http://stackoverflow.com/a/29373891

KaptenJansson:
Should we just assume that the browsers we care about (WebRTC enabled) use AudioContext() now? If so then LGTM.

src/content/peerconnection/create-offer/js/main.js
KaptenJansson: I agree, `options` make more sense right?

fippo:
well, webkitAudioContext is deprecated in Chrome even so... :-)

KaptenJansson:
LGTM

PR #617, opened 2015-08-13
Enable FF beta and fix FF41 issue where it does not start

### LINKED ISSUE #583: Investigate why firefox unstable stalls on the first test
Running it locally results in the Firefox browser being started but stops at https://github.com/webrtc/samples/blob/master/src/content/getusermedia/gum/js/test.js#L24. Seems like the Promise callback is never triggered

KaptenJansson:
Fixes #583 

Note that Firefox unstable (FF42) still fails due to it no longer allows turning off the check for unsigned extensions. See the issue for more details.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,7 +17,6 @@ matrix:
 
   allow_failures:
     - env: BROWSER=chrome  BVER=unstable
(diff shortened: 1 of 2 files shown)


fippo:
LGTM

PR #618, opened 2015-08-13
test codec munging using getStats

fippo:
that was why I wanted getStats :-)


fippo:
TIL always make sure t.end() is called... not doing so leads to weird failures elsewhere.

KaptenJansson:
Nice test! LGTM.

PR #619, opened 2015-08-18
align bitrate and packet rate calculation with webrtc-stats example

fippo:
http://w3c.github.io/webrtc-stats/#example-1 showed a much nicer way to calculate the bit- and packetrates. In particular that way also works (without changes) when multiple tracks are used.

Oh... and this now uses promises just because.


KaptenJansson:
LGTM

PR #621, opened 2015-08-23
munge-sdp: keep offer/answer objects around

fippo:
IMO the offer and answer objects should be kept around and just the SDP should be replaced.
More detailed rationale should follow soon :-)

### DIFF
--- a/src/content/peerconnection/munge-sdp/js/main.js
+++ b/src/content/peerconnection/munge-sdp/js/main.js
@@ -25,7 +25,9 @@ createAnswerButton.onclick = createAnswer;
 setAnswerButton.onclick = setAnswer;
 hangupButton.onclick = hangup;
 
+var offer;
 var offerSdpTextarea = document.querySelector('div#local textarea');
+var answer;
 var answerSdpTextarea = document.querySelector('div#remote textarea');
 
 var audioSelect = document.querySelector('select#audioSrc');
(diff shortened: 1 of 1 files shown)


src/content/peerconnection/munge-sdp/js/main.js
```diff
@@ -183,10 +185,8 @@ function onCreateSessionDescriptionError(error) {
 function setOffer() {
   var sdp = offerSdpTextarea.value;
   sdp = maybeAddLineBreakToEnd(sdp);
-  var offer = new RTCSessionDescription({
```
KaptenJansson: Why do we not need to use the RTCSessionDescription interface anymore?

src/content/peerconnection/munge-sdp/js/main.js
```diff
@@ -183,10 +185,8 @@ function onCreateSessionDescriptionError(error) {
...
-    type: 'offer',
-    sdp: sdp
-  });
+  sdp = sdp.replace(/\n/g, '\r\n');
```
KaptenJansson: Is this done due to not using RTCSessionDescription anymore?

fippo: no, just to recreate the CRLF style which is mangled (hah) by the textarea. Now SDP parser are required to deal with both CRLF and LF but I prefer CRLF.

src/content/peerconnection/munge-sdp/js/main.js
fippo: we keep the original session description (including any extra things it may contain) around.

KaptenJansson: I see, do you know the historic reason for why we did not?

src/content/peerconnection/munge-sdp/js/main.js
KaptenJansson: I see. Thanks for explaining.

KaptenJansson:
LGTM

PR #622, opened 2015-08-24
s/iceTransports/iceTransportPolicy/

fippo:
since https://github.com/webrtc/adapter/pull/97 makes chrome do the right thing... it should be also working in Firefox now but doesn't for me?!


KaptenJansson:
LGTM

PR #631, opened 2015-08-29
audio demo: also support outbound-rtp in addition to outboundrtp

fippo:
I am not sure what the plan is WRT to outbound-rtp vs outboundrtp. The former is used by webrtc-pc and ortc, the latter by webrtc-stats. Either way, this should make things work, we can remove the wrong types at some point in the future :-)

### DIFF
--- a/src/content/peerconnection/audio/js/main.js
+++ b/src/content/peerconnection/audio/js/main.js
@@ -254,6 +254,7 @@ window.setInterval(function() {
       var packets;
       var now = report.timestamp;
       if ((report.type === 'outboundrtp') ||
+          (report.type === 'outbound-rtp') ||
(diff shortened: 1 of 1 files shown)


KaptenJansson:
For some reason I thought you deleted outboundrtp and added outbound-rtp??? I wonder why :)

Anyway LGTM.

jan-ivar:
The right one is "outboundrtp" AFAIK.

fippo:
well, we currently have one not-yet-compliant browser so having another one doesn't hurt :-)
And outboundrtp is the first choice

jan-ivar:
Yes, hopefully this will make it to a stats shim some day ;)

jan-ivar:
See https://github.com/w3c/webrtc-stats/issues/5#issuecomment-68800622

PR #632, opened 2015-08-30
use spec constraints in offers

fippo:
this removes most of the answer options (which are currrently only supported in Chrome and use the legacy format) and uses offer options instead. Not sure why answer options were used even, looks like copy-paste without thinking :-)

One thing that broke is the googUseRtpMux on the candidate gathering page but I suppose that should be updated to use the bundle/rtcp-mux constraints anyway @juberti @pthatcherg?

### DIFF
--- a/src/content/peerconnection/audio/js/main.js
+++ b/src/content/peerconnection/audio/js/main.js
@@ -30,10 +30,9 @@ var packetSeries;
 var lastResult;
 
 var sdpConstraints = {
-  'mandatory': {
-    'OfferToReceiveAudio': true,
-    'OfferToReceiveVideo': false
-  }
+  offerToReceiveAudio: 1,
+  offerToReceiveVideo: 0,
+  voiceActivityDecection: false
 };
 
 function gotStream(stream) {
@@ -48,7 +47,7 @@ function gotStream(stream) {
   trace('Adding Local Stream to peer connection');
 
   pc1.createOffer(gotDescription1, onCreateSessionDescriptionError,
-    {voiceActivityDetection: false});
+      sdpConstraints);
 
   bitrateSeries = new TimelineDataSeries();
   bitrateGraph = new TimelineGraphView('bitrateGraph', 'bitrateCanvas');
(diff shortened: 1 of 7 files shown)


src/content/peerconnection/audio/js/main.js
```diff
@@ -30,10 +30,9 @@ var packetSeries;
 var lastResult;
 
 var sdpConstraints = {
-  'mandatory': {
-    'OfferToReceiveAudio': true,
-    'OfferToReceiveVideo': false
-  }
+  offerToReceiveAudio: 1,
+  offerToReceiveVideo: 0,
+  voiceActivityDecection: false
```
KaptenJansson: voiceActivityDecection > voiceActivityDetection

src/content/peerconnection/audio/js/main.js
```diff
@@ -30,10 +30,9 @@ var packetSeries;
 var lastResult;
 
 var sdpConstraints = {
-  'mandatory': {
-    'OfferToReceiveAudio': true,
-    'OfferToReceiveVideo': false
-  }
+  offerToReceiveAudio: 1,
```
KaptenJansson: Nice, correct type;)

FTR http://www.w3.org/TR/2015/WD-webrtc-20150210/#dictionary-rtcofferoptions-members

KaptenJansson:
Also as an future improvement for all of us, when changing stuff due to spec changes or compliance, we should always link to the particular spec, i.e. source of truth. This makes it easier for everyone wanting to check up on the spec.

src/content/peerconnection/audio/js/main.js
fippo: hah. good point about calling this offerOptions, less ambiguous than 'sdpConstraints'

src/content/peerconnection/trickle-ice/js/main.js
```diff
@@ -94,11 +94,11 @@ function start() {
   // on whether the unbundle RTCP checkbox is checked.
   var config = {'iceServers': iceServers, iceTransportPolicy: iceTransports};
   var pcConstraints = {};
-  var offerConstraints = {'mandatory': {'OfferToReceiveAudio': true}};
+  var offerConstraints = {offerToReceiveAudio: 1};
```
KaptenJansson: For the sake of overall consistency, should this be named `offerOptions` as well?

fippo: +1 for overall consistency -- if grep misses it then that's bad. Fixed.

KaptenJansson:
LGTM

PR #633, opened 2015-08-31
Use partial match for fake camera device as workaround for chrome bug

KaptenJansson:
This will prevent local test run failures on Linux. 

This should be removed once http://crbug.com/526633 is fixed.


fippo:
LGTM -- finally tests are running again locally!

PR #635, opened 2015-09-02
Add datachannels to munge-sdp demo

### LINKED ISSUE #625: Add data channels to munge-sdp demo
It would be useful for testing data channels specifically.

KaptenJansson:
Do we really need a checkbox to toggle dataChannels or is it fine to always include it (like it does in this PR)?

Fixes #625

### DIFF
--- a/src/content/peerconnection/munge-sdp/index.html
+++ b/src/content/peerconnection/munge-sdp/index.html
@@ -64,7 +64,7 @@
         <h2>Offer SDP</h2>
         <textarea></textarea>
         <h2>DataChannel send</h2>
-        <textarea id="dataChannelSend" disabled>Data sent </textarea>
+        <textarea id="dataChannelSend" disabled>Data sent</textarea>
         <br>
         <br>
       </div>
--- a/src/content/peerconnection/munge-sdp/index.html
+++ b/src/content/peerconnection/munge-sdp/index.html
@@ -63,6 +63,8 @@
         <video autoplay muted></video>
         <h2>Offer SDP</h2>
         <textarea></textarea>
+        <h2>DataChannel send</h2>
+        <textarea id="dataChannelSend" disabled>Data sent </textarea>
         <br>
         <br>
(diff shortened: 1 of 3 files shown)


KaptenJansson:
I also fixed localStream.stop().

src/content/peerconnection/munge-sdp/index.html
```diff
@@ -63,6 +63,8 @@
         <video autoplay muted></video>
         <h2>Offer SDP</h2>
         <textarea></textarea>
+        <h2>DataChannel send</h2>
+        <textarea id="dataChannelSend" disabled>Data sent </textarea>
```
fippo: is the extra whitespace intentional here?

fippo:
LGTM -- the way you update the remote value almost implements a realtime chat. In the most inefficient way possible :-)

But i wonder if we really want to merge the datachannel chat demo into this... well, we can remove it when it becomes a maintenance issue.

KaptenJansson:
Haha, yeah well I just wanted some continuous data flowing, we can remove the send box and not let the user affect the data sent?

src/content/peerconnection/munge-sdp/index.html
KaptenJansson: No, it was lefter over when I had the counter on the same line (Data sent: ). Will remove.

fippo:
to see the non-effect this has now? +1 on removing the box and just sending something in the background

KaptenJansson:
Do you want both boxes gone or just the send one?

fippo:
oh.. both actually. Even though leaving the receive box makes this testable from selenium... hm..!

KaptenJansson:
Hmm, what about storing the data in a global variable instead? Or an invisible element not attach to the DOM?

fippo:
global variable (just a count of the amount of data received) sounds easiest.

KaptenJansson:
Fixed in #638

Issue #609, opened 2015-07-31
Move to ESLint for JavaScript linting and code style checking


samdutton:
I think we should probably move to ESLint for JavaScript linting: 

http://eslint.org
https://www.npmjs.com/package/grunt-eslint

Haven't used it myself, but it's quite popular among Chrome Dev Rel.

Pro and con:

http://blog.lauritz.me/linting-javascript-in-2015
http://www.sitepoint.com/comparison-javascript-linting-tools

Should be able to get rid of JSCS and JSHint if we opt for ESLint.

KaptenJansson:
SGTM

KaptenJansson:
@fippo WDYT?

KaptenJansson:
@samdutton did you plan adding this yourself?

samdutton:
>
> @samdutton <https://github.com/samdutton> did you plan adding this
> yourself?


Yes — though happy if anyone else wants to :).

samdutton:
Working on this...

PR #636, opened 2015-09-04
Fixing typo that breaks setOffer in the sdp munging sample.

taylor-b:
The 'cyrillic small letter dze' (http://www.fileformat.info/info/unicode/char/0455/index.htm) was somehow used instead of an 's'. So "offer.sdp = sdp" wasn't actually changing the sdp of the offer as intended.

### DIFF
--- a/src/content/peerconnection/munge-sdp/js/main.js
+++ b/src/content/peerconnection/munge-sdp/js/main.js
@@ -201,7 +201,7 @@ function setOffer() {
   var sdp = offerSdpTextarea.value;
   sdp = maybeAddLineBreakToEnd(sdp);
   sdp = sdp.replace(/\n/g, '\r\n');
-  offer.ѕdp = sdp;
+  offer.sdp = sdp;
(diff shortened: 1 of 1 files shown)


KaptenJansson:
Thanks for the PR.

Wow, what a catch ;). I wonder if this could be added to linting. @samdutton WDYT (WRT to #609)?

@fippo this was added in https://github.com/webrtc/samples/commit/4d00655f896db7613da307e94011a6b56ab55710, do you know what happened? Copy + paste from somewhere? ;).

FTR, I verified this by installing the "Character Table" plugin in sublime, copied the `ѕ` from `offer.ѕdp` to the Character table lookup function (ctrl+k then ctrl+l). Also you can notice that sublime (or any editor that preserves char encoding) cannot find "offer.sdp" when searching for it unless you copy the cyrillic letter ;).

char

fippo:
argh... not the first time this happens to me... vim going mad. Impossible to spot in codereview.

PR #638, opened 2015-09-04
Remove text boxes for datachannel transfer in munge-sdp demo

KaptenJansson:
PTAL.

How do you feel about the console logging spam in sendData and onReceiveMessageCallback?


fippo:
LGTM -- the log spam is only an issue if your console keeps autoscrolling and scrolling up a bit is the best solution

PR #646, opened 2015-09-18
preliminary update of adapter with edge support

fippo:
pending some discussion of whether ORTC should be shimmed in adapter I would like the demos to work in Edge. This makes a bunch of them work:
 * pc1
 * audio
 * multiple
 * munge-sdp (kind of... sdp munging is not supported. There is a limit how far I want to go)
 * constraints
 * states
 * trickle-ice (need to recheck the TURN support in Edge)

### DIFF
--- a/src/js/adapter.js
+++ b/src/js/adapter.js
@@ -11,7 +11,9 @@
    eqeqeq: true, forin: false, globalstrict: true, node: true,
    quotmark: single, undef: true, unused: strict */
 /* global mozRTCIceCandidate, mozRTCPeerConnection, Promise,
-mozRTCSessionDescription, webkitRTCPeerConnection, MediaStreamTrack */
+mozRTCSessionDescription, webkitRTCPeerConnection, MediaStreamTrack,
+MediaStream, RTCIceGatherer, RTCIceTransport, RTCDtlsTransport,
+RTCRtpSender, RTCRtpReceiver*/
 /* exported trace,requestUserMedia */
 
 'use strict';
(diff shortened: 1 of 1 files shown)


alvestrand:
My biggest issue is that it increases the size of adapter.js from 487 lines to more than 1500, and the >1000 added lines can't be tested by our current CI solution.

Are there any solutions that avoid this bloat, without destroying the simple "script src=adapter.js" usage?

I'd even be willing to consider solutions that generate the adapter.js file from an (npm-included) ortc-adapter.js file and a non-edge-adapter.js file....

jan-ivar:
Getting ORTC working with 1000 lines of code is actually quite a feat, considering other attempts at least twice the size.

fippo:
I agree that a small-module approach would be better. I'll see if I can generate a browserified adapter that is based on a adapter+edge module which requires webrtc-adapter-test and adds Edge support.

WRT to testing the samples this doesn't make this worse. We have Chrome and Edge covered and Edge can consider Edge experimental.  Actually I just noticed that the munge-demo broke since it now uses datachannels so I need to investigate whether the issue I ran into with the edge webdriver is still a thing...

fippo:
I just did an experiment:
1) moved all ORTC code to adapter-edge.js
2)  required adapter in adapter-edge, set module.exports to that
3) made attachMediaStream a function on window to break out of the module. Same for trace.
4) Browserified adapter-edge, put it into this repo and ran the tests
5) watched the device test fail because it queries webrtcDetectedBrowser from JS.

I think trace() does not belong into adapter.js, it is something the samples should share. attachMediaStream and reattachMediaStream should disappear as well imo. I can make it happen for the samples. And the test should be easy to fix. Does that look like the way to move this forward for the samples?
I think there is still some work to be done in https://github.com/webrtc/adapter/pull/136 but that is a different issue :-)

samdutton:
>
> I think trace() does not belong into adapter.js, it is something the
> samples should share.


I agree. What do you think is best — just include the code in a separate
.js file?


> attachMediaStream and reattachMediaStream should disappear as well imo.


Agree. As it is, the code is more complex than without these functions.


> I can make it happen for the samples.
>

Yes please!

samdutton:
@KaptenJansson @juberti:  Any objections to merging this? 

I don't have Edge to test with at the moment.

KaptenJansson:
@fippo WRT to 1, are you planning on committing this?
WRT to 5, are you planning on adding a work around for this? Or do you think we should handle that test differently?

PR #647, opened 2015-09-18
Use { exact: } on deviceId constraints (suppress FF doorhanger choices)

jan-ivar:
In-content selectors should use the { [`exact`](https://developer.mozilla.org/en-US/docs/Web/API/MediaDevices/getUserMedia#Parameters): } keyword in constraints, to make the request mandatory.

Several people got confused by http://webrtc.github.io/samples/src/content/devices/input-output/ because the Firefox doorhanger gives them choices, letting them override the in-content choice (optional constraints only affect the default choice in Firefox). This fixes that.

### DIFF
--- a/src/content/devices/input-output/js/main.js
+++ b/src/content/devices/input-output/js/main.js
@@ -94,8 +94,8 @@ function start() {
   var audioSource = audioInputSelect.value;
   var videoSource = videoSelect.value;
   var constraints = {
-    audio: {deviceId: audioSource? { exact: audioSource } : undefined },
-    video: {deviceId: videoSource? { exact: videoSource } : undefined }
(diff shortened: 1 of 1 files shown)


fippo:
+1 -- but jshint doesn't like your whitespace style :-)

samdutton:
LGTM.

xdumaine:
@jan-ivar @samdutton @fippo This is great - I was able to get this working for me with this format as well. Could you link me to where this constraint syntax is defined? I wasn't able to find reference to it anywhere.

xdumaine:
And just to clarify - this only works in FF 42+, right?

jan-ivar:
If you mean `deviceId`, then yes it only does something starting in FF 42. The `exact` keyword works back to FF 38.

PR #648, opened 2015-09-21
Add options page for Network Limiter Extension.

guoweis:


src/content/extensions/multipleroutes/src/options.js
```diff
@@ -0,0 +1,87 @@
...
+
+  chrome.privacy.network.webRTCMultipleRoutesEnabled.set({
+    'value': multiRoutes
+  });
+  try {
+    chrome.privacy.network.webRTCNonProxiedUdpEnabled.set({
+      'value': nonProxiedUdp
+    });
+  } catch (err) {
+    // Shouldn't run into here.
```
pthatcherg: Should we at least console.log something?

src/content/extensions/multipleroutes/src/options.js
```diff
@@ -0,0 +1,87 @@
...
+
+// Saves options.
+function saveOptions() {
+  var multiRoutes;
+  var nonProxiedUdp;
+  var radios = document.getElementsByName('routeselection');
+  for (var i = 0, length = radios.length; i < length; i++) {
+    if (radios[i].checked) {
+      multiRoutes = i < 1;
+      nonProxiedUdp = i < 2;
```
pthatcherg: Seems like a named constant or two would go a long way to readability:

var multiRouteIndex = 0;
var nonProxiedUdpIndex = 1;

This would help readability here, and then below when you restore the radio buttons.

src/content/extensions/multipleroutes/src/options.js
```diff
@@ -0,0 +1,87 @@
...
+  }
+}
+
+function restoreRadios(multiRoutes, nonProxiedUdp) {
+  var radios = document.getElementsByName('routeselection');
+  if (multiRoutes) {
+    if (nonProxiedUdp) {
+      radios[0].checked = true;
+    } else {
+      radios[0].checked = false;
```
pthatcherg: What effect will .checked = false have here?

src/content/extensions/multipleroutes/src/options.js
```diff
@@ -0,0 +1,87 @@
...
+      alert(
```
pthatcherg: Maybe if multiRoutes: true, nonProxiedUdp: false  comes up, we should just downgrade to multiRoutes: false, nonProxiedUdp: false.

src/content/extensions/multipleroutes/src/options.js
```diff
@@ -0,0 +1,87 @@
...
+        nonProxiedUdp = details.value;
+        restoreRadios(multiRoutes, nonProxiedUdp);
+      });
+    document.getElementById('for_multirouteOffUdpOff_notSupported').innerHTML =
+      '';
+  } catch (err) {
+    document.getElementById('multirouteOffUdpOff').disabled = true;
+    var chromeVersion = /Chrome\/([0-9.]+)/.exec(navigator.userAgent)[1];
+    document.getElementById('for_multirouteOffUdpOff_notSupported').innerHTML +=
+      ' Current version is: ' + chromeVersion;
```
pthatcherg: If you use +=, could this possibly end up appending the version over and over?

src/content/extensions/multipleroutes/src/options.js
```diff
@@ -0,0 +1,87 @@
...
+    });
+  try {
+    chrome.privacy.network.webRTCNonProxiedUdpEnabled.get({},
+      function(details) {
+        nonProxiedUdp = details.value;
+        restoreRadios(multiRoutes, nonProxiedUdp);
+      });
+    document.getElementById('for_multirouteOffUdpOff_notSupported').innerHTML =
+      '';
+  } catch (err) {
```
pthatcherg: Why do we try/catch one of the settings but not the other?

src/content/extensions/multipleroutes/src/options.js
guoweis: done.

src/content/extensions/multipleroutes/src/options.js
guoweis: we have the minimum chrome version specified so multipleRoutes is always supported.

src/content/extensions/multipleroutes/src/options.js
guoweis: no. at least not in the screen shot that I shared.

src/content/extensions/multipleroutes/src/options.js
guoweis: If this happens, the customer has gone through the pain to modify preferences file. I think we should just leave as it is.

src/content/extensions/multipleroutes/src/options.js
guoweis: that radio box will be disabled.

src/content/extensions/multipleroutes/src/options.js
guoweis: I added a comment. Let's see if this is clear.

pthatcherg:
lgtm

guoweis:
anything else before this can be accepted?

pthatcherg:
Nothing from me

src/content/extensions/multipleroutes/src/options.js
KaptenJansson: What's the rationale of adding the string to the element rather than just assigning it every time?

KaptenJansson:
LGTM

guoweis:
Hmm, I can't find this change in https://github.com/webrtc/samples. How was this merged?

KaptenJansson:
It was merged to master but then @samdutton merged master into gh-pages so we do not have do an extra step merging to gh-pages (this branch is served by the github web server) for every change.

It should still be in your forked repo though.

guoweis:
so we should resubmit or remerge these? What should I do at this point?

Thanks,
Guowei

PR #651, opened 2015-09-22
remove attachMediaStream usage

fippo:
second part of the change proposed in https://github.com/webrtc/samples/pull/646#issuecomment-142157106 -- removes attachMediaStream. And some comments related to it.

Even the odd comment in peerconnection/audio that claims attachMediaStream is used even though it is not.


src/content/peerconnection/constraints/js/main.js
```diff
@@ -158,7 +158,7 @@ function createPeerConnection() {
   };
   remotePeerConnection.onaddstream = function(e) {
     console.log('remotePeerConnection got stream');
-    attachMediaStream(remoteVideo, e.stream);
+    remoteVideo = e.stream;
```
samdutton: This should be `remoteVideo.srcObject = e.stream` (shouldn't it?)

samdutton:
LGTM apart from comment.

src/content/peerconnection/constraints/js/main.js
fippo: good catch

fippo:
looks like jshintrc needs a more general cleanup. Removed some obsolete stuff but there is still some things which I don't know about. Well... jshint still works.

samdutton:
LGTM.

samdutton:
Pushed to gh-pages — LGTM.

PR #656, opened 2015-10-05
Improve setup for new developers …

xdumaine:
... by including a simple server script and using relative links. I noticed all branches of this (including localhost and @fippo's [branch](http://fippo.github.io/webrtc/) have links back to this repo, instead of relative. I used relative links to simplify. Also added a simple script to do the grunt work (excuse the pun) of setting up a simple static server for devs over https. This is especially useful with upcoming Chrome release which will stop gUM from working off of https.

Now, to be off and running, all that's needed is `npm install && node server.js`

PR #657, opened 2015-10-05
fix samples link for gh-pages

fippo:
#656 broke the links for https://webrtc.github.io/samples/ actually.
Not sure i'm happy about serving '..' from express but...

### DIFF
--- a/index.html
+++ b/index.html
@@ -76,63 +76,63 @@
 
       <h3 id="getusermedia">getUserMedia</h3>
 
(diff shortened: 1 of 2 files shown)


xdumaine:
@fippo [this PR](https://github.com/webrtc/samples/pull/658) should fix them for both localhost and gh pages without having to serve `..` from express. Thoughts?

fippo:
I like your fix more. Thanks!

PR #658, opened 2015-10-06
fix relative links

xdumaine:
Don't append `/` to relative links to ensure they don't remove path.


xdumaine:
@fippo this should let them work for your fork as well.

KaptenJansson:
LGTM, sorry was to fast, I did not actually test it :( My bad.

PR #664, opened 2015-10-19
dont use webrtcDetectedBrowser

fippo:
'cause https://github.com/webrtc/adapter/pull/146

easier to use process.env.BROWSER @KaptenJansson :-)


KaptenJansson:
The rationale to remove it from all tests here is that we have an explicit test in the adapter.js repo for this (in addition to webrtc/adapter#146)?

fippo:
no, we won't have it available anymore in non-module cases. But it wasn't used much.

KaptenJansson:
I see, LGTM.

PR #691, opened 2015-11-18
Update create offer constraints and bump travis-multirunner to 3.0 and enable FF

KaptenJansson:
Not sure if the pranswer test page actually works due to a webrtc bug in Chrome and FF has not implmeneted pranswer yet.

However I think it makes sense to setup the expectations in the offer due offerToReceiveVideo/Audio are not valid createAnswer constraints (only voiceActivityDetection is a valid answer and offer constraint according to http://w3c.github.io/webrtc-pc/#idl-def-RTCAnswerOptions)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,7 +21,6 @@ matrix:
 
   allow_failures:
     - env: BROWSER=chrome  BVER=unstable
-    - env: BROWSER=firefox BVER=beta
     - env: BROWSER=firefox BVER=nightly
 
 before_script:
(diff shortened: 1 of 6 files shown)


fippo:
pranswer should work in 48, see webrtc issue 3349.
LGTM, Firefox beta test failure is not related to the changes here.

KaptenJansson:
Right, I haven't been following the progress on pranswer, I just tested on Chrome 47. 

The firefox issue is a transient error, need to look into it and see if we can fix it in the test

Note I just reran the test and now it's green ;)

KaptenJansson:
Yay, can confirm that Pranswer works in M48!

PR #693, opened 2015-11-20
Update Network Limiter for Chrome M48

guoweis-work:


KaptenJansson:
You can run all the lint checks offline if you do `npm install` locally and then run them via `grunt`.

To make eslint happy in this case, add getPolicyFromBooleans to https://github.com/webrtc/samples/blob/gh-pages/.eslintrc#L49.

src/content/extensions/multipleroutes/src/_locales/en/messages.json
```diff
@@ -1,15 +1,18 @@
 {
-   "NETLI_MULTIROUTEONUDPON_RADIO": {
+   "NETLI_DEFAULT_RADIO": {
```
pthatcherg, line 2: Should we call this something without the word "default" in it, like "NETLI_BEST_MEDIA_RADIO"?

src/content/extensions/multipleroutes/src/eventPage.js
```diff
@@ -1 +1,74 @@
...
+// to avoid repeated conversions and overwrite the current setting.
+function resetOldBooleans(callback) {
+  pn.webRTCNonProxiedUdpEnabled.set({
+    value: true
+  }, function() {
+    pn.webRTCMultipleRoutesEnabled.set({
+      value: true
+    }, function() {
+      if (callback) {
+        callback('Successfully reset the boolean');
```
pthatcherg: boolean => booleans?

pthatcherg: Why are we hardcoding a message in this function?  Why not just fire the callback and let the callback handle the message?

src/content/extensions/multipleroutes/src/eventPage.js
```diff
@@ -1 +1,74 @@
...
+      }
+    });
+  });
+}
+
+// Converts the old booleans to the new policy in Preferences and restores the 2
+// previous booleans to the default. Future chrome updates could trigger this
```
pthatcherg, line 42: restores => resets?

src/content/extensions/multipleroutes/src/eventPage.js
```diff
@@ -1 +1,74 @@
...
+        );
+      }
+      return;
+    }
+
+    getPolicyFromBooleans(isInstall, function(policy) {
+      pn.webRTCIPHandlingPolicy.set({
+        value: policy
+      }, resetOldBooleans(callback));
+    });
```
pthatcherg, line 71: Doesn't this need to be "function() { resetOldBooleans(callback) }" ?

src/content/extensions/multipleroutes/src/eventPage.js
```diff
@@ -1 +1,74 @@
...
+// the previous booleans have been restored to default, it'll be translate to
+// "default" again.
+function convertBooleansToPolicy(isInstall, callback) {
+  if (pn.webRTCIPHandlingPolicy === undefined) {
+    return;
+  }
+
+  pn.webRTCIPHandlingPolicy.get({}, function(details) {
+    if (details.value !== chrome.privacy.IPHandlingPolicy.DEFAULT) {
+      if (callback) {
```
pthatcherg, line 54: Other places you don't check for a null callback.  Should you?

src/content/extensions/multipleroutes/src/eventPage.js
```diff
@@ -1 +1,74 @@
...
+        callback(
+          'webRTCIPHandlingPolicy has a non-default value, stop now.'
+        );
```
pthatcherg, line 57: Again, would it make sense to let the callee take care of the messages?

src/content/extensions/multipleroutes/src/_locales/en/messages.json
pthatcherg, line 2: Or maybe NETLI_ALLOW_MULTIPLE_ROUTES would match the code down below.

src/content/extensions/multipleroutes/src/utils.js
```diff
@@ -0,0 +1,48 @@
...
+    }
+    return pi.DEFAULT;
+  }
+
+  if (allowUdp) {
+    return pi.DEFAULT_PUBLIC_INTERFACE_ONLY;
+  }
+
+  return pi.DISABLE_NON_PROXIED_UDP;
+}
```
pthatcherg, line 47: This might be more readable with more early returns:

if (!allowUdp) {
  return pi.DISABLE_NON_PROXIED_UDP;
}

if (!allowMultiRoute) {
  return pi.DEFAULT_PUBLIC_INTERFACE_ONLY;
}

if (isInstall && pn.webRTCIPHandlingPolicy !== undefined) {
  return pi.DEFAULT_PUBLIC_AND_PRIVATE_INTERFACES;
}

return pi.DEFAULT;

src/content/extensions/multipleroutes/src/utils.js
```diff
@@ -0,0 +1,48 @@
...
+
+pi = chrome.privacy.IPHandlingPolicy;
+
+// Helper function to convert the parameters to policy synchronously.
+function convertToPolicy(allowMultiRoute, allowUdp, isInstall) {
+  if (allowMultiRoute) {
+    if (isInstall && pn.webRTCIPHandlingPolicy !== undefined) {
+      // If we're installing the extension, we'll default to a more private
+      // mode.
+      return pi.DEFAULT_PUBLIC_AND_PRIVATE_INTERFACES;
```
pthatcherg: Why does the output of the convert function check pn.webRTCIPHandlingPolicy?  Shouldn't the convert function always return the same thing regardless of pn.webRTCIPHandlingPolicy?

src/content/extensions/multipleroutes/src/utils.js
```diff
@@ -0,0 +1,48 @@
...
+    }
```
pthatcherg: Would it make more sense to have convertToPolicy just return pi.DEFAULT even when it is an install, and then the layer of the code that knows it's an install do the following?

var policy = convertToPolicy(...);
if (isInstall && policy == pi.DEFAULT) {
  policy = pi.DEFAULT_PUBLIC_AND_PRIVATE_INTERFACES.
}

src/content/extensions/multipleroutes/src/options.js
```diff
@@ -1,97 +1,114 @@
...
+var pi = chrome.privacy.IPHandlingPolicy;
+
+var mapPolicyToRadioId = {};
+mapPolicyToRadioId[pi.DEFAULT] = 0;
+mapPolicyToRadioId[pi.DEFAULT_PUBLIC_AND_PRIVATE_INTERFACES] = 1;
+mapPolicyToRadioId[pi.DEFAULT_PUBLIC_INTERFACE_ONLY] = 2;
+mapPolicyToRadioId[pi.DISABLE_NON_PROXIED_UDP] = 3;
+
+var mapRadioIdToPolicy = {};
+if (pn.webRTCIPHandlingPolicy === undefined) {
```
pthatcherg: Can you make a helper method called something like browserSupportsPolicy() just to be a little more readable and use that instead of "pn.webRTCIPHandlingPolicy === undefined"?

src/content/extensions/multipleroutes/src/options.js
```diff
@@ -1,97 +1,114 @@
...
+
 // Saves options.
 function saveOptions() {
-  var multiRoutes;
-  var nonProxiedUdp;
-  var radios = document.getElementsByName('routeselection');
-  for (var i = 0, length = radios.length; i < length; i++) {
+  var radios = document.getElementsByName('ip_policy_selection');
+  var i = 0;
+  for (; i < radios.length; i++) {
```
pthatcherg: Putting "var i = " in the for loop is the same thing,  It's JS, so it's all function level anyway.

src/content/extensions/multipleroutes/src/options.js
```diff
@@ -1,97 +1,114 @@
...
-  try {
-    chrome.privacy.network.webRTCNonProxiedUdpEnabled.set({
-      'value': nonProxiedUdp
+  if (pn.webRTCIPHandlingPolicy !== undefined) {
+    pn.webRTCIPHandlingPolicy.set({
+      value: mapRadioIdToPolicy[i]
+    });
+  } else {
+    var mapping = mapRadioIdToPolicy[i];
+    pn.webRTCMultipleRoutesEnabled.set({
```
pthatcherg, line 49: Might be more readable as:

var oldBools = mapRadioIdToPolicy[i];
...
value: oldBools.allowMultipleRoutes
value: oldBools.allowUdp

Then you need to change your [true, false], etc above into {allowMultipleRoutes: true, allowUdp: false}, but that would be more readable, I think.

src/content/extensions/multipleroutes/src/options.js
```diff
@@ -1,97 +1,114 @@
...
-      ' Current version is: ' + chromeVersion;
+// Returns the supported modes in the UI option.
+function getSupportedModes() {
+  if (pn.webRTCNonProxiedUdpEnabled === undefined) {
+    return [true, false, true, false];
   }
+  if (pn.webRTCIPHandlingPolicy === undefined) {
+    return [true, false, true, true];
+  }
+  return [true, true, true, true];
```
pthatcherg: Might be more readable returning a dictionary/object:

var supports {
  foo: true,
  bar: true,
  baz: true,
  yolo: true
}

if (!browserSupportsYolo()) {
  supports.yolo = false;
}
if (!browserSupportsBar()) {
  supports.bar = false;
} 

return supports;

(please fill in the right names for foo, bar, baz, ...)

pthatcherg: Maybes it's Mode1, Mode2, Mode3, Mode4, etc, to match the element ID.  That might make the mode hiding stuff below easier.

src/content/extensions/multipleroutes/src/utils.js
```diff
@@ -0,0 +1,48 @@
...
+  if (allowUdp) {
+    return pi.DEFAULT_PUBLIC_INTERFACE_ONLY;
+  }
+
+  return pi.DISABLE_NON_PROXIED_UDP;
+}
+
+// This function just returns the new policy value based on the 2 booleans
+// without changing any preferences.
+getPolicyFromBooleans = function(isInstall, callback) {
```
guoweis-work: without var in the beginning, chrome doesn't install this extension. Wtih var, grunt gives me error "no-unused-var"

KaptenJansson: Should definitely use var there. Seems like grunt is not picking up the .eslintrc file. To rule out any local grunt issues you might have, could you add back var and push it again?

src/content/extensions/multipleroutes/src/utils.js
```diff
@@ -0,0 +1,48 @@
+'use strict';
```
KaptenJansson, line 12: Missing copyright header here

src/content/extensions/multipleroutes/src/utils.js
KaptenJansson: Maybe it's an eslintrc environment issue?https://github.com/eslint/eslint/issues/4051

fippo: uhm... why not just use
`function getPolicyFromBooleans(...)`
alternatively, window.getPolicyFromBooleans is the other hack to stop such complaints ;-)

src/content/extensions/multipleroutes/src/_locales/en/messages.json
guoweis-work, line 2: This was with an older version.

src/content/extensions/multipleroutes/src/eventPage.js
guoweis-work, line 42: Done.

src/content/extensions/multipleroutes/src/eventPage.js
guoweis-work: the callback is really just a logging facility.

src/content/extensions/multipleroutes/src/eventPage.js
guoweis-work, line 71: I guess currying is allowed in javascript. It worked for me here.

src/content/extensions/multipleroutes/src/eventPage.js
guoweis-work, line 54: Removed all the checking.

src/content/extensions/multipleroutes/src/utils.js
guoweis-work, line 47: Done.

src/content/extensions/multipleroutes/src/utils.js
guoweis-work: This function serves as the purpose of mapping the old settings to the new world (even in older version of chrome). Since public_and_private_interfaces doesn't exist, we can't just return this out.

src/content/extensions/multipleroutes/src/utils.js
guoweis-work: Agree. Modified.

src/content/extensions/multipleroutes/src/options.js
guoweis-work: Done.

src/content/extensions/multipleroutes/src/options.js
guoweis-work: Done

src/content/extensions/multipleroutes/src/options.js
guoweis-work, line 49: Done

src/content/extensions/multipleroutes/src/options.js
guoweis-work: Done

KaptenJansson:
LGTM

PR #694, opened 2015-11-20
Add extra is visible check to see if it solves firefox issue

KaptenJansson:
The input output test fails on Firefox but only on travis, not locally. Try to solve with an extra check to see if the element is visible.


KaptenJansson:
Comitting to get a green tree, please comment anyway and I will fix it later.

PR #699, opened 2015-11-24
Fix broken selector cloning

### LINKED ISSUE #698: /content/devices/multi demo doesn't work
In Chrome 46.0.2490.86 m I get the error

```
Error:  TypeError: alloutputSelectors.forEach is not a function(…)
```

The problem is [here](https://github.com/webrtc/samples/blob/gh-pages/src/content/devices/multi/js/main.js#L41). `document.querySelectorAll` returns a `NodeList`, not an array, and `NodeList`s do not have a `forEach` method.

KaptenJansson:
Fixes #698

### DIFF
--- a/src/content/devices/multi/js/main.js
+++ b/src/content/devices/multi/js/main.js
@@ -37,12 +37,13 @@ function gotDevices(deviceInfos) {
   }
 
   // Clone the master outputSelector and replace outputSelector placeholders.
-  var alloutputSelectors = document.querySelectorAll('select');
-  alloutputSelectors.forEach(function(selector) {
+  var allOutputSelectors = document.querySelectorAll('select');
+  for (var selector = 0; selector < allOutputSelectors.length; selector++) {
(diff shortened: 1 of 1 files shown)


KaptenJansson:
We should add a test for this but it's hard to do that when the feature is behind a flag...We can ofc try to enable said flag when setting up the browser in webdriver.

samdutton:
Thanks for fixing — should have spotted this myself :/. 

LGTM and tested in Chrome 46.0.2490.86.

BTW: I get a DOMException if I attempt to select audio output for gumAudio before pressing play. (What does a greyed-out play button mean?) The stream is set as the `srcObject` for both gumAudio and gumVideo, but does that actually happen?  Not sure if we can fix this — or maybe the selector for gumAudio should be disabled before play starts.

KaptenJansson:
I believe the DOMexception is a chrome issue that has been fixed after M46, it does not happen in M48 (probably fixed in M47 as well but not 100%).

PR #702, opened 2015-11-26
Fix eslint

KaptenJansson:
It turns out eslint ignores global and export options when using env node (http://eslint.org/docs/rules/no-unused-vars.html#exporting-variables).

Also I think it's better to specify node environments per file as they are in general used for test framework stuff, not the samples themselves.

I found this looking into why @guoweis is having so much trouble in #693.

After this PR your should when exporting a global variable/function to be used in a different file declare it at top of the file as: `/* exported variableName */` and then in the file you call upon it declare it as /\* `global variableName */`. If they are immutable, add :false after the variableName, e.g. `/* global variableName:false */`.

You can also add the global variable in the global object in .eslintrc like we have done before but I feel we should move away from it I like the explicit export and "import" per file better, WDYT? If you agree, I will follow up in a separate PR cleaning up the rest, an exception might be adapter.js or other imported JS files.


KaptenJansson:
Check the first commit for the important bits, the 2nd is just copyright year update (I could not help myself;))

samdutton:
I haven't actually run this, but LGTM.

> After this PR your should when exporting a global variable/function to be used in a different file declare it at top of the file as: /\* exported variableName _/ and then in the file you call upon it declare it as /_ global variableName _/. If they are immutable, add :false after the variableName, e.g. /_ global variableName:false */.

Why 'after this PR'?

> You can also add the global variable in the global object in .eslintrc like we have done before but I feel we should move away from it I like the explicit export and "import" per file better, WDYT? If you agree, I will follow up in a separate PR cleaning up the rest, an exception might be adapter.js or other imported JS files.

I think you're right. The .eslintrc declarations are probably better done locally. This is more explicit and possibly less prone to error.

KaptenJansson:
> Why 'after this PR'?

Due to node being enabled globally which disables eslint exported and global functions. This PR disables Node environment globally, then add them to the files that actually need it.

samdutton:
> Due to node being enabled globally which disables eslint exported and global functions.

Is [the documentation](http://eslint.org/docs/rules/no-unused-vars.html#exporting-variables) you linked to saying that, or that it's not possible to set `/* exported */` values in Node environments? (I think it's unclear and haven't tried it out.)

KaptenJansson:
It does say in that documentation:
Note that/\* exported */ has no effect when used with
the node or commonjs environments or when ecmaFeatures.modules is true

samdutton:
OK — happy if you want to merge this.

Good call with the copyright year, too. (We should probably automate this somehow, as I think we discussed at some point.)

PR #703, opened 2015-11-26
record demo: use adapter + srcObject

fippo:
just because...

Oddly, in play() one can't use srcObject so createObjectUrl still has to be used. The spec looks like doing that for blobs should be possible but that does not seem to be implemented yet -- https://crbug.com/506273)

### DIFF
--- a/src/content/getusermedia/record/index.html
+++ b/src/content/getusermedia/record/index.html
@@ -52,6 +52,7 @@
 
   </div>
 
+  <script src="../../../js/adapter.js"></script>
   <script src="js/main.js"></script>
   <script src="../../../js/lib/ga.js"></script>
 
--- a/src/content/getusermedia/record/index.html
+++ b/src/content/getusermedia/record/index.html
@@ -52,6 +52,7 @@
 
   </div>
 
+  <!-- include adapter for srcObject shim -->
   <script src="../../../js/adapter.js"></script>
   <script src="js/main.js"></script>
   <script src="../../../js/lib/ga.js"></script>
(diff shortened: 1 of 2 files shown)


samdutton:
Thanks for this — good spot!

I'm in two minds about using adapter.js for this demo. It shims against API differences, but it also introduces a layer of 'magic' and complexity for the developer trying to understand what's going on. I love that the MediaRecorder API is so simple and understandable — that this demo is complete in one JavaScript file with no more than about a hundred lines of code.

@KaptenJansson WDYT?

Whatever, we should definitely update `play()` (or add just the `srcObject` shim).

fippo:
yeah... but adapter is only used for the srcObject here. We don't want people to use the createObjectURL stuff anymore.

Making the srcObject shim deal with Blobs sounds like a  good idea though :-)

fippo:
updating play() to use srcObject works in Chrome with the existing shim but passing a Blob to srcObject does not work in Firefox (should be shim-able) cc @mairereavy @jan-ivar

KaptenJansson:
This LGTM however maybe we should add a comment that srcObject is shimmed via adapter.js in this demo (maybe it's obvious due to including adapter.js)?

samdutton:
I don't mind all that much if we include adapter.js, but I think we should add a comment in the HTML ('adapter.js included only to shim srcObject.') and in the JavaScript ('Shimmed in adapter.js').

KaptenJansson:
SGTM

samdutton:
@fippo Could you add comments to the code? Thanks.

fippo:
done ;-)

samdutton:
LGTM — thanks!

PR #723, opened 2016-01-04
Tweaks to MediaRecorder example: redundant asserts, etc.

samdutton:
Various updates to cope with different codecs, etc. Also removed `console.assert()` statements, which are now redundant (@miguelao says).

Live [here](https://samdutton.github.io/webrtc/src/content/getusermedia/record/).

PR includes minor tweak to utils.js to remove redundant linting declarations.

### DIFF
--- a/src/content/getusermedia/record/js/main.js
+++ b/src/content/getusermedia/record/js/main.js
@@ -19,9 +19,6 @@ var mediaRecorder;
 var recordedBlobs;
 var sourceBuffer;
 
-navigator.getUserMedia = navigator.getUserMedia ||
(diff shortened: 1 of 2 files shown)


src/content/getusermedia/record/js/main.js
```diff
@@ -89,24 +81,29 @@ function toggleRecording() {
 }
 
 function startRecording() {
+  var options = {mimeType: 'video/vp9'};
```
KaptenJansson, line 84: Does this mean VP9 will always be used? Will it fall back to VP8 if VP9 is not supported (older Chrome or Firefox)?

KaptenJansson:
LGTM

PR #728, opened 2016-01-06
add ice restart example

fippo:
@pthatcherg -- let's document this :-)

I am not really happy that it is quite hard to detect the ice restart at pc2.
pc1 triggers an iceconnectionstatechange to connected (or completed), but pc2 does not which, without onselectedcandidatepairchange, makes it very hard to determine a restart happened. Polling getStats is not an option. Shall I file an issue?

### DIFF
--- a/index.html
+++ b/index.html
@@ -124,6 +124,8 @@
 
       <p><a href="src/content/peerconnection/trickle-ice/">ICE candidate gathering from STUN/TURN servers</a></p>
 
+      <p><a href="src/content/peerconnection/restart-ice/">Do an ICE restart</a></p>
+
       <p><a href="src/content/peerconnection/webaudio-input/">Web Audio output as input to peer connection</a></p>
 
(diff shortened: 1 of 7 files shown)


fippo:
oops, test.js is simply copied from pc1, should fix before this gets merged!

samdutton:
Nice one!

LGTM.

fippo:
added a proper test. Finally I had a chance to put the getStats stuff to use ,-)

KaptenJansson:
LGTM

KaptenJansson:
Really neat demo! Thanks!

One thought, after clicking Restart ICE there isn't any feedback visually on the page indicating that the restart has been done. Maybe we could add some indicator that iceRestart has been performed?

fippo:
yeah, but for the proper visual indicator (showing the remote ip address like in the constraints sample) I need a different behaviour in the browser :-)

PR #735, opened 2016-01-13
some ice restart tuning

fippo:
The html/css was mostly copied from the constraints page. @samdutton one day we will sit down and clean up everything, ok? ;-)

Biggest technical change is that the restart button is only re-enabled when pc1s iceconnectionstate changes to connected. This is mostly because ice restarts (sometimes?) cause the connectionstate to go from completed to connected immediately. I expected 'checking' here and want to avoid users clicking "ice restart" while one is in progress.

Biggest visual change is that the remote addresses are shown. I'm updating both now on pc1s iceconnectionstatechange to completed which should be ok.

### DIFF
--- a/src/content/peerconnection/restart-ice/css/main.css
+++ b/src/content/peerconnection/restart-ice/css/main.css
@@ -16,13 +16,14 @@ button#hangupButton {
 
 video {
   height: 225px;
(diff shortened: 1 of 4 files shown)


KaptenJansson:
LGTM

PR #736, opened 2016-01-13
Add audio to multiple-relay sample

samdutton:
PR version online at https://samdutton.github.io/webrtc/src/content/peerconnection/multiple-relay.

### DIFF
--- a/src/content/peerconnection/multiple-relay/css/main.css
+++ b/src/content/peerconnection/multiple-relay/css/main.css
@@ -16,13 +16,20 @@ div#buttons {
 }
 
 div#status {
-  height: 1em;
-  margin: 0 0 20px 0;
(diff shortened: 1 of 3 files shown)


tomasgunn:
Great - one more request, can we mute the local video preview?  Just to avoid the unnecessary echo from the local player. Echo from the remote one is fine (and preferred).

samdutton:
> can we mute the local video preview

Added muted attribute to video1.

tomasgunn:
lgtm!

fippo:
can you also add a note in which version this is supported? Maybe only enable the checkbox depending on version but... lets keep things simple.

Awesome @tomasgunn!

samdutton:
Added note in #737.

PR #741, opened 2016-01-25
more useful chrome flags

fippo:
@KaptenJansson I think we should move the selenium lib into a npm package.

### DIFF
--- a/test/selenium-lib.js
+++ b/test/selenium-lib.js
@@ -43,7 +43,10 @@ function buildDriver() {
       .setChromeBinaryPath('node_modules/.bin/start-chrome')
       .addArguments('allow-file-access-from-files')
       .addArguments('use-fake-device-for-media-stream')
-      .addArguments('use-fake-ui-for-media-stream');
+      .addArguments('use-fake-ui-for-media-stream')
+      .addArguments('disable-translate')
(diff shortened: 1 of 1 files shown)


KaptenJansson:
Yes, I will do that this week.

KaptenJansson:
FTR What's the reason behind enabling these flags?

fippo:
mute-audio is useful for local testing unless you like the beeps. The other two seemed like good common practice from http://peter.sh/experiments/chromium-command-line-switches/

KaptenJansson:
I cannot find the no-process-singleton-dialog flag at http://peter.sh/experiments/chromium-command-line-switches/. I googled it and basically it seems like it suppresses a warning if another instance is using the profile?

While we are at good testing practices, maybe we should include --no-first-run --no-default-browser-check as well?

fippo:
hrm... I wonder where I found the flag but peter.sh gave me funny versions in the last days -- some showing the big TODO for tommi, other that did not. But yes, suppressing that warning seemed like a good best practice for testing.

+1 on -no-first-run and -no-default-browser-check, even though those mostly affect some ui bars which are ignored anyway.

Will make a new PR in the utilities repo. Maybe @phoglund knows some more useful flags offhand?

fippo:
see utilities repo

PR #742, opened 2016-01-29
MediaRecorder demo to use old-style gUM

samdutton:
Plus a minor CSS improvement.


fippo:
shouldn't we enable the flag anyway? We don't do in the tests but that is just a reason to migrate to https://github.com/webrtc/utilities ;-)

KaptenJansson:
I'm going to merge this with the error as I cannot see how the latest commit relates to the code and will look into the failure separately.

Regarding the flags, that's a separate discussion ;).

LGTM

PR #750, opened 2016-02-09
Use webrtc-utilities for selenium, update adapter.js to 0.2.10

KaptenJansson:


KaptenJansson:
Turns out we need > 1024 in pixel height in order for the Gather candidates button in the turn/stun server  demo to be clickable now ;).

PR #754, opened 2016-02-14
use the root folder instead of the test one

juandebravo:
This commit https://github.com/webrtc/samples/commit/139da5b0c819b7149eef596e30af5722014ea1dc moved the `server.js` file to the `test` folder. We need to _mount_ the `..` folder to be served via express.

### DIFF
--- a/test/server.js
+++ b/test/server.js
@@ -21,7 +21,7 @@ pem.createCertificate({days: 1, selfSigned: true}, function(err, keys) {
 
   var app = express();
 
-  app.use(express.static('.'));
+  app.use(express.static('../'));
 
   // Create an HTTPS service.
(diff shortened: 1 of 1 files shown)


KaptenJansson:
Well spotted, I've not used it via express hence this did not occur to me. Thanks!

juandebravo:
cool, thanks @KaptenJansson .

BTW, I included _[samples](https://github.com/webrtc/samples)_ in https://github.com/juandebravo/webrtc-box so it's easier to get started with it.

KaptenJansson:
Neat, that's exactly what I did on my Windows machine at home ;). Nice to have a turn key vm ;)

PR #755, opened 2016-02-15
Check for error message in status element for empty file test

KaptenJansson:
Currently the test waits for the Download element to appear for 90 seconds before continuing, it never appears for the empty file test thus causing the empty file test taking way too long.


KaptenJansson:
@fippo PTAL

Issue #756, opened 2016-02-18
Add issue and PR templates


KaptenJansson:
https://github.com/samccone/issue-and-pr-templates

PR #758, opened 2016-02-22
Added ISSUE_TEMPLATE.md

samdutton:
Re  #756.

We can agree the template for /samples here, then create versions for the other WebRTC repos.

### DIFF
--- a/ISSUE_TEMPLATE.md
+++ b/ISSUE_TEMPLATE.md
@@ -1,0 +1,13 @@
+**Description**
+
+
+**Steps to reproduce**
+
+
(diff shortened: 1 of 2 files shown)


KaptenJansson:
Maybe move browser version to the top like we have in crbug.com?

KaptenJansson:
Also, I think we should add a pull request template with Description and either Purpose, Justification or Why headers.

samdutton:
Good idea @KaptenJansson — tweaked and added.

KaptenJansson:
LGTM

PR #759, opened 2016-02-26
Warper

hlundin:
**Description**
Crate an audio "warping" effect.

**Purpose**
Just for fun.

### DIFF
--- a/src/content/peerconnection/pc1/index.html
+++ b/src/content/peerconnection/pc1/index.html
@@ -55,7 +55,9 @@
 
   <script src="../../../js/adapter.js"></script>
   <script src="../../../js/common.js"></script>
+  <script src="js/webaudioextended.js"></script>
   <script src="js/main.js"></script>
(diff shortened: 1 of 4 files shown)


samdutton:
Nice work!

I'd suggest fixing the ESLint errors — they're all trivial, but probably best to keep Travis happy. 

If you fix webaudioextended.js, it'd be worth doing a PR to change the original version as well.

src/content/peerconnection/pc1/index.html
```diff
@@ -55,7 +55,9 @@
 
   <script src="../../../js/adapter.js"></script>
   <script src="../../../js/common.js"></script>
+  <script src="js/webaudioextended.js"></script>
   <script src="js/main.js"></script>
+  <script src="../../../../../dsp.js/dsp.js"></script>
```
juberti, line 60: looks like you'll need to add this?

PR #768, opened 2016-03-28
MediaDeviceInfo:'id' should be 'deviceId'

xcorpio:
**Description**
MediaDeviceInfo:'id' should be 'deviceId'

**Purpose**
MediaDeviceInfo:'id' should be 'deviceId'


xcorpio:
I signed it!

KaptenJansson:
Thank you for your contribution! We already shim this in adapter.js but it's better to use the new property deviceId.

LGTM

PR #769, opened 2016-03-29
Bump adapter.js to 0.2.4

ghost:
**Description**

**Purpose**


KaptenJansson:
This is an old version of adapter.js. I will update soon which requires more work since it's been refactored.

PR #773, opened 2016-04-21
add bandwidth setter example

fippo:
**Description**
adds an example that shows how to use b=AS to control bandwidth, on the fly even.

**Purpose**
so we can say 'there is a sample for that on discuss-webrtc. Now it is showing b=AS instead of TIAS but... soon :-)

cc @taylor-b

### DIFF
--- a/src/content/peerconnection/bandwidth/css/main.css
+++ b/src/content/peerconnection/bandwidth/css/main.css
@@ -1,0 +1,45 @@
+/*
+ *  Copyright (c) 2016 The WebRTC project authors. All Rights Reserved.
+ *
+ *  Use of this source code is governed by a BSD-style license
+ *  that can be found in the LICENSE file in the root of the source
+ *  tree.
+ */
+video {
+  display: inline-block;
+  position: relative;
+  top: 9px;
+  width: calc(100% - 120px);
+}
+
+button {
+  margin: 0 20px 0 0;
+  width: 96px;
+}
+
+video {
+  height: 225px;
+  margin: 0 0 20px 0;
+  vertical-align: top;
+  width: calc(50% - 12px);
+}
+
+video#localVideo {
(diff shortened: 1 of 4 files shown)


src/content/peerconnection/bandwidth/css/main.css
```diff
@@ -0,0 +1,45 @@
+/*
+ *  Copyright (c) 2015 The WebRTC project authors. All Rights Reserved.
```
KaptenJansson: Update copyright throughout.

src/content/peerconnection/bandwidth/js/main.js
```diff
@@ -0,0 +1,207 @@
...
+      pc2.createAnswer(gotDescription2, onCreateSessionDescriptionError);
+    }, onSetSessionDescriptionError);
+  }, onSetSessionDescriptionError);
+}
+
+function gotDescription2(desc) {
+  pc2.setLocalDescription(desc, function() {
+    trace('Answer from pc2 \n' + desc.sdp);
+    // insert b=AS after c= line.
+    desc.sdp = desc.sdp.replace(/c=IN IP4 (.*)\r\n/,
```
KaptenJansson, line 100: Hmm this made me wonder, is this set to IP4 when on a IP6 network?

fippo, line 100: good question. let me test... or just go for c=IN :-)

fippo, line 100: doesn't look like. I think because of interop concerns, there was this bug @phoglund found a while back where Firefox did not like ipv6 c=

KaptenJansson, line 100: ah yes you right, the change was reverted for that reason.

KaptenJansson:
LGTM

PR #783, opened 2016-05-07
munge the correct way

fippo:
since I heard rumors that munging between createOffer and SLD will be prohibited... it is not even necessary to do it that way.

(I did some evil things in #782)


KaptenJansson:
LGTM.

PR #791, opened 2016-05-10
re-add rtcp mux policy

### LINKED ISSUE #690: Re-enable "Gather unbundled RTCP candidates" checkbox logic in Trickle ICE sample page
https://webrtc.github.io/samples/src/content/peerconnection/trickle-ice/ includes a "Gather unbundled RTCP candidates" checkbox. However, the state of the checkbox doesn't change the output of the gathered candidates at all.

On further inspection, it looks this is because the js logic related to the checkbox has been commented out. Here are what I believe are related snippets from src/content/peerconnection/trickle-ice/js/main.js:

`// var unbundleCheck = document.querySelector('input#unbundle');`
...
`// offerOptions.optional = [{'googUseRtpMUX': !unbundleCheck.checked}];`

There's no TODO near either of those commented out lines. Does anyone recall why they were commented out, and if that reason is no longer valid, can they be reinstated?

fippo:
fixes #690 -- rtcpmux policy made it into the spec.


KaptenJansson:
LGTM

PR #796, opened 2016-05-26
Resolving |Failed to load resource: net::ERR_FILE_NOT_FOUND" error on Chrome

vikasTmz:
**Description**
 Added http for testing on local file system. 
chrome adds file:// if no protocol is specified.

**Purpose**
Doubt non-developers will be able to figure where the error is by checking their console.log(), if they are testing on their local system.

### DIFF
--- a/src/content/devices/input-output/index.html
+++ b/src/content/devices/input-output/index.html
@@ -23,7 +23,7 @@
   <title>Select audio and video sources</title>
 
   <link rel="icon" sizes="192x192" href="../../../images/webrtc-icon-192x192.png">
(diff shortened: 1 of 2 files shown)


samdutton:
This makes sense if you want to run from a file:// URL — but these samples should be run from a local web server, in which case the protocol-relative URL works OK. 

Having said that, we should probably specify HTTPS for both these.

@KaptenJansson WDYT?

KaptenJansson:
Protocol relative URLs are fine by me. I do not see the need to change them to HTTPS for the reason you mentioned.

samdutton:
I think I'll close this, for the reasons given above.

PR #803, opened 2016-06-03
Modified "constraints" demo to use getSettings for video stats.

alvestrand:
Tested on Chromium tip-of-tree, Chrome 50.0.2661 and Firefox.

When getSettings is available, frame rate shows up.
When getSettings returns width and height, the "x" will be in uppper case.
(Not implemented in Chrome yet)


src/content/peerconnection/constraints/js/main.js
```diff
@@ -202,6 +203,43 @@ function onAddIceCandidateError(error) {
...
+  if (video.videoWidth) {
+    sizeFromTag = video.videoWidth + 'x' + video.videoHeight;
+  }
+  // Use track's info if available.
+  // NOTE: getSettings is under development, so code adapts to various
+  // in-flight situations that aren't according to spec.
+  if (stream) {
+    if (stream.getVideoTracks()[0]) {
+      if (stream.getVideoTracks()[0].getSettings) {
+        var settings = stream.getVideoTracks()[0].getSettings();
```
samdutton, line 223: Might be worth doing something like `var settings = window.settings = ...` so that `settings` is visible from the console.

samdutton:
Code LGTM. WFM (though without `getSettings`) in 53.0.2760.0 on Mac.

src/content/peerconnection/constraints/js/main.js
KaptenJansson, line 223: WDYT @alvestrand ?

alvestrand, line 223: I can do that - I've found it useful in other cases. The global object should probably be videoSettings. (apologies for being slow)

KaptenJansson:
Ping ;)

alvestrand:
Thanks for the ping - didn't get to it before I went on vacation. Will return to it, but not this week.

KaptenJansson:
Ping

KaptenJansson:
@alvestrand do you plan to update this?

KaptenJansson:
Closing due to age but will leave the branch.

PR #806, opened 2016-06-16
allow ICE restarts from connected state

nils-ohlmeier:
**Description**
Allow the ICE Restart button to be pressed from the ICE 'connected' state instead of 'completed'.

**Purpose**
This allows to demo ICE restart with Firefox, and also works with Google Chrome.

### DIFF
--- a/src/content/peerconnection/restart-ice/js/main.js
+++ b/src/content/peerconnection/restart-ice/js/main.js
@@ -121,7 +121,7 @@ function call() {
   };
   pc1.oniceconnectionstatechange = function(e) {
     onIceStateChange(pc1, e);
-    if (pc1 && pc1.iceConnectionState === 'completed') {
+    if (pc1 && pc1.iceConnectionState === 'connected') {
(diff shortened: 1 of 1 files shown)


KaptenJansson:
I thought one side on Chrome is set to completed and the other connected, cannot remember which side does what however :(. 

Which versions of Chrome did you test this on?

nils-ohlmeier:
I thought Chrome goes from connected to completed when it has tried all candidates. That at least would match the spec.
I tested it with Chrome 51.

But in general I think it should also not matter what the final ICE connection state is. The question is more from which state on can you do an ICE restart. And I'm pretty sure connected should be good enough for attempting an ICE restart.

fippo:
it is. I wanted to avoid users spamming the button but this is just a sample so... \o/

KaptenJansson:
Aha that might be true just that the time for trying all the candidates differs greatly on the sender and receiver side (at least when I was testing stuff on AppRTC on the same machine).

Then LGTM.

PR #816, opened 2016-08-10
Fix broken regex replacement in bandwidth sample

kolmodin:
**Description**
The syntax for referring to the first matched group is $1, not $(1).
The previous incorrect use added the string $(1) to the SDP.

**Purpose**

Fix bug.


fippo:
hrm... i have seen the issue earlier and wonder why it isn't already fixed. But thanks for catching it!

kolmodin:
I'll have to redo this with my `@google.com` email.

kolmodin:
Please look at #817 instead.

PR #817, opened 2016-08-10
Fix broken regex replacement in bandwidth sample

kolmodin:
**Description**
The syntax for referring to the first matched group is $1, not $(1).
The previous incorrect use added the string $(1) to the SDP.

**Purpose**
Fix bug.


KaptenJansson:
LGTM

PR #819, opened 2016-08-10
Update trace() to handle objects

samdutton:
**Description**
- Enable `trace()` to handle objects.
- Remove redundant shim code, since `window.performance.now()` is now widely supported.
- Remove code to remove line break at end of string. Line breaks work OK with `console.log()`, but we could add this back if it causes too much white space in logging.

**Purpose**
Enable richer logging and avoid having to stringify objects before using `trace()`.

@KaptenJansson PTAL.


KaptenJansson:
Ping @alvestrand

src/js/common.js
```diff
@@ -1,12 +1,7 @@
...
-    console.log(now + ': ' + text);
-  } else {
-    console.log(text);
-  }
+ /* exported trace */
+
+// Logging utility function.
+function trace(arg) {
+  var now = (window.performance.now() / 1000).toFixed(3);
+  console.log(now + ': ', arg);
```
alvestrand, line 6: Will this print [object Object] for objects that don't have a stringifier?
This can be "caller's responsibility". It would be nice to have this implicit contract noted in the function comment.
I also note that we've dropped responsibility for removing trailing NL. That's good.

src/js/common.js
```diff
@@ -1,12 +1,7 @@
-function trace(text) {
-  // This function is used for logging.
-  if (text[text.length - 1] === '\n') {
-    text = text.substring(0, text.length - 1);
-  }
-  if (window.performance) {
-    var now = (window.performance.now() / 1000).toFixed(3);
```
alvestrand, line 7: This means that we crash on Chrome versions before performance.now being added.
Does that affect "min supported version" for the demos? Where is that documented?
I know we have a check in adapter.js on this, but I'm not sure we ever do anything with it.

alvestrand:
I think this is OK, but it doesn't do what it says in the description any more. I guess the rest got taken care of with the other merge.

Please rebase.

KaptenJansson:
@samdutton ping

src/js/common.js
samdutton: Are we likely to encounter objects without a stringifier in the this context? `console.log(fu, bar)` seems to work with everything I've thrown at it :).

> removing trailing NL

Yes :). I removed this, because it adds complexity and removing it doesn't seem to cause much trouble.

src/js/common.js
samdutton: `performance.now` is supported by Chrome 25+ on Android; Chrome 20+, IE10+ and Firefox 15+ on desktop, so I don't think we need worry.

samdutton:
I'll merge this, but we should keep an eye out for anomalies.

PR #828, opened 2016-09-12
Fix secure origin check

saghul:
Fixes serving these samples from localhost, in a non-default port.  In this case, location.host can be "localhost:8000" and location.hostname will be "localhost".  Check for the latter when deciding if an origin is secure, before redirecting to HTTPS.

### DIFF
--- a/index.html
+++ b/index.html
@@ -176,7 +176,7 @@
 
   <script type="application/javascript">
   var isSecureOrigin = location.protocol === 'https:' ||
-      location.host === 'localhost';
+      location.hostname === 'localhost';
   if (!isSecureOrigin) {
     console.log('These demos must be run from a secure origin. \n' +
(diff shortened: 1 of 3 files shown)


samdutton:
Good catch! Thank you.

Could you do me a huge favour and correct this in the following two examples as well, so we can incorporate them in the same PR:
- capture/canvas-record/js/main.js
- getusermedia/record/js/main.js

We don't need these checks on this site any more, since we're forcing HTTPS, but it's probably worth keeping for others who re-use the code elsewhere.

saghul:
@samdutton sure, will do!

saghul:
Done!

samdutton:
Nice one! Appreciated.

saghul:
Excellent.  Thanks for the prompt response @samdutton!

PR #829, opened 2016-09-15
constraints sample: better way to access remote candidate object

fippo:
**Description**
the current code is rather clumsy. Why search all objects for one that matches id if you can just access it directly.

cc @jan-ivar

### DIFF
--- a/src/content/peerconnection/constraints/js/main.js
+++ b/src/content/peerconnection/constraints/js/main.js
@@ -249,13 +249,7 @@ setInterval(function() {
         }
       });
       if (activeCandidatePair && activeCandidatePair.remoteCandidateId) {
-        Object.keys(results).forEach(function(result) {
-          var report = results[result];
-          if (report.type === 'remotecandidate' &&
-              report.id === activeCandidatePair.remoteCandidateId) {
-            remoteCandidate = report;
-          }
-        });
+        remoteCandidate = results[activeCandidatePair.remoteCandidateId];
       }
       if (remoteCandidate && remoteCandidate.ipAddress &&
           remoteCandidate.portNumber) {


KaptenJansson:
Maybe we couldn't before due to differences in getStats?

KaptenJansson (approved):
LGTM

fippo:
very legacy. We couldn't do it as easy in the old chrome stats model

src/content/peerconnection/constraints/js/main.js
```diff
@@ -249,13 +249,7 @@ setInterval(function() {
...
       });
       if (activeCandidatePair && activeCandidatePair.remoteCandidateId) {
-        Object.keys(results).forEach(function(result) {
-          var report = results[result];
-          if (report.type === 'remotecandidate' &&
-              report.id === activeCandidatePair.remoteCandidateId) {
-            remoteCandidate = report;
-          }
-        });
+        remoteCandidate = results[activeCandidatePair.remoteCandidateId];
```
jan-ivar, line 252: Spec getStats is maplike now, so this needs to be:

```
remoteCandidate = results.get(activeCandidatePair.remoteCandidateId));
```

jan-ivar (commented):
Yay, but it should be maplike.

fippo:
small changes, I am not ready for maplike ;-)

jan-ivar:
@fippo Why not? Maplike was polyfilled back in https://github.com/webrtc/adapter/pull/277.

https://jsfiddle.net/sc3dh9wr/ appears to work in Chrome too.

Here's my maplike cheatsheet if it helps inspire: https://jsfiddle.net/4ea0nj0s/

fippo:
@jan-ivar i'd like to have codemod for that

PR #831, opened 2016-09-27
munge-sdp: fix onicecandidate handler for Edge

fippo:
the onicecandidate handler somehow ended up inside the check for
createDataChannel so it was not hooked up in edge which lead to ice failures.

Many thanks to @shijuns for reporting this!


samdutton:
LGTM (and change works OK) though I'm not sure why this was in the createDataChannel check. 

I also don't have Edge to hand.

alvestrand:
lgtm
I assume this means that all other browsers would initialize the data channel...

fippo:
@alvestrand any browser which can create a datachannel will do that, yes

Issue #827, opened 2016-09-09
Outdated captureStream demo


wpp:
**Browsers and versions affected**

Chrome Version 53.0.2785.101 (64-bit)

**Description**

The [captureStream demo](https://webrtc.github.io/samples/src/content/capture/video-video/) logs `captureStream() not supported` in Chrome 53, despite the page noting:

> This demo requires Firefox 47, Chrome 53 (or later).

**Steps to reproduce**

Visit https://webrtc.github.io/samples/src/content/capture/video-video/ with Chrome 53.

**Expected results**

Either update the required browser version or a working captureStream demo.

**Actual results**

Page lists Chrome 53 as a required browser but logs `captureStream() not supported` console and the demo is not working.

KaptenJansson:
@samdutton PTAL

IvRRimum:
Any update on this?

IvRRimum:
Has chrome removed support for captureStream in chrome 54?

samdutton:
Thanks for reminding me to sort this out.

I've added a note about enabling Experimental Web Platform Features in chrome://flags.

wpp:
Hi!

Good to know! Unfortunately there is still an issue even after enabling the flag. The demo only seems to work as intended when I drag the player slider on the left.

wpp:
Just a quick update after enabling the flag this demo works now. I'll close this issue, hope everyone is ok with that.

jhonny-villarroel:
Sorry, for the question:

In the future will there be any version that supports "chrome://flags/#enable-experimental-web-platform-features" by default, instead of enable manually ?

Thanks so much in advance.

fippo:
@yellowdoge do you know (or know someone who knows)?

jhonny-villarroel:
@samdutton, any comment about that ?

In the future will there be any version that supports "chrome://flags/#enable-experimental-web-platform-features" by default, instead of enable manually ?

Because currently I am using Chrome Versión 59.0.3071.104 (Build oficial) (64 bits) and the Flag is not supported by default.


Thanks in advance

fippo:
there is an intent to ship on [blink-dev](https://groups.google.com/a/chromium.org/forum/?utm_medium=email&utm_source=footer#!msg/blink-dev/6Vai94SPoEQ/DLmSiLH3BAAJ) which means this could ship in 61.

go @yellowdoge ;-)

cameronelliott:
[...]

PR #840, opened 2016-11-09
Add  chrome://flags note

samdutton:
In response to #827.

### DIFF
--- a/src/content/capture/video-video/index.html
+++ b/src/content/capture/video-video/index.html
@@ -44,7 +44,7 @@
 
     <video id="rightVideo" autoplay controls></video>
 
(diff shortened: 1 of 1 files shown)


KaptenJansson:
Seems like Chrome beta has changed the name of the fake audio devices. It's now `Fake Default Audio Input`. 

This is not pretty but maybe we need to add another if or switch case for Chrome beta (version) in com/webrtc/samples/blob/gh-pages/src/content/devices/input-output/js/test.js#L82 ?

Adapter provides the browser version via browserDetails, see com/webrtc/samples/blob/gh-pages/src/content/devices/input-output/js/test.js#L82.

samdutton:
@KaptenJansson OK to merge?

KaptenJansson:
LGTM

Issue #710, opened 2015-12-11
Firefox Media Stream recording confusing error


MattyBalaam:
If the camera is already being used by a different browser when recording is attempted an unrelated error alert pops up.

https://webrtc.github.io/samples/src/content/getusermedia/record/

![firefox-record-fail](https://cloud.githubusercontent.com/assets/1246923/11742824/b013b262-9ffa-11e5-950d-ffd007cda307.png)

samdutton:
I don't get this error — @KaptenJansson, can you reproduce?

MattyBalaam:
Can I check you noticed I edited the issue? have now noticed this on Chrome too, so is not browser specific.

To reproduce:
1. Start on one browser, allow camera access, click record and leave running.
2. Start on other browser, allow camera access, click record: alert appears.

fippo:
@MattyBalaam most likely getUserMedia fails since only one process/browser can access the camera device at a time.

KaptenJansson:
Working as intended, however that error alert should indicate that getUserMedia failed (in the scenario described in MattBalaam's comment), not that MediaRecoder is not supported.

samdutton:
I can't seem to reproduce this :/.

I've tried opening the MediaRecorder demo (and starting recording) on Chrome Stable, then Firefox — or Chrome Stable then Chrome Canary, or Canary then Firefox. No problems.

I can also run multiple simple getUserMedia() demos.

Am I missing something?

I'm on Mac (using the internal camera). @MattyBalaam what OS are you on?

MattyBalaam:
I encountered the error on a windows 7 machine, with a usb camera.

I have just tired on a Macbook Pro running El Capitan, and accessing the internal camera I can record using both browsers at once.

samdutton:
Thanks for testing this out.

I don't have a Windows machine to hand — will have a look when I get home. 

I take it you don't actually get a stream (from the second browser) when the page opens? 

We could add a check in `startRecording()` to see if `window.stream` is defined.

MattyBalaam:
The Windows machine is at work, but from memory there is no picture on the second browser.

MattyBalaam:
I'm back at work this morning. Looking at the console, on initial attempt to attach the video source from the browser pop-up the error is logged:

`navigator.getUserMedia error:  MediaStreamError { name: "SourceUnavailableError", message: "Failed to allocate videosource", constraint: "", stack: "" }`

samdutton:
[...]

PR #842, opened 2016-11-10
Re #710: Start Recording button disabled until gUM success

samdutton:
**Description**
Add disabled attribute to **Start Recording** button. Enable button only if `getUserMedia()` succeeds.

**Purpose**
Fix problem described in #710.


KaptenJansson:
Does this mean that you have to reload the page if you want to record again?

samdutton:
> Does this mean that you have to reload the page if you want to record again?

No, it shouldn't — once you get a MediaStream, the Record button is enabled (and not disabled again).

PR #845, opened 2016-11-25
device selection: remove flag note

fippo:
**Description**
removes the flag note since a long time has passed since Chrome 45.

**Purpose**
 This seems to [confuse people](http://stackoverflow.com/questions/40791581/webrtc-getusermedia-works-only-on-firefox/40793349?noredirect=1#comment68824647_40793349)


samdutton:
LGTM.

PR #847, opened 2016-11-28
Some cleanup of the recorder demo

aullman:
**Description**

Some cleanup of the recorder demo

**Purpose**

- Removed comment about using old getUserMedia because we're not. I guess this is an old comment.
- Removed duplicate 'use-strict' and globals comment.


aullman:
I signed it!

KaptenJansson:
Thanks for the cleanup. LGTM!

PR #849, opened 2016-12-07
use adapter-latest from github

fippo:
this replaces the local copy of adapter.js with pulling the latest and greatest version from
https://webrtc.github.io/adapter/adapter-latest.js


fippo:
I suspect firefox tests fail because they're not using marionette yet. Time for an update of the utils :-/

samdutton:
LGTM.

Nice one!

fippo:
I spent too much time hunting down an Edge issue because adapter was _very_ outdated ;-)

KaptenJansson:
LGTM

alvestrand:
Hm. We already have webrtc-adapter in npm dependencies (package.json).
What are the pros & cons of switching that to adapter (we've renamed it) and pulling the local copy that way compared to including adapter-latest directly from the repo?

Agree that one of the two needs to be done, not sure of the pros & cons of each.
(If it was someone else's adapter, I'd immediately want to go the npm route to isolate us from uncoordinated changes. But it's not.)

KaptenJansson:
Put it this way:

Since we are serving on gh-pages, we need to run npm update locally and push to gh-pages EVERYTIME we want to update adapter.js. This is nice because it's controlled and succinct but will result in us updating less often.

If we point to adapter-latest directly, we only need to do commits IF adapter is broken, but that also means in case of a breakage we have to change adapter-latest to adapter-<previous-version> in all the files. This is nice because there is in general less commits (assuming we do not screw up often), adapter gets more test coverage, samples uses the latest and we discover browser related issue early.

Since we control both repos I do not see the problem with the adapter-latest approach.

alvestrand:
OK, let's try it this way. We know how to back out.

KaptenJansson:
Also, this is another reason why we should add a cron job that runs the tests daily, not just on commits.

PR #851, opened 2016-12-08
use maplike instead of Object.keys in getStats

fippo:
this uses the maplike syntax instead of the older Object.keys() iteration. Maplike is shimmed by adapter for Chrome.

Code is similar to http://w3c.github.io/webrtc-pc/#example

### DIFF
--- a/src/content/peerconnection/constraints/js/main.js
+++ b/src/content/peerconnection/constraints/js/main.js
@@ -291,13 +291,13 @@ setInterval(function() {
 function dumpStats(results) {
   var statsString = '';
   results.forEach(function(res) {
-    statsString += '<h3>Report ';
+    statsString += '<h3>Report type=';
     statsString += res.type;
     statsString += '</h3>\n';
+    statsString += 'id ' + res.id + '<br>\n';
     statsString += 'time ' + res.timestamp + '<br>\n';
-    statsString += 'type ' + res.type + '<br>\n';
     Object.keys(res).forEach(function(k) {
-      if (k !== 'timestamp' && k !== 'type') {
+      if (k !== 'timestamp' && k !== 'type' && k !== 'id') {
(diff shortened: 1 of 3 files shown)


src/content/peerconnection/constraints/js/main.js
```diff
@@ -290,10 +290,9 @@ setInterval(function() {
 // might be named toString?
 function dumpStats(results) {
   var statsString = '';
-  Object.keys(results).forEach(function(key, index) {
-    var res = results[key];
+  results.forEach(function(res) {
     statsString += '<h3>Report ';
-    statsString += index;
+    statsString += res.type;
```
alvestrand, line 295: Is the type field really the same as the index from the previous iteration?
(just reassure me....)

fippo: hah. I wanted to point out that change. I don't think it makes sense to have a numeric index here. I tried the report id but that got very long and ugly

fippo:
@jan-ivar can you please take a look whether this is what we want to show?

jan-ivar:
Lgtm.

src/content/peerconnection/constraints/js/main.js
alvestrand: @KaptenJansson as samples-maintainer, does this make sense for you?

KaptenJansson: I find it a little confusing that we are adding the type to statsString twice. Maybe add a comment to why?

fippo: changed the output format slightly in 486ee457d30f374247a77f37b4cf6234cf7dde55 -- now type, timestamp and id are special

KaptenJansson (approved):
LGTM

PR #853, opened 2016-12-08
Set up video tag capture after oncanplay events.

pbos:
Previously tag capture was set up on the onplay event, when the video
tag started playing, but that happens on page load which can be before
the video file is loaded and video tracks enumerated.

### DIFF
--- a/src/content/capture/video-pc/js/main.js
+++ b/src/content/capture/video-pc/js/main.js
@@ -22,7 +22,10 @@ var offerOptions = {
 
 var startTime;
 
-leftVideo.onplay = function() {
+function maybeCreateStream() {
+  if (stream) {
+    return;
+  }
   if (leftVideo.captureStream) {
     stream = leftVideo.captureStream();
     console.log('Captured stream from leftVideo with captureStream',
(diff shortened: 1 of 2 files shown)


src/content/capture/video-pc/js/main.js
```diff
@@ -22,7 +22,10 @@ var offerOptions = {
 
 var startTime;
 
-leftVideo.onplay = function() {
+function createStream() {
```
KaptenJansson: Change to maybeCreateStream() since it does not always create the stream?

src/content/capture/video-video/js/main.js
```diff
@@ -13,7 +13,10 @@ var rightVideo = document.getElementById('rightVideo');
 
 var stream;
 
-leftVideo.onplay = function() {
+function createStream() {
```
KaptenJansson: Change to maybeCreateStream() since it does not always create the stream?

pbos: Done!

src/content/capture/video-pc/js/main.js
pbos: Done!

pbos:
The travis build looks flaky (fails beta, passes stable, unrelated test). I think this might be good to go. PTAL.

fippo:
it has passed before... I think its safe to blame selenium 3 for being flaky :-/

PR #855, opened 2016-12-09
all: remove RTCIceCandidate

fippo:
RTCIceCandidate is no longer required (when using adapter), this removes
it. Also cleans up some places where callback style was used.

(the next step in my evil plan is to remove the event.candidate existence checks and just call addIceCandidate(undefined) ;-))


KaptenJansson:
LGTM

PR #856, opened 2016-12-09
test: re-enable firefox testing

fippo:
by updating utilities and changing travіs

PR #857, opened 2016-12-09
bump travis to node6

fippo:
this should avoid test failures as seen in #853 and #856


fippo:
sporadic failure caused by travis not being able to download and install node. Fun!

PR #858, opened 2016-12-20
Add demo page for MediaStreamTrack content hints.

pbos:
Makes use of https://wicg.github.io/mst-content-hint/ implemented in tip
of tree Chromium/Chrome Canary.

The demo page spawns two PeerConnections that transfer video locally at
50kbps, significantly under the required bitrates for good quality. It
then makes use of MediaStreamTrack content hints to say whether to
prefer fluid motion over individual frame details or not.


pbos:
See https://crbug.com/653531 for more info on the tip of tree Chromium implementation.

fippo:
can you use promise-style for all the peerconnection APIs? callback-style is so legacy ;-)

pbos:
Is there any demo/example of this done with promises? I'm not normally a JavaScript dev. :)

I can also do this as a follow-up if this is merge-able as is. We'd like to get it out timely so we can start pushing spec interest.

fippo:
let me run https://github.com/fippo/webrtc-codemods once this is merged ;-)

pbos:
Done, per your suggested diff. :)

fippo:
the work of mindless machines ;-)

pbos:
@alvestrand: FWIW: Failure above looks like an unrelated flake.

src/content/capture/video-contenthint/index.html
```diff
@@ -0,0 +1,76 @@
...
+    <div id="videos">
+      <div class='video-container'>
+      <h2><span>Source video file (high bitrate)</span></h2>
+        <video id="srcVideo" controls muted loop>
+          <source src="../../../video/mixed-content.webm" type="video/webm" />
+          <p>This browser does not support the video element.</p>
+        </video>
+        </div>
+      <div class='video-container'>
+        <h2><span>"Fluid" video @ 50kbps</span></h2>
```
alvestrand: I don't see a reason for the &lt;span> inside the &lt;h2>...

src/content/capture/video-contenthint/js/main.js
```diff
@@ -0,0 +1,151 @@
...
+    onSetSessionDescriptionError
+  );
+  // Since the 'remote' side has no media stream we need
+  // to pass in the right constraints in order for it to
+  // accept the incoming offer of audio and video.
+  pc2.createAnswer().then(
+    function(answerDesc) {
+      onCreateAnswerSuccess(pc1, pc2, answerDesc);
+    },
+    onCreateSessionDescriptionError
```
alvestrand: Style preference: I prefer .catch(errorhandler) instead of the 2-argument .then, since that will cause errors inside the first .then handler to be caught too.

src/content/capture/video-contenthint/js/main.js
```diff
@@ -0,0 +1,151 @@
...
+  );
+}
+
+function onSetSessionDescriptionError(error) {
+  trace('Failed to set session description: ' + error.toString());
+}
+
+function gotRemoteStream(video, event) {
```
alvestrand: With promises, there's usually little need for these one-line, use-once functions. Just write the code inline. This particular one can be inlined despite it not being a promise.

src/content/capture/video-contenthint/js/main.js
```diff
@@ -0,0 +1,151 @@
...
+  video.srcObject = event.stream;
+}
+
+function onCreateAnswerSuccess(pc1, pc2, desc) {
```
alvestrand, line 103: Once you've inlined everything that's only declared once, you end up with a .then-chain that you can end with a .catch(function(e) { log "something went wrong" } ) - for a demo, that's sufficient error handling.

src/content/capture/video-contenthint/js/main.js
```diff
@@ -0,0 +1,151 @@
...
+    function() {},
+    onSetSessionDescriptionError
+  );
+  pc1.setRemoteDescription(desc).then(
+    function() {},
+    onSetSessionDescriptionError
+  );
+}
+
+function onIceCandidate(pc, otherPc, event) {
```
alvestrand, line 112: This is probably the one function that you can't inline.....

src/content/capture/video-contenthint/js/main.js
```diff
@@ -0,0 +1,151 @@
...
+function maybeCreateStream() {
+  if (srcStream) {
+    return;
+  }
+  if (srcVideo.captureStream) {
+    srcStream = srcVideo.captureStream();
+    console.log('Captured stream from srcVideo with captureStream',
+      srcStream);
+    call();
+  } else if (srcVideo.mozCaptureStream) {
```
alvestrand: For this particular demo, perhaps we don't care about legacy Mozila compatibility. Just handle what's in the standard.

src/content/capture/video-contenthint/js/main.js
```diff
@@ -0,0 +1,151 @@
...
+    } else {
+      trace('MediaStreamTrack contentHint attribute not supported');
+    }
+  });
+}
+
+function call() {
+  fluidStream = srcStream.clone();
+  setVideoTrackContentHints(fluidStream, 'fluid');
+  establishPC(fluidVideo, fluidStream);
```
alvestrand, line 66: This sequence creates 4 RTCPeerConnections. You might want to put in a comment that the two need to be independent so that they both get the same independent limit (50k).

alvestrand (commented):
I think this demo is absolutely beautiful. The code can use another pass to be more in the "promise style".

Should be ready to go with another pass!

src/content/capture/video-contenthint/index.html
pbos: Done!

src/content/capture/video-contenthint/js/main.js
pbos: Done!

src/content/capture/video-contenthint/js/main.js
pbos: Comment done in call().

src/content/capture/video-contenthint/js/main.js
pbos: Me too, this got prettier, PTAL.

src/content/capture/video-contenthint/js/main.js
pbos: Done!

src/content/capture/video-contenthint/js/main.js
pbos: Pretty much done, take another look and see if you agree. There's a lot of catches to onSetSessionDescriptionError.

src/content/capture/video-contenthint/js/main.js
pbos: I believe so too. :)

src/content/capture/video-contenthint/js/main.js
fippo: there is a semantic difference here (as @jan-ivar pointed out to me a while ago):
With
.then(something)
.catch(errhandler)
errhandler catches errors in something as well which may be unintended

alvestrand: Yep, which is why I think .catch is mostly useful for "there's been an error, I give up" types of error handling - however that's frequently exactly what you want, especially in error cases in short demos. Jan-Ivar and I have discussed this before...

src/content/capture/video-contenthint/js/main.js
```diff
@@ -0,0 +1,120 @@
...
+
+  pc1.addStream(stream);
+  pc1.createOffer(offerOptions).then(
+    function(desc) {
+      pc1.setLocalDescription(desc).catch(onSetSessionDescriptionError);
+      pc2.setRemoteDescription(desc).catch(onSetSessionDescriptionError);
+      // Since the 'remote' side has no media stream we need
+      // to pass in the right constraints in order for it to
+      // accept the incoming offer of audio and video.
+      pc2.createAnswer().then(
```
alvestrand: You ought to chain createAnswer() off setRemoteDescription, like this:

pc2.setRemoteDescription(desc).then(function(){pc2.createAnswer()}).then(function(answerDesc)....

(suitably line-broken and indented).
A then-function should usually return a promise, which can then be passed to the next .then-function.

alvestrand:
Apart from the issue of properly sequencing createAnswer() after setRemoteDescription() finishes, I think this is ready to merge.

src/content/capture/video-contenthint/js/main.js
pbos: Done, can you double check with a Canary before merging or I'll have to get into the office first (no CrOS Canary).

PR #862, opened 2017-01-12
Use new ('motion'/'detail') contenthint strings.

pbos:
Old strings exist set as a fallback to not break existing canary. Once
all browser versions (dev, canary) are on the new strings the old ones
can be removed.

Strings were updated as part of
https://github.com/WICG/mst-content-hint/issues/20.

**Description**

**Purpose**


pbos:
PTAL @alvestrand :)

I want this in when submitting https://codereview.chromium.org/2623263003/.

alvestrand (approved):

PR #864, opened 2017-01-19
fix getStats calls

### LINKED ISSUE #863: Ice Restart Local And Remote Video Always Shows "Not connected"
Chrome version : 57.0.2970.0 / dev
Chrome Device: Ubuntu 14.04, Windows 10

Steps to reproduce:
Precondition: Have the JS console open 

1. Go to https://webrtc.github.io/samples/src/content/peerconnection/restart-ice/
2. Confirm that both the local (PC1) and remote windows (PC2) do not show any video, and the message below says "Not connected" 
3. Click "Start" and wait till local video PC1 (in left window) appears
4. Click "Call" 
- Confirm that the remote video PC2 appears
5. Note both the addresses of PC1 and PC2, as seen below the respective windows
6. Click "Restart ICE"
7. Now, note the addresses of PC1 and PC2
8. Press "Hang up"

Expected results:
1. No JS console errors should be seen
2. The last 5 digits in the address of PC1 from step 5, before the ICE restart, should be different from the address of PC1 from step 7 
3. PC2 address should remain the same before/after ICE restart
4. After Hang up, only PC1 has video

Actual results:
Local and remote videos. i.e PC1 & PC2, always show "Not connected" after Step 4 and subsequent steps

Discussing with jansson@ he thinks, maybe, the newer Adapter 3.0 release could have caused this. 

[restart_ice_not_conencted_1.1.33.txt](https://github.com/webrtc/samples/files/716260/restart_ice_not_conencted_1.1.33.txt)

fippo:
this fixes primarily the ice restart demo which was broken in the adapter 3.0
release. Some tests which rely on getStats are still broken, this needs to
be fixed in https://github.com/webrtc/utilities

Fixes #863. The issue with the test is that JSON.stringify(map) is {} and we use that for communication between webdriver and node. See https://github.com/webrtc/utilities/pull/21 for a fix to that

### DIFF
--- a/src/content/peerconnection/audio/js/test.js
+++ b/src/content/peerconnection/audio/js/test.js
@@ -51,8 +51,7 @@ test('Audio-only sample codec preference', function(t) {
     })
     .then(function(stats) {
       // Find the sending audio track.
-      Object.keys(stats).forEach(function(name) {
-        var report = stats[name];
+      stats.forEach(function(report) {
         if (report.type === 'ssrc' && report.googTrackId === trackId) {
           t.ok(codecName === report.googCodecName, 'preferring ' + codecName);
         }
--- a/package.json
+++ b/package.json
@@ -36,6 +36,6 @@
     "grunt-htmlhint": ">=0.9.12",
     "pem": "^1.8.1",
     "webrtc-adapter": "^1.1.0",
-    "webrtc-utilities": ">=0.0.1"
+    "webrtc-utilities": "^1.0.0"
   }
 }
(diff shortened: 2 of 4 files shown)


fippo:
might require a version bump of utils (which should be major since this is a breaking API change)

src/content/peerconnection/audio/js/test.js
```diff
@@ -51,8 +51,8 @@ test('Audio-only sample codec preference', function(t) {
     })
     .then(function(stats) {
       // Find the sending audio track.
-      Object.keys(stats).forEach(function(name) {
-        var report = stats[name];
+      console.log('STATS', JSON.stringify(stats));
```
KaptenJansson: Do we want logging left?

KaptenJansson (commented):
Restarted the tests using the latest webrtc-utilities 1.0 version and now the getStats error is gone but the restart Ice test still fails.

fippo:
yeah... I wonder if we should skip that test. Its artificial and not really test-able :-/

PR #866, opened 2017-01-20
all: remove e.candidate check before addIceCandidate

fippo:
removes the
  if (e.candidate) otherPc.addIceCandidate(e.candidate)
check from all demos. addIceCandidate will act the right way
to e.candidate not being set.

Without this Edge broke in adapter 3.0 which removed the
  type endOfCandiates
hack.

Also unifies the onicecandidate handler style as much as possible.

@KaptenJansson extra nitpicky-mode please ;-)

### DIFF
--- a/src/content/capture/canvas-pc/js/main.js
+++ b/src/content/capture/canvas-pc/js/main.js
@@ -136,18 +136,17 @@ function onCreateAnswerSuccess(desc) {
 }
 
 function onIceCandidate(pc, event) {
-  if (event.candidate) {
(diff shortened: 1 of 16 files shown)


KaptenJansson:
LGTM! @JaspalK headsup that this change affects most demo pages in this repo. Could you make sure to keep an extra keen eye on the javascript logs and the general behavior of the demos while testing?

JaspalK:
KaptenJansson: sure.

I

PR #874, opened 2017-02-10
Fix Ice restart test and demo, fix node server.js warnings and update pem + express

KaptenJansson:
**Description**
* Fixed the Restart ice demo and it's test issues with different getStats variants.
* Updated Pem and Express used for HTTPS server
* Added callback to fs.unlink for the the datachannel test since node 7 requires this
* Remove webrtc-adapter from package.json since we link to it directly.

Qustion: I noticed the IP address does NOT have to change after ICE restart, I guess that's not a requirement this maybe we should check that the ID of the candidate pair has changed instead?

### DIFF
--- a/src/content/datachannel/filetransfer/js/test.js
+++ b/src/content/datachannel/filetransfer/js/test.js
@@ -83,6 +83,8 @@ test('Filetransfer via Datachannels: empty file', function(t) {
     fs.writeFileSync(emptyFilePath, '');
     sendFile(t, emptyFilePath);
     // Remove the empty file.
-    fs.unlink(emptyFilePath);
+    fs.unlink(emptyFilePath, function(error) {
+      console.log('Failed to remove file: ' + error);
+    });
   }
 });
--- a/src/content/peerconnection/multiple/js/test.js
+++ b/src/content/peerconnection/multiple/js/test.js
@@ -26,9 +26,12 @@ test('PeerConnection multiple sample', function(t) {
   })
   .then(function() {
     t.pass('got media');
-    return driver.findElement(webdriver.By.id('callButton')).click();
+    return driver.wait(function() {
+      return driver.findElement(webdriver.By.id('callButton')).isEnabled();
+    });
   })
   .then(function() {
+    driver.findElement(webdriver.By.id('callButton')).click();
     return driver.wait(function() {
       return driver.executeScript(
           'return pc1Remote && pc1Remote.iceConnectionState === \'connected\'' +
--- a/src/content/peerconnection/munge-sdp/js/test.js
+++ b/src/content/peerconnection/munge-sdp/js/test.js
@@ -25,11 +25,18 @@ test('Munge SDP sample', function(t) {
     return driver.findElement(webdriver.By.id('getMedia')).click();
   })
   .then(function() {
+    return driver.wait(function() {
+      return driver.executeScript('return typeof window.localStream ' +
+        '!== \'undefined\'');
+    });
+  })
+  .then(function() {
     t.pass('got media');
     return driver.findElement(webdriver.By.id('createPeerConnection')).click();
   })
   .then(function() {
-    return driver.findElement(webdriver.By.id('createOffer')).click();
+    return driver.wait(webdriver.until.elementIsVisible(
+        driver.findElement(webdriver.By.id('createOffer')))).click();
   })
   .then(function() {
     // Need to wait for createOffer to succeed which takes some time
--- a/src/content/peerconnection/restart-ice/js/main.js
+++ b/src/content/peerconnection/restart-ice/js/main.js
@@ -43,6 +43,11 @@ remoteVideo.onresize = function() {
     var elapsedTime = window.performance.now() - startTime;
     trace('Setup time: ' + elapsedTime.toFixed(3) + 'ms');
     startTime = null;
+    // Have run these functions again in order to get the getStats() reports
+    // with type candidatePair||googCandidatePair and populate the candidate id
+    // elements.
+    checkStats(pc1);
+    checkStats(pc2);
   }
 };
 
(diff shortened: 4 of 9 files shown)


fippo:
checking that the id of the candidate pair has changed sounds like a better test!

src/content/datachannel/filetransfer/js/test.js
```diff
@@ -83,6 +83,8 @@ test('Filetransfer via Datachannels: empty file', function(t) {
     fs.writeFileSync(emptyFilePath, '');
     sendFile(t, emptyFilePath);
     // Remove the empty file.
-    fs.unlink(emptyFilePath);
+    fs.unlink(emptyFilePath, function(error) {
+      console.log('Failed to remove file.');
```
fippo: console.log('Failed to remove file:', error)
-- not sure why the linter doesn't complain about the unhandled error

src/content/peerconnection/restart-ice/js/main.js
```diff
@@ -236,40 +236,47 @@ function onIceStateChange(pc, event) {
...
+    if (activeCandidatePair && activeCandidatePair.remoteCandidateId) {
+      results.forEach(function(report) {
+        if (report.type === 'remote-candidate' &&
+            report.id === activeCandidatePair.remoteCandidateId) {
+          remoteCandidate = report;
         }
       });
     }
-  }
+    if (remoteCandidate && remoteCandidate.ipAddress || remoteCandidate.ip &&
```
fippo: missing () both for ip and port. I'd put both `remoteCandidate.ipAddress || remoteCandidate.ip` and the port on a line of their own for better readability.

src/content/peerconnection/restart-ice/js/main.js
```diff
@@ -236,40 +236,47 @@ function onIceStateChange(pc, event) {
...
+        remoteCandidate.portNumber || remoteCandidate.port) {
+      var ip = typeof remoteCandidate.ipAddress !== 'undefined' ?
```
fippo: I'd just || them, here and elsewhere.

I think this whole block gets better if you do this:
if (remoteCandidate) {
  ip = remoteCandidate.ipAddress || remoteCandidate.ip || 'missing from stats';
  ...
  console.log(...)
}

src/content/peerconnection/restart-ice/js/test.js
```diff
@@ -27,13 +27,21 @@ function getTransportAddresses(stats) {
   var localAddress;
   var remoteAddress;
   stats.forEach(function(report) {
-    if (report.googActiveConnection === 'true') {
+    if (report.googActiveConnection === 'true' ||
+        report.state === 'succeeded') {
```
fippo: I think with the new stats the approach needs to be different. You select the connection, then get the [selectedCandidatePairId](https://w3c.github.io/webrtc-stats/#dom-rtctransportstats-selectedcandidatepairid)

Which gets very complicated however so i'm ok with doing this (which works) in the test.

src/content/peerconnection/restart-ice/js/main.js
```diff
@@ -236,40 +236,47 @@ function onIceStateChange(pc, event) {
...
+function checkStats(pc) {
+  pc.getStats(null).then(function(results) {
+    // figure out the peer's ip
+    var activeCandidatePair = null;
+    var remoteCandidate = null;
+
+    // search for the candidate pair
+    results.forEach(function(report) {
+      console.log(report);
+      if (report.type === 'candidate-pair' && report.state === 'succeeded' ||
```
fippo: removing report.selected will break Firefox. Since we don't run the tests in Firefox thats ok. I volunteer to fix it

fippo (approved):
LGTM with style nits.

This breaks in Firefox since it removes code checking for a now-outdated spec variant but the tests don't run in Firefox and I hereby volunteer to fix that

src/content/peerconnection/restart-ice/js/main.js
fippo: oops, this is not in the test but the main thing :-/

src/content/datachannel/filetransfer/js/test.js
KaptenJansson: haha my bad ;)

src/content/peerconnection/restart-ice/js/main.js
KaptenJansson: I wanted to be explicit using typeof but I see your point. Will fix.

src/content/peerconnection/multiple/js/test.js
```diff
@@ -26,9 +26,12 @@ test('PeerConnection multiple sample', function(t) {
   })
   .then(function() {
     t.pass('got media');
-    return driver.findElement(webdriver.By.id('callButton')).click();
+    return driver.wait(function() {
+      return driver.findElement(webdriver.By.id('callButton')).isEnabled();
+    });
```
KaptenJansson, line 31: Did this on purpose, i.e. want to see which is more stable, isEnabled or elementIsVisible.

src/content/peerconnection/munge-sdp/js/test.js
```diff
@@ -25,11 +25,18 @@ test('Munge SDP sample', function(t) {
...
+        '!== \'undefined\'');
+    });
+  })
+  .then(function() {
     t.pass('got media');
     return driver.findElement(webdriver.By.id('createPeerConnection')).click();
   })
   .then(function() {
-    return driver.findElement(webdriver.By.id('createOffer')).click();
+    return driver.wait(webdriver.until.elementIsVisible(
```
KaptenJansson, line 38: Did this on purpose, i.e. want to see which is more stable, isEnabled or elementIsVisible.

KaptenJansson:
It's still flaky but seems to succeed on the 2nd attempt....

PR #876, opened 2017-02-14
disable ice-restart until we can get it stable

KaptenJansson:
**Description**

**Purpose**


fippo (approved):
How about
skip: true?
in the test call? unless the linter starts acting up then

KaptenJansson:
Did the same for DTMF

PR #881, opened 2017-03-08
candidate gathering: use icegatheringstatechange

fippo:
prefers the icegatheringstatechange over the onicecandidate(null)
way of determining whether gathering is done.

Will land in Chrome shortly, supported in Firefox 53+

cc @henbos

### DIFF
--- a/src/content/peerconnection/trickle-ice/js/main.js
+++ b/src/content/peerconnection/trickle-ice/js/main.js
@@ -106,6 +106,7 @@ function start() {
         ', constraints=' + JSON.stringify(pcConstraints));
   pc = new RTCPeerConnection(config, pcConstraints);
   pc.onicecandidate = iceCallback;
+  pc.onicegatheringstatechange = gatheringStateChange;
   pc.createOffer(
     offerOptions
   ).then(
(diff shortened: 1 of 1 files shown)


aullman:
Hey @fippo we're having the same issue with Firefox 52 and travis-multirunner. Our tests just started failing, getUserMedia is hanging. Any idea what's going on there?

fippo:
hrm... @jan-ivar any idea?

aullman:
Strangely when I install the same version of Ubuntu on my virtual machine and run travis-multirunner manually things seem to work fine. Not sure what I'm missing.

aullman:
@fippo @jan-ivar Any more thoughts on this? To unblock us I ended up modifying travis-multirunner to let me install a specific browser version and install Firefox 51.0.1 which is obviously not ideal.

aiham:
@fippo @jan-ivar On Travis with Firefox 52, `getUserMedia` does not call either success nor failure callbacks and the test case just times out, even when the timeout is increased to 1 minute. Since it's headless we took a screenshot of the browser using image magick's `import` command and it did not show any pending dialogue or indication of why it's hanging. We could not reproduce this with an Ubuntu VM on our own machine using identical steps to the Travis config. Any ideas?

jan-ivar:
Not sure what's going on. I'm interested in any information that could lead to solving this, such as how to reproduce locally. There's [bug 1345954](https://bugzil.la/1345954), but it's not a regression AFAIK, so it doesn't quite fit as a suspect.

aiham:
FYI @jan-ivar adding pulseaudio like #885 fixed FF52 for us.

KaptenJansson:
LGTM

PR #883, opened 2017-03-09
minor improvements in JS file

juandebravo:
**Description**

Simplify JS code in PeerConnection.bandwidth example

**Purpose**

Minor code changes

PR #885, opened 2017-04-07
travis: allow failure in FF

fippo:
FF52 is not behaving well on travis so allow failure :-/

@KaptenJansson can you take another look at #881 and #883 please?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,6 +30,7 @@ matrix:
 
   allow_failures:
     - env: BROWSER=chrome  BVER=unstable
+    - env: BROWSER=firefox BVER=beta
     - env: BROWSER=firefox BVER=nightly
 
 before_install:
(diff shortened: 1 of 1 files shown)


fippo:
ugh. https://github.com/webrtc/utilities/pull/26 should help.

fippo:
so it looks like we're better now but things stil fail :-/

fippo:
FF beta still fails like this in a number of tests:
not ok 34 InvalidArgumentError: Expected [object Undefined] undefined to be a string

i suspect a geckodriver/selenium issue :-/

fippo:
will probably need to update / pin geckodriver and selenium-webdriver in utilities. Or: make a master dev-dependency that has both things as dependencies

aullman:
@fippo am I right that adding that pulseaudio thing fixed the issue with getUserMedia timing out?

aullman:
Oh wow, that totally worked. Good find!

fippo:
current status is: re-enable firefox after starting pulseaudio. Versions are still pinned.

KaptenJansson:
LGTM

PR #888, opened 2017-04-13
all: conditionally use addTrack where available

fippo:
used https://github.com/fippo/webrtc-codemods/blob/master/addTrack
to make the changes.

cc @jan-ivar @foolip

### DIFF
--- a/src/content/capture/canvas-pc/js/main.js
+++ b/src/content/capture/canvas-pc/js/main.js
@@ -77,7 +77,16 @@ function call() {
   };
   pc2.onaddstream = gotRemoteStream;
 
-  pc1.addStream(stream);
+  if (RTCPeerConnection.prototype.addTrack) {
+    stream.getTracks().forEach(
+      function(track) {
+        pc1.addTrack(track);
+      }
+    );
+  } else {
+    pc1.addStream(stream);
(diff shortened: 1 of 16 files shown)


fippo:
oh joy. It seems I am replacing a CallExpression with an if statement but need to replace the CallExpressions parent with the if statement.

(now done; heavy cursing of AST ensued)

henbos:
I don't think we should have a conditional, it adds "adapter.js-like behavior" in order to pass tests.

We should definitely have tests using addTrack, since that is the spec way of doing things. If a browser does not implement it it should fail tests.

If addStream should be supported for legacy/interoperability purposes, we should have tests to cover that, in addition to addTrack. Otherwise we shouldn't have tests that use it.

jan-ivar:
@henbos I agree for tests, but these are samples. Ideally, we'd want adapter.js to polyfill `addTrack` but it's hard to do so. In the interest of showing correct examples, I'll take another stab at it though.

henbos:
Whoops, I was replying as if this was a web-platform-tests change rather than samples!

fippo:
will consider doing the right thing if we manage to shim addTrack. If not... will reopen.

PR #901, opened 2017-05-16
all: use addTrack

fippo:
uses addTrack now that adapter shims it in Chrome.
Edge (the ORTC shim) should also be supported.

This is the result of running the addTrack codemod
```
jscodeshift -t addTrack src/content
```
https://github.com/fippo/webrtc-codemods

PR #902, opened 2017-05-16
all: use ontrack instead of onaddstream

fippo:
no codemod so extra 👓 👀 please

based on #901 

Some samples reuse the video element. This means the !videoEl.srcObject check does not work.


KaptenJansson:
LGTM

PR #906, opened 2017-05-27
samples: add upgrade test

fippo:
adds a sample for upgrading an audio call to an audio-video call.

(I needed something to test Edge and it has shown at least one issue in Chrome too... working around it)

### DIFF
--- a/src/content/peerconnection/upgrade/css/main.css
+++ b/src/content/peerconnection/upgrade/css/main.css
@@ -1,0 +1,47 @@
+/*
+ *  Copyright (c) 2017 The WebRTC project authors. All Rights Reserved.
+ *
+ *  Use of this source code is governed by a BSD-style license
+ *  that can be found in the LICENSE file in the root of the source
+ *  tree.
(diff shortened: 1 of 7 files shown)


fippo:
Not sure how the UX of this should be so I went for a minimal approach. Using video first and then adding audio would be a way but this makes testing harder.

src/content/peerconnection/upgrade/css/main.css
```diff
@@ -0,0 +1,47 @@
+/*
+ *  Copyright (c) 2015 The WebRTC project authors. All Rights Reserved.
```
KaptenJansson: Please set the copyright to the current year for new files.

KaptenJansson (approved):
LGTM with a nit.

KaptenJansson:
Merging this regardless of test failure since it's unrelated. Will try to figure out what's causing it.

fippo:
its the lack of addTrack which should be fixed in adapter soon

PR #909, opened 2017-06-15
try things

KaptenJansson:
**Description**

**Purpose**

### DIFF
--- a/src/content/capture/canvas-pc/js/test.js
+++ b/src/content/capture/canvas-pc/js/test.js
@@ -22,6 +22,9 @@ test('PeerConnection pc1 sample', function(t) {
       '/src/content/peerconnection/pc1/index.html')
   .then(function() {
     t.pass('page loaded');
+    // Override the trace function to ensure console logging works for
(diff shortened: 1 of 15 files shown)


src/content/capture/canvas-pc/js/test.js
```diff
@@ -22,6 +22,9 @@ test('PeerConnection pc1 sample', function(t) {
       '/src/content/peerconnection/pc1/index.html')
   .then(function() {
     t.pass('page loaded');
+    // Override the trace function to ensure console logging works for
+    // webdriver.
+    seleniumHelpers.overrideTrace(driver);
```
fippo, line 27: i would do this in buildDriver (which could reset/flush the output... hopefully)

KaptenJansson: I thought about that but wanted the test to explicitly do it in case we want the original trace logging but I guess that's something we could address with a toggle method later if needed in buildDriver. Will do as you suggest.

PR #910, opened 2017-06-15
constraints: make it work with spec-stats

fippo:
makes the "connected to" in the constraints sample work with spec-stats

@henbos can you please take a look?

### DIFF
--- a/src/content/peerconnection/constraints/js/main.js
+++ b/src/content/peerconnection/constraints/js/main.js
@@ -247,22 +247,35 @@ setInterval(function() {
       var activeCandidatePair = null;
       var remoteCandidate = null;
 
-      // search for the candidate pair
+      // Search for the candidate pair, spec-way first.
       results.forEach(function(report) {
-        if (report.type === 'candidatepair' && report.selected ||
-            report.type === 'googCandidatePair' &&
-            report.googActiveConnection === 'true') {
-          activeCandidatePair = report;
+        if (report.type === 'transport') {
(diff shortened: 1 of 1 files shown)


src/content/peerconnection/constraints/js/main.js
```diff
@@ -249,20 +249,31 @@ setInterval(function() {
...
-            report.type === 'googCandidatePair' &&
-            report.googActiveConnection === 'true') {
-          activeCandidatePair = report;
+        if (report.type === 'transport') { 
+          activeCandidatePair = results.get(report.selectedCandidatePairId); 
         }
       });
+      if (!activeCandidatePair) {
+        results.forEach(function(report) {
+          if (report.type === 'candidate-pair' && report.selected ||
```
fippo: using candidate-pair here fix things in Firefox too

henbos: Can you merge these two searches and add a comment that transport.selectedCandidatePairId is spec way and candidatePair.selected is Firefox way?

henbos (approved):

src/content/peerconnection/constraints/js/main.js
fippo: comments added. I think it is easier to read and understand what is the legacy fallback when the forEach's are separated

fippo: also added a comment about ipAddress and portNumber below.

KaptenJansson:
LGTM

PR #912, opened 2017-06-18
devices: enable output dropdown based on sinkId availability

fippo:
checks whether the sinkId is settable before allowing to choose output.


fippo:
@jan-ivar can you take a look? Mostly because you probably know whether it is safe to assume both will be implemented at the same time

jan-ivar (approved):
lgtm.

KaptenJansson (approved):

PR #923, opened 2017-08-19
trickle-ice: fix ip handling note, add check

fippo:
clarifies the ip handling note and adds a check if getUserMedia
permissions have been given.

@juberti can you take a look? This confused @nils-ohlmeier recently


KaptenJansson (approved):

nils-ohlmeier:
For the record it not only confused me, but also people who opened issues against specs pointing to exactly this page here :-)
So thanks a lot for making the world a better place!

fippo:
I should have said this confused *even* you ;-)

PR #935, opened 2017-10-19
Add a slider for doing applyConstraints.

alvestrand:
**Description**
This PR does two changes:

- Uses in-frame error message rather than alert()
- Adds a slider that can be used to do applyConstraints()
on the video track after having opened it.

**Purpose**
To make the effect of calling applyConstraints() visible.

### DIFF
--- a/src/content/getusermedia/resolution/js/main.js
+++ b/src/content/getusermedia/resolution/js/main.js
@@ -17,6 +17,15 @@ var qvgaButton = document.querySelector('#qvga');
 var hdButton = document.querySelector('#hd');
 var fullHdButton = document.querySelector('#full-hd');
 
+var videoblock = document.querySelector('#videoblock');
+var messagebox = document.querySelector('#errormessage');
+
+var widthInput = document.querySelector('div#width input');
+var widthOutput = document.querySelector('div#width span');
+
+var currentWidth = 0;
+var currentHeight = 0;
+
 vgaButton.onclick = function() {
   getMedia(vgaConstraints);
 };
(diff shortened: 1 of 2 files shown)


src/content/getusermedia/resolution/js/main.js
```diff
@@ -52,6 +59,16 @@ var fullHdConstraints = {
 function gotStream(mediaStream) {
   window.stream = mediaStream; // stream available to console
   video.srcObject = mediaStream;
+  messagebox.style.display = 'none';
+  videoblock.style.display = 'block';
+}
+
+function errorMessage(who, what) {
+  let message = who + ': ' + what;
+  messagebox.innerHTML = message;
```
fippo: i'd go for .innerText here -- mostly because .innerHTML is 👿

src/content/getusermedia/resolution/js/main.js
```diff
@@ -60,10 +77,26 @@ function displayVideoDimensions() {
   }
   dimensions.innerHTML = 'Actual video dimensions: ' + video.videoWidth +
```
fippo: innerText

src/content/getusermedia/resolution/js/main.js
```diff
@@ -60,10 +77,26 @@ function displayVideoDimensions() {
...
+  widthInput.value = video.videoWidth;
+  widthOutput.textContent = video.videoWidth;
 }
 
 video.onloadedmetadata = displayVideoDimensions;
 
+function constraintChange(e) {
+  widthOutput.textContent = e.target.value;
+  let track = window.stream.getVideoTracks()[0];
+  let constraints = {video: {width: {exact: e.target.value}}};
```
guidou: This shouldn't have  the "video:" attribute since the track already knows that it's a video track.
|constraints| should be {width: {exact: e.target.value}}

guidou: Note that Firefox only supports native resolutions, so it will reject if there is no native mode with the given target value.

guidou:
lgtm, but look at fippo's comment

src/content/getusermedia/resolution/js/main.js
```diff
@@ -52,17 +59,62 @@ var fullHdConstraints = {
...
+    console.log(dimensions.innerHTML);
+  } else {
+    dimensions.innerHTML = 'Video not ready';
+  }
 }
 
 video.onloadedmetadata = displayVideoDimensions;
+// Video size takes a few seconds to change when size changes,
+// so we track it every half second.
+setInterval(displayVideoDimensions, 500);
```
fippo: actually (and thanks @pehrsons) this is https://bugs.chromium.org/p/chromium/issues/detail?id=601470 you're working around here. Maybe fix the actual bug? :-)

src/content/getusermedia/resolution/js/main.js
alvestrand: done

alvestrand:
PTAL - it turns out that video.onresize works reliably (so perhaps that other bug is gone?)

src/content/getusermedia/resolution/js/main.js
```diff
@@ -52,17 +61,71 @@ var fullHdConstraints = {
...
+function clearErrorMessage() {
+  messagebox.style.display = 'none';
 }
 
-function displayVideoDimensions() {
-  if (!video.videoWidth) {
-    setTimeout(displayVideoDimensions, 500);
+function displayVideoDimensions(whereSeen) {
+  if (video.videoWidth) {
+    dimensions.innerHTML = 'Actual video dimensions: ' + video.videoWidth +
```
fippo: innerText (just search and replace)

fippo:
ignore the firefox CI failure, @KaptenJansson or me will look at it elsewhere (maybe...). Still innerHTML nits but onresize seems to work (i just checked) so @guidou can close that chrome bug :-)

PR #939, opened 2017-10-23
Add aspect and size locks to resolution demo

alvestrand:
This allows to constrain the size of the video (for
watching pixellated images) or the aspect ratio (for
making sure we get the same aspect ratio).


alvestrand:
Asking @jan-ivar to take a look.
It illustrates scaling nicely.

fippo (approved):
I can see it does things. Would be good to have tests which describe the expectations -- but for that I would like to have chai instead of tape ;-)

fippo:
@alvestrand did this land in M63?

PR #958, opened 2017-11-29
Update deps

KaptenJansson:
**Description**
Update eslint
Add .eslintcache to gitignore
Update webdriver and chromedriver.

**Purpose**
Fix breakages and npm warnings

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,16 +26,19 @@
     "eslint-config-webrtc": "^1.0.0",
     "express": "^4.14.1",
     "faucet": "0.0.1",
+    "geckodriver": "1.4.0",
     "grunt": "^0.4.5",
     "grunt-cli": ">=0.1.9",
     "grunt-contrib-compress": "^0.13.0",
     "grunt-contrib-csslint": ">=0.3.1",
-    "grunt-eslint": "^17.2.0",
+    "grunt-eslint": "^18.1.0",
     "grunt-githooks": "^0.3.1",
     "grunt-htmlhint": ">=0.9.12",
     "pem": "^1.9.4",
-    "webrtc-utilities": "^1.0.0",
-    "geckodriver": "1.4.0",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -26,16 +26,19 @@
     "eslint-config-webrtc": "^1.0.0",
     "express": "^4.14.1",
     "faucet": "0.0.1",
+    "geckodriver": "1.4.0",
```
fippo: pinning should not be required... also this should pull from utils?

package.json
```diff
@@ -26,16 +26,19 @@
...
     "grunt-htmlhint": ">=0.9.12",
     "pem": "^1.9.4",
-    "webrtc-utilities": "^1.0.0",
-    "geckodriver": "1.4.0",
-    "selenium-webdriver": "3.3.0"
+    "selenium-webdriver": "^3.4.0",
+    "webrtc-utilities": "^1.0.0"
+  },
+  "dependencies": {
+    "chromedriver": "^2.33.2"
```
fippo: should be dev-dependency (and come from utils)

package.json
KaptenJansson: Its strange, when pulled from utilities, it's installed it says, but does not execute all the steps (when I do an explicit install I see it's added to node modules) and thus is not in node_modules nor in .bin but it's in the tree if I do npm ls.....

package.json
KaptenJansson: right my bad from testing.

package.json
KaptenJansson: Aha, I know why, utilities do not have geckodriver listed as a devDeps, just a normal deps.

fippo: if its a dep it will be puled and installed. if its a dev-dep it wont. Lets get rid of the utils and pull everything directly?

PR #966, opened 2017-12-05
remove package-lock

fippo:
with fire


fippo:
I don't think package-lock mades sense since we only have dev dependencies

KaptenJansson (approved):

PR #972, opened 2017-12-08
dtmf: do not use addTrack in chrome

fippo:
since addTrack + the DTMF shim from adapter do not work
together. See https://github.com/webrtc/adapter/issues/733


fippo:
let me also try to fix the test which is deactivated

fippo:
the DTMF test has been disabled due to flakyness in the past and is the main source of flakyness in adapter... lets see if its a good idea to reenable...

henbos (approved):
lgtm

PR #992, opened 2018-01-12
all: remove legacy pc constraints

fippo:
this was removed from the spec long ago and not used meaningful anymore
with the exception of the trickle-ice page. The use there was has been removed.

### DIFF
--- a/src/content/datachannel/basic/js/main.js
+++ b/src/content/datachannel/basic/js/main.js
@@ -12,7 +12,6 @@ var localConnection;
 var remoteConnection;
 var sendChannel;
 var receiveChannel;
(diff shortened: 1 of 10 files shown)


fippo:
remarked [here](https://stackoverflow.com/questions/48230970/what-is-was-the-two-parameter-constructor-for-rtcpeerconnection)

fippo:
needed to update fake device names (and cleaned up a TODO for FF46...) in a separate commit.

I don't recall seeing anything in the M64 release notes about that. @KaptenJansson can you check please?

KaptenJansson (approved):
Thanks for doing this!

PR #993, opened 2018-01-16
Add a demo for "Unprocessed volume"

alvestrand:
**Description**
Add a demo that demonstrates the difference between processed and unprocessed audio volume.

This doesn't work right as of Chrome 63 (and probably 65); the PR
is created in order to get the implementation work going.


src/content/getusermedia/unprocessed-volume/index.html
```diff
@@ -0,0 +1,76 @@
+<!DOCTYPE html>
+<!--
+ *  Copyright (c) 2015 The WebRTC project authors. All Rights Reserved.
```
KaptenJansson: Update copyright year.

src/content/getusermedia/unprocessed-volume/js/main.js
```diff
@@ -0,0 +1,85 @@
+/*
+ *  Copyright (c) 2015 The WebRTC project authors. All Rights Reserved.
```
KaptenJansson: Update copyright

src/content/getusermedia/unprocessed-volume/js/main.js
```diff
@@ -0,0 +1,85 @@
...
+      instantMeter.value = instantValueDisplay.innerText =
+          soundMeter.instant.toFixed(2);
+      slowMeter.value = slowValueDisplay.innerText =
+          soundMeter.slow.toFixed(2);
+    }, 200);
+  });
+  console.log('First track settings:',
+              JSON.stringify(stream.getAudioTracks()[0].getSettings()));
+  // Set up second track with audio processing disabled
+  constraints.audio = {echoCancellation: {exact: false}};
```
KaptenJansson: Why do you have to use exact here?

src/content/getusermedia/unprocessed-volume/js/main.js
```diff
@@ -0,0 +1,85 @@
...
+  video: false
+};
+
+function handleSuccess(stream) {
+  // Put variables in global scope to make them available to the
+  // browser console.
+  window.stream = stream;
+  var soundMeter = window.soundMeter = new SoundMeter(window.audioContext);
+  soundMeter.connectToSource(stream, function(e) {
+    if (e) {
```
KaptenJansson: You could expand `e` to error since you have space. Also not sure if alert is preferred anymore but not outputting text on the page might not be clear enough.

src/content/getusermedia/unprocessed-volume/js/main.js
```diff
@@ -0,0 +1,85 @@
...
+var unprocessedMeter = document.querySelector('#unprocessed meter');
+
+var instantValueDisplay = document.querySelector('#instant .value');
+var slowValueDisplay = document.querySelector('#slow .value');
+var unprocessedValueDisplay = document.querySelector('#unprocessed .value');
+
+try {
+  window.AudioContext = window.AudioContext || window.webkitAudioContext;
+  window.audioContext = new AudioContext();
+} catch (e) {
```
KaptenJansson, line 25: You are not printing the actual error, `e` is unused.

src/content/getusermedia/unprocessed-volume/js/main.js
```diff
@@ -0,0 +1,85 @@
...
+      return;
+    }
+    setInterval(function() {
+      instantMeter.value = instantValueDisplay.innerText =
+          soundMeter.instant.toFixed(2);
+      slowMeter.value = slowValueDisplay.innerText =
+          soundMeter.slow.toFixed(2);
+    }, 200);
+  });
+  console.log('First track settings:',
```
KaptenJansson: Use trace instead?

src/content/getusermedia/unprocessed-volume/js/main.js
```diff
@@ -0,0 +1,85 @@
...
+      return;
+    }
+    setInterval(function() {
+      unprocessedMeter.value = unprocessedValueDisplay.innerText =
+          unprocMeter.slow.toFixed(2);
+    }, 200);
+  });
+}
+
+function handleError(error) {
```
KaptenJansson: You could make a generic error handler here and pass it to connectToSource error callbacks as well.

KaptenJansson (dismissed):

KaptenJansson:
Any cycles to finish this?

KaptenJansson:
Ping

src/content/getusermedia/unprocessed-volume/index.html
```diff
@@ -0,0 +1,76 @@
...
+        <div class="value"></div>
+      </div>
+    </div>
+
+    <p>The 'instant' volume changes approximately every 50ms; the 'slow' volume approximates the average volume over about a second.</p>
+    <p>The third meter shows the volume of a track without audio processing applied; this should be the "input volume".</p>
+    <p>Note that you will not hear your own voice; use the <a href="../audio">local audio rendering demo</a> for that.</p>
+    <p>The <code>audioContext</code>, <code>stream</code> and <code>soundMeter</code> variables are in global scope, so you can inspect them from the console.</p>
+
+    <a href="https://github.com/webrtc/samples/tree/gh-pages/src/content/getusermedia/volume" title="View source for this page on GitHub" id="viewSource">View source on GitHub</a>
```
fippo: wrong link

ErikHellman:
Fixed all the outstanding issues and converted all vars to const/let.

Seems like the CLA check doesn't like my other email. I'll fix this for future PRs, but they

src/content/getusermedia/unprocessed-volume/js/main.js
henbos: I believe you can do try { } catch { }.

src/content/getusermedia/unprocessed-volume/js/main.js
henbos: +1

src/content/getusermedia/unprocessed-volume/js/main.js
```diff
@@ -0,0 +1,88 @@
...
+try {
+  window.AudioContext = window.AudioContext || window.webkitAudioContext;
+  window.audioContext = new AudioContext();
+} catch (e) {
+  alert('Web Audio API not supported.');
+}
+
+// Put variables in global scope to make them available to the browser console.
+let constraints = window.constraints = {
+  audio: {echoCancellation: true},
```
henbos, line 31: Should this be echoCancellation: {exact:true} so that we fail if it cannot be satisfied?

src/content/getusermedia/unprocessed-volume/js/main.js
henbos: I think exact is correct if we want two tracks, one with and one without echoCancellation, and demo the difference.

src/content/getusermedia/unprocessed-volume/js/main.js
```diff
@@ -0,0 +1,88 @@
...
+const errorMsg = document.querySelector('#errorMsg')
+
+try {
+  window.AudioContext = window.AudioContext || window.webkitAudioContext;
+  window.audioContext = new AudioContext();
+} catch (e) {
+  alert('Web Audio API not supported.');
+}
+
+// Put variables in global scope to make them available to the browser console.
```
henbos, line 29: I'm not sure what what this means.

src/content/getusermedia/unprocessed-volume/js/main.js
henbos: I find it confusing and error prone that the same constraints are setup up as a global variable, then used, then modified. If the demo were to have a reset button it would stop working unless this is explicitly reset. I've also not seen "let constraints = window.constraints = ...". Wouldn't "let constraints" work? Or if that didn't work, why not "var constraints"? I don't see this variable adding anything.

Can we just pass the desired constraints when we call getUserMedia? Not have a variable for it.

src/content/getusermedia/unprocessed-volume/js/main.js
```diff
@@ -0,0 +1,88 @@
...
+let constraints = window.constraints = {
+  audio: {echoCancellation: true},
+  video: false
+}
+
+function handleSuccess(stream) {
```
henbos: Can we do async/await instead? Promise-chains with callbacks are harder to follow.

henbos (changes requested):

henbos:
Did I review the latest changes or did googlebot block you? I saw comments from February on the files I commented on

src/content/getusermedia/unprocessed-volume/js/main.js
ErikHellman: While I agree that it is easier to read, I'm not sure it's suitable for these samples. My guess is that most JavaScript developers are more familiar with promises than async/await.

src/content/getusermedia/unprocessed-volume/js/main.js
ErikHellman: Ping @alvestrand. I didn't update this part. :)

src/content/getusermedia/unprocessed-volume/js/main.js
fippo: that is the reason there was no push for es6... stackoverflow is still full of people who use the callback API but we had to move the needle there

ErikHellman: I've changed my mind on this. async/await is better in most cases than promise-chains. We should be good citizens and push the use of that. :)

ErikHellman:
@alvestrand I need your approval on this so that the Google CLA gets happy again. :)

src/content/getusermedia/unprocessed-volume/js/main.js
fippo: sam wrote this... but var at top scope have always been available in the console or window[]. No longer true for es6 const/let notably.

src/content/getusermedia/unprocessed-volume/js/main.js
```diff
@@ -0,0 +1,105 @@
...
+    }, 200);
+  } catch (e) {
+    handleError(e);
+  }
+
+  const audioSettings = JSON.stringify(stream.getAudioTracks()[0].getSettings());
+  console.log('First track settings:', audioSettings);
+
+  // Set up second track with audio processing disabled
+  constraints.audio = {echoCancellation: {exact: false}};
```
fippo, line 59: any reason why this is {exact: false} and not false?

ErikHellman:
Closing this for now as we're not ready to publish this yet.

PR #1000, opened 2018-01-25
Issue 961 - Updating the test file with a title and text

rantonysamy:
As requested in the issue 961, https://github.com/webrtc/samples/issues/961 Updating the test page with a title.

**Description**
Adding a title and a link to go back to the previous page

**Purpose**
If user ends up in this page, It gives no clue about the test. so adding a title, and a link to go back to the previous page.

### DIFF
--- a/src/content/peerconnection/pr-answer/index.html
+++ b/src/content/peerconnection/pr-answer/index.html
@@ -8,27 +8,50 @@
 -->
 <html>
 <head>
-  <title>PeerConnection PRANSWER Demo</title>
-  <link rel="icon" sizes="192x192" href="../../../images/webrtc-icon-192x192.png">
-  <!-- Load the polyfill to switch-hit between Chrome and Firefox -->
-  <style>
-    video {
-      border:5px solid black;
-      width:320px;
-      height:240px;
-    }
-  </style>
+    <meta charset="utf-8">
+    <meta name="description" content="WebRTC code samples">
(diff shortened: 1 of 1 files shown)


KaptenJansson (changes requested):
Could you make it look like the other samples by applying the the same CSS styles etc?

fippo:
also indent (makes me wonder if there is htmllint because arguments about indent are the worst :-))

KaptenJansson:
Yes please fix indents and it's weird that HTMLhint does not trigger on this, might be config issue, will check.

KaptenJansson:
https://github.com/yaniswang/HTMLHint/wiki/Rules indicates it does not check for indentation.....

KaptenJansson:
https://github.com/htmllint/htmllint/wiki/Options has indent checks. Will try it out.

fippo:
@samdutton do you have an idea why htmllint does not behave as we expect?

samdutton:
@fippo @KaptenJansson I don't get any errors reported by HTMLHint for the HTML in this file, but I do get an error that the CSS rules aren't in alphabetical order.

fippo:
oh so there is htmlhint and htmllint... splendid!

KaptenJansson:
@rantonysamy  please do another pass here and try to get the indendation correct.

KaptenJansson:
Ping

PR #1004, opened 2018-01-30
pc1: add getstats call after close

fippo:
to have some real-world coverage

### DIFF
--- a/src/content/peerconnection/pc1/js/main.js
+++ b/src/content/peerconnection/pc1/js/main.js
@@ -234,4 +234,5 @@ function hangup() {
   pc2 = null;
   hangupButton.disabled = true;
   callButton.disabled = false;
+  pc1.getStats(); // just to have something testing getstats-after-close.
 }


src/content/peerconnection/pc1/js/main.js
```diff
@@ -234,4 +234,5 @@ function hangup() {
   pc2 = null;
   hangupButton.disabled = true;
   callButton.disabled = false;
+  pc1.getStats(); // just to have something testing getstats-after-close.
```
alvestrand, line 237: can you do a .then(update some text field with some random stat)?
would be nice to have "total bytes sent" from the transport stats displayed. If you want to be nasty, display ", X video, X audio" with numbers from the tracks, which will expose the fact that they've been zeroed by the patch I'm about to write.

fippo: lets do that in a different sample/PR? Thinking of updating constraints.html...

KaptenJansson (approved):
LGTM as is.

PR #1009, opened 2018-02-03
audio: use replaceTrack for "hard" mute

fippo:
**Description**
uses replaceTrack for hard muting. This is different from soft mute, i.e. setting track.enabled = false
which (in chrome) continues to send 50 packets/s at a slightly lower bitrate

**Purpose**
I want to use replaceTrack(null) so I would like to see some basic QA :-)

@henbos PTAL since you wanted this as well.
@KaptenJansson wdyt about adding "QA" instructions on the page?
@jan-ivar replaceTrack(null) is only supported in 59+, right? Graphs freeze after replaceTrack(null), afaics getStats() does not return any rtp stats?

### DIFF
--- a/src/content/peerconnection/audio/index.html
+++ b/src/content/peerconnection/audio/index.html
@@ -56,7 +56,8 @@
         <option value="PCMU">PCMU</option>
       </select>
       <button id="callButton">Call</button>
-      <button id="hangupButton">Hang Up</button>
+      <button id="muteButton" disabled>Mute</button>
+      <button id="hangupButton" disabled>Hang Up</button>
     </div>
     <div class="graph-container" id="bitrateGraph">
       <div>Bitrate</div>
--- a/src/content/peerconnection/audio/js/main.js
+++ b/src/content/peerconnection/audio/js/main.js
@@ -12,11 +12,14 @@
 var audio2 = document.querySelector('audio#audio2');
 var callButton = document.querySelector('button#callButton');
 var hangupButton = document.querySelector('button#hangupButton');
+var muteButton = document.querySelector('button#muteButton');
 var codecSelector = document.querySelector('select#codec');
-hangupButton.disabled = true;
 callButton.onclick = call;
 hangupButton.onclick = hangup;
+muteButton.onclick = toggleMute;
 
+var supportsReplaceTrack = window.RTCRtpSender &&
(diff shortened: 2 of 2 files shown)


fippo:
so in the "stage 2" model there is an issue. Suppose I have an audio/video connection. Then I replace both the audio and the video track with null.
Then I want to re-attach the audio track. How do I find it when using getSenders()? Since he have no track, we have no kind.

In the transceiver model this is pretty easy to solve. Iterate over getTransceivers(), find the sender where track is null and the receivers track.kind is audio.

Stuck in the middle.

jan-ivar:
I'm against this, because:
 1. It seems like a hack to work around one implementation not following spec recommendation.
 2. It undermines `track.enabled` as the spec-preferred API for muting,
 3. The concept of "soft"/"hard" mute is confusing, and
 4. It doesn't let browsers know whether cameras and microphones are still in use.

The samples should help guide people to using the spec, not work around Chrome bugs. I don't see that the "soft"/"hard" distinction serves any other purpose here.

The fourth point is important to us at Mozilla, because Firefox 60 now implements [turning off the camera on mute](https://bugzilla.mozilla.org/show_bug.cgi?id=1299515), a key step to getting sites like Hangouts to stop using `track.stop()` for face mute (which causes unnecessary grief/redundant permission re-prompts for privacy-sensitive Firefox users).

> which (in chrome) continues to send 50 packets/s at a slightly lower bitrate

This is a Chrome issue IMHO. Spec now [says](http://w3c.github.io/webrtc-pc/#dom-rtcrtpsender-track): *"In the case of video, the RTCRtpSender SHOULD send one black frame per second."*

fippo:
@jan-ivar i think you're more opposed to #996 then which shows that technique.

But we need some way to QA this and doing it here where we have both @KaptenJansson+crew and the general public prod at it...

jan-ivar:
> How do I find it when using getSenders()? Since he have no track, we have no kind.

I suggest storing them as `videoSender` and `audioSender`.

henbos (approved):

henbos (changes requested):
Sorry I missed the discussion on this issue...

fippo:
probably move this stuff to https://github.com/webrtc/test-pages

PR #1011, opened 2018-02-05
remove extra getStats call

fippo:
it did not even work because the peerconnection was null'd.
Also we have updated the constraints sample to make sure this
case is covered


KaptenJansson:
Fixes #1011

PR #1015, opened 2018-02-08
pc1: remove getStats call

fippo:
we have a better place for getstats-after-close now. Also this was
just throwing an exception


fippo:
already done in #1011

PR #1017, opened 2018-02-16
Fix: Disabled call button when there is an ongoing call

### LINKED ISSUE #986: Use case of Call and Accept button needs a review. [causing javascript errors]
**Browsers and versions affected**
64.0.3282.39 / Macbook High sierra, Macbook Air El Caption, Lenovo win 7, Asus win 10 (Possible in all other OS and devices.)

**Steps to reproduce**
0. Browse to https://webrtc.github.io/samples/src/content/peerconnection/pr-answer/
1. Open the javascript console (alt+cmd+j or ctr+shift+j) and watch for any errors
2. Check if the local video is seen in the left pane
3. Press "Call"
4. Press "Accept" 
keep repeating step 3 and 4. 

**Expected results**
Solution 1: 
Disable call button if Accept button is clicked. Thus we can avoid the javascript error.
Solution 2:
If there is a reason like checking the relay (adding more calls) then inform the user say for example 5 calls is the maximum we support.  

**Actual results**
It is possible to call so many calls and accept it. Thus at one some point no video is seen in the right box. Also there are so many javascript exceptions thrown. 

**100% reproducible**
[webrtc.github.io-1514535525036.log](https://github.com/webrtc/samples/files/1592818/webrtc.github.io-1514535525036.log)
[webrtc.github.io-1514535283679.log](https://github.com/webrtc/samples/files/1592819/webrtc.github.io-1514535283679.log)

vkkkumar:
**Description**
Disabled call button when there is an ongoing call. It avoids the javascript error as mentioned in the linked bug.

**Purpose**

Fixes #986

### DIFF
--- a/src/content/peerconnection/pr-answer/js/main.js
+++ b/src/content/peerconnection/pr-answer/js/main.js
@@ -10,17 +10,17 @@
 
 var vid1 = document.getElementById('vid1');
 var vid2 = document.getElementById('vid2');
-var btn1 = document.getElementById('btn1');
-var btn2 = document.getElementById('btn2');
-var btn3 = document.getElementById('btn3');
+var callbtn = document.getElementById('callbtn');
+var acceptbtn = document.getElementById('acceptbtn');
+var hangupbtn = document.getElementById('hangupbtn');
 
-btn1.addEventListener('click', start);
-btn2.addEventListener('click', accept);
-btn3.addEventListener('click', stop);
+callbtn.addEventListener('click', start);
+acceptbtn.addEventListener('click', accept);
+hangupbtn.addEventListener('click', stop);
 
-btn1.disabled = true;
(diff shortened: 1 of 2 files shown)


src/content/peerconnection/pr-answer/js/main.js
```diff
@@ -156,7 +156,7 @@ function accept() {
     onCreateAnswerError
   );
   btn2.disabled = true;
-  btn1.disabled = false;
+  btn1.disabled = true;
```
KaptenJansson: Great, but could you update the button names to what they do (e.g call etc)? btn2 and btn1 are horrible names,

KaptenJansson (changes requested):

KaptenJansson:
Great, thanks!

Issue #974, opened 2017-12-11
Boundary values can be provided for this example (Usability feedback)


vkkkumar:
**Browsers and versions affected**
All

**Steps to reproduce**
0. Browse to https://webrtc.github.io/samples/src/content/datachannel/datatransfer/
1. Open the javascript console (alt+cmd+j or ctr+shift+j) and watch for any errors
2. Click on "Generate and Send data" for the default size 128 MB
3. Enter a very large number for example (15000)

**Expected results**
Provide the supported boundary values. It takes so much time if given very large time. or Provide a text it will take more time beyond a limit.

**Actual results**
no warning that this value is very large value and no clue how long it will take.

KaptenJansson:
Feel free to add some text in a PR warning the user about high values.

ErikHellman:
I added a warning about large values. Since the limit depends on many different factors, it is difficult to provide an upper limit.

PR #1018, opened 2018-02-16
Adding a warning when the values are more than 1000

vkkkumar:
**Description**
Adding a warning when the values are more than 1000

**Purpose**
This is an improvement to the existing samples page. When a user enters a very large value it can misguide that the progress is not happening. By adding this note it informs the user to wait patiently or try with a smaller value. 

#974


src/content/datachannel/datatransfer/js/main.js
```diff
@@ -29,6 +29,9 @@ megsToSend.addEventListener('change', function(e) {
   if (this.value <= 0) {
     sendButton.disabled = true;
     errorMessage.innerHTML = '<p>Please enter a number greater than zero.</p>';
+  } else if (this.value > 1000) {
+        errorMessage.innerHTML = '<p>Progress can be slow when the values are greater than 1000.</p>';
```
KaptenJansson: Indentation should match the above `if` statement. Also this is not really an error, its a warning and should be displayed as such.

KaptenJansson (changes requested):

KaptenJansson:
Ping

ErikHellman:
This PR is replaced by https://github.com/webrtc/samples/pull/1079

PR #1022, opened 2018-03-08
Remove face demo

### LINKED ISSUE #757: Face tracking demo video is not shown in Chrome 50.0.2645.4 dev (64-bit)
[Face tracking](https://webrtc.github.io/samples/src/content/getusermedia/face/) demo video is not shown in Chrome 50.0.2645.4 dev (64-bit). The app actually already gets video, because you can see the rectangle of the detection of your face. But the video just does not show.
The demo works well in old version Chrome.

I am not sure this is because of Chrome bug or the demo has bug. Because other demos work well in Chrome 50.0.2645.4 dev (64-bit).

KaptenJansson:
**Description**
Fixes #757 

**Purpose**


samdutton:
LGTM though we also need to remove the link from index.html (which corresponds to https://webrtc.github.io/samples/).

samdutton (approved):
I will remove the link from index.html as well.

samdutton:
Confirm that all authors are ok with their commits being contributed to this project.

PR #1034, opened 2018-03-28
ice restart: copy "are you connected" logic from constraints

fippo:
updated logic for chrome spec stats. With this one can see how the
local candidate id changes during an ice restart.

### DIFF
--- a/src/content/peerconnection/restart-ice/js/main.js
+++ b/src/content/peerconnection/restart-ice/js/main.js
@@ -44,7 +44,7 @@ remoteVideo.onresize = function() {
     trace('Setup time: ' + elapsedTime.toFixed(3) + 'ms');
     startTime = null;
     // Have run these functions again in order to get the getStats() reports
-    // with type candidatePair||googCandidatePair and populate the candidate id
+    // with type candidatePair and populate the candidate id
     // elements.
     checkStats(pc1);
     checkStats(pc2);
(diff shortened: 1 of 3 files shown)


fippo:
so the local candidate id changes but the remote does not. I think that is because the right side is only updated when the iceconnectionstate changes... might be worth updating both local and remote when either connection state changes. Or to poll like constraints does.

src/content/peerconnection/restart-ice/js/main.js
```diff
@@ -261,14 +261,22 @@ function checkStats(pc) {
...
-        activeCandidatePair = report;
+      if (report.type === 'transport') {
+        activeCandidatePair = results.get(report.selectedCandidatePairId);
       }
     });
+    // Fallback for Firefox and Chrome legacy stats.
+    if (!activeCandidatePair) {
+      results.forEach(function(report) {
+        if (report.type === 'candidate-pair' && report.state === 'succeeded' &&
+            report.selected || report.type === 'googCandidatePair' &&
```
alvestrand: It bothers me to see things broken up this way. Can you add parentheses and linebreaks so that the || appears visually between the && sequences?

fippo: re-reviewing this I should actually remove the legacy way :-)
Will do here and in constraints

alvestrand: Does that mean it'll stop working on version `<old>`?` Better document that in a comment or something...

fippo: no backward compat contract here. spec-stats have been implemented for way over a year so...

fippo:
there were more demos using legacy stats :-|

PR #1035, opened 2018-03-29
audio/bandwidth/constraints: get rid of legacy stats

fippo:
gets rid of chrome-legacy stats and non-hyphenated stats types

@henbos / @jan-ivar  PTAL. There is still a case of goog-stats in the test that i'll try to get rid of as well (now done)

### DIFF
--- a/src/content/peerconnection/audio/js/main.js
+++ b/src/content/peerconnection/audio/js/main.js
@@ -283,9 +283,7 @@ window.setInterval(function() {
       var bytes;
       var packets;
       var now = report.timestamp;
-      if ((report.type === 'outboundrtp') ||
-          (report.type === 'outbound-rtp') ||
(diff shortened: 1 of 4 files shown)


fippo:
Q: should this prefer RTCRtpSender.getStats when available? @henbos also you should update samples *before* intending to ship :-)

henbos (commented):
RTCRtpSender.getStats/RTCRtpReceiver.getStats sounds like a good idea (when available). I'm confused, the current code expected goog-stats, but the getStats() being called is invoking the spec-compliant one? Were all of these samples broken for all this time?

fippo:
the web-samples were handling multiple variants.
the test was ticking this off as an optional box so the failure went unnoticed

henbos (approved):
Ok!

PR #1042, opened 2018-04-24
bandwidth: use RTCRtpSender.setParameters

fippo:
uses RTCRtpSender.setParameters to change the bandwidth. Only supported in Chrome 68+

For QA purposes this also allows setting the maximum bandwidth in the JS console before starting the call.  This allows testing that setParameters is operating within the envelope of the SDP b= restrictions.
For example you could set `maxBandwidth=750`, make a call and if you set the bandwidth to 1000 in the dropdown it should still only go to 750kbps.

### DIFF
--- a/src/content/peerconnection/bandwidth/js/main.js
+++ b/src/content/peerconnection/bandwidth/js/main.js
@@ -22,6 +22,10 @@ let pc1;
 let pc2;
 let localStream;
 
+// Can be set in the console before making a call to test this keeps
+// within the envelope set by the SDP. In kbps.
+let maxBandwidth = 0;
+
 let bitrateGraph;
 let bitrateSeries;
 
@@ -98,12 +102,16 @@ function gotDescription2(desc) {
   pc2.setLocalDescription(desc).then(
     () => {
       trace('Answer from pc2 \n' + desc.sdp);
-      pc1
-        .setRemoteDescription({
+      let p;
+      if (maxBandwidth) {
+        p = pc1.setRemoteDescription({
           type: desc.type,
-          sdp: updateBandwidthRestriction(desc.sdp, '500')
-        })
-        .then(() => {}, onSetSessionDescriptionError);
+          sdp: updateBandwidthRestriction(desc.sdp, maxBandwidth)
+        });
+      } else {
+        p = pc1.setRemoteDescription(desc);
+      }
+      p.then(() => {}, onSetSessionDescriptionError);
     },
     onSetSessionDescriptionError
   );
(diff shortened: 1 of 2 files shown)


fippo:
:clap: @Orphis (and you are now volunteered for review and probably a session with @rantonysamy about how to test this :-))

alvestrand (commented):
Nope, I don't think we're there yet.
You're changing the logic to be "set the bandwidth after connecting" on default, with a console-only option to set the session bandwidth.
I'd rather you add two dropdowns, one for "Negotiated session bandwidth" (causes renegotiation when set after session start) and one for "sender's bandwidth" (causes setParameters).

fippo:
this will complicate things a bit since it will not work in Firefox. At the same time I want this to be QA-ready so things like "the bandwidth stays within the negotiated envelope can be tested.

alvestrand:
What I'd do is to enable the negotiate-time bandwidth switcher always and disable the runtime bandwidth switcher if setParameters isn't available. What I don't like is that we change something from working to non-working on platforms that support negotiate-time bandwidth but not runtime bandwidth.

fippo:
On Firefox (or anything that does not have setParameters; Firefox has but they can't update on-the-fly) this will fall-through to using the negotiation-way of doing this. See the last  chunk after expanding the context a bit

src/content/peerconnection/bandwidth/js/main.js
```diff
@@ -185,6 +195,26 @@ bandwidthSelector.onchange = function() {
   bandwidthSelector.disabled = true;
   var bandwidth = bandwidthSelector.options[bandwidthSelector.selectedIndex]
       .value;
+
+  if (adapter.browserDetails.browser === 'chrome' &&
```
fippo: should probably add a comment here that explains what this is doing and why its chrome-only

fippo:
updated comments and commit message to clear things up!

benbro:
Maybe on Firefox we can use setParameters and force renegotiation?
https://bugzilla.mozilla.org/show_bug.cgi?id=1253499
> Parameters set with setParameters currently only take if set before negotiation happens. It should work to change settings mid-call as well, without renegotiation.

fippo:
@benbro that seems like the worst combination :-)

@pehrsons wdyt?

Pehrsons:
Either way seems all right to me. What's the concern with renegotiation? Does the sample talk to any remotes that require renegotiation interop?

fippo:
@Pehrsons its mostly a "how many hacks do we want to show" :-)

setParameters will just require renegotiation but you don't need to change the bandwidth in the SDP, right?

Pehrsons:
Right. You shouldn't have to touch the sdp for this to work.

ErikHellman:
@fippo Rebase? :)

fippo:
rebased!

ErikHellman (approved):

PR #1050, opened 2018-05-23
use RTCRtpSender.getStats

fippo:
'cause that is more awesome


henbos (approved):

fippo:
meh. This is going to break Edge (shim) since that only fixes outboundrtp to outbound-rtp on RTCPeerConnection.getStats.

Break away, this increases my motivation to fix :-)

fippo:
Dang. This broke in Safari which might mean we have to shim sender.getStats there too.

PR #1053, opened 2018-06-08
Convert samples code to ES6

ErikHellman:
**Description**
A first step in converting the samples code to ES6.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -13,7 +13,7 @@
             2
         ],
         "key-spacing": [2, {"beforeColon": false, "afterColon": true}],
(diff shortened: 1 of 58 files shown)


ErikHellman:
The tests became flaky with these changes. 

Looking into that.

samdutton:
Wow! Nice work — I'll take a look.

fippo:
I think this is not flakyness. I've seen quite some issues recently when trying to access ES6 let/const variables from a selenium executeScript. Basically it looks like the geckodriver scope can not access those variables. The console scope can.

This is specific to geckodriver, chromedriver has no such issues.

ErikHellman:
closing this and will make this changes incrementally instead.

PR #1055, opened 2018-07-06
Refactor gum sample to ES6 code

ErikHellman:
**Description**
Refactor the gum sample to ES6.

**Purpose**
This is a small change to see that the tests still work and Travis accepts the changes. The following refactorings will be larger.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -35,7 +35,7 @@
     ],
     "max-len": [
       2,
-      80,
+      120,
       2,
       {
         "ignoreUrls": true
--- a/src/content/getusermedia/gum/js/main.js
+++ b/src/content/getusermedia/gum/js/main.js
@@ -5,47 +5,44 @@
  *  that can be found in the LICENSE file in the root of the source
  *  tree.
  */
-
 'use strict';
 
-var errorElement = document.querySelector('#errorMsg');
-var video = document.querySelector('video');
-
 // Put variables in global scope to make them available to the browser console.
-var constraints = window.constraints = {
+const constraints = window.constraints = {
   audio: false,
   video: true
 };
 
 function handleSuccess(stream) {
-  var videoTracks = stream.getVideoTracks();
+  const video = document.querySelector('video');
+  const videoTracks = stream.getVideoTracks();
   console.log('Got stream with constraints:', constraints);
-  console.log('Using video device: ' + videoTracks[0].label);
-  stream.oninactive = function() {
-    console.log('Stream inactive');
-  };
+  console.log(`Using video device: ${videoTracks[0].label}`);
   window.stream = stream; // make variable available to browser console
   video.srcObject = stream;
 }
 
 function handleError(error) {
   if (error.name === 'ConstraintNotSatisfiedError') {
-    errorMsg('The resolution ' + constraints.video.width.exact + 'x' +
-        constraints.video.width.exact + ' px is not supported by your device.');
+    let v = constraints.video;
+    errorMsg(`The resolution ${v.width.exact}x${v.height.exact} px is not supported by your device.`);
   } else if (error.name === 'PermissionDeniedError') {
     errorMsg('Permissions have not been granted to use your camera and ' +
       'microphone, you need to allow the page access to your devices in ' +
       'order for the demo to work.');
   }
-  errorMsg('getUserMedia error: ' + error.name, error);
+  errorMsg(`getUserMedia error: ${error.name}`, error);
 }
 
 function errorMsg(msg, error) {
-  errorElement.innerHTML += '<p>' + msg + '</p>';
+  const errorElement = document.querySelector('#errorMsg');
+  errorElement.innerHTML += `<p>${msg}</p>`;
   if (typeof error !== 'undefined') {
     console.error(error);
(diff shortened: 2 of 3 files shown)


.eslintrc
```diff
@@ -35,7 +35,7 @@
     ],
     "max-len": [
       2,
-      80,
+      120,
```
henbos: What's this? Line width?

src/content/getusermedia/gum/js/test.js
```diff
@@ -5,98 +5,80 @@
...
+const webdriver = require('selenium-webdriver');
+const seleniumHelpers = require('webrtc-utilities').seleniumLib;
 
-var webdriver = require('selenium-webdriver');
-var seleniumHelpers = require('webrtc-utilities').seleniumLib;
+const path = '/src/content/getusermedia/gum/index.html';
+const url = (process.env.BASEURL ? process.env.BASEURL : ('file://' + process.cwd())) + path;
 
-test('Video width and video height are set on GUM sample', function(t) {
+test('Video width and video height are set on GUM sample', t => {
```
henbos, line 21: Does test() or equivalent allow for "async t = { ... }" instead of "t => { ... }"?

src/content/getusermedia/gum/js/test.js
```diff
@@ -5,98 +5,80 @@
...
 
-  driver.get((process.env.BASEURL ? process.env.BASEURL :
-      ('file://' + process.cwd())) +
-      '/src/content/getusermedia/gum/index.html')
-  .then(function() {
-    t.pass('Page loaded');
-  })
-  .then(function() {
-    return driver.wait(function() {
+  driver.get(url)
```
henbos, line 25: ^--- ... if so, here and equivalent elsewhere:
await driver.get(url);

src/content/getusermedia/gum/js/test.js
```diff
@@ -5,98 +5,80 @@
...
-      videoElement.getAttribute('videoHeight').then(function(h) {
-        height = h;
-        t.pass('Got videoHeight ' + h);
-        if (width && height) {
-          resolve([width, height]);
-        }
+    }, 30 * 1000))
+    // Check that there is a stream with a video track.
+    .then(() => driver.executeScript('return stream && stream.getTracks().length'))
+    .then(numberOfStreams => t.ok(numberOfStreams === 1, 'Stream exists and has one track'))
```
henbos, line 32: Drive-by comment about code, unrelated to refactoring: There is no need to verify the number of tracks in a stream. If GUM() was not rejected and we have a stream, it will have a track. Let's not unittest that the browser is not buggy unless there are known issues (there aren't).

src/content/getusermedia/gum/js/main.js
```diff
@@ -5,47 +5,44 @@
...
+  const errorElement = document.querySelector('#errorMsg');
+  errorElement.innerHTML += `<p>${msg}</p>`;
   if (typeof error !== 'undefined') {
     console.error(error);
   }
 }
 
-navigator.mediaDevices.getUserMedia(constraints).
-    then(handleSuccess).catch(handleError);
+navigator.mediaDevices
```
henbos, line 45: Can we async/await and put the success/error code inline instead of use-once helper functions?

```
(async () => {
  try {
    const stream = await navigator.mediaDevices.getUserMedia(constraints);
    ... do whatever handleSuccess(stream) does ...
  } catch (e) {
    ... do whatever handleError(e); does ...
  }
})();
```

henbos (commented):
Can we async/await? If so let me review again after changes, if there is a good reason not to or you want to do that separately I can approve the change as-is.

.eslintrc
ErikHellman: Yes. I changed this as it would otherwise make a lot of the code look really ugly with unnecessary line breaks.

I believe we should remove all the local rules here and use JavaScript standard or Google code-style directly, but that would be a different PR.

src/content/getusermedia/gum/js/test.js
ErikHellman: Yes. I plan to migrate to async/await in a later PR. Didn't want to make too much changes in this PR.

src/content/getusermedia/gum/js/test.js
ErikHellman: Yes, I have some opinions about the quality of the tests in general. Some of them are not relevant and other are testing the "wrong" things. Will fix in later PRs.

ErikHellman:
The comments about async/await are relevant, but I'd like to make those changes later as it would mean much larger refactoring, as well as some heavy restructuring of the code. 

I plan to replace the custom selenium driver with something more up-to-date (also so we can show that testing WebRTC apps works with standard test runners).

src/content/getusermedia/gum/js/test.js
henbos: Ok sgtm!

src/content/getusermedia/gum/js/test.js
henbos: Cool

henbos (approved):
Ship it!

PR #1062, opened 2018-07-09
Refactor peerconnection samples to ES6

ErikHellman:
**Description**
Refactor peerconnection samples to ES6


henbos (approved):
I scrolled through pretty quickly due to the size of this PR and didn't see any surprises. If you did any non-style changes that I need to take a closer look at that I might have missed let me know, otherwise LGTM!

samdutton:
I also had a quick look and LGTM aside from Travis errors.

ErikHellman:
I had to disable a few tests due to them being flaky (mostly in Firefox). These tests should be re-written properly and not rely on `driver.executeScript()`. This is the goal with another change I'm working on where I'll update the testing dependencies as well.

fippo:
they are not flaky, they try to access a variable declared as const in a different context which for some reason behaves different in geckodriver than in chrome. The best workaround is probably to expose the variables on window.

ErikHellman:
@fippo Good catch. I still think the tests should be refactored to use something more modern (see my webdriver.io + mocha PR). It makes more sense to use something common in samples like this and it will reduce our need for custom solutions (our current testrunner).

PR #1063, opened 2018-07-11
Migrate to webdriverio tests

ErikHellman:
**Description**
This is to test how Webdriver.IO works with Travis. The purpose here is to get rid of our custom test runner and use something that is more common. 

Currently, it only contains a single test and only tests on Chrome. I'll add more tests and browsers as soon as I see that this will pass on Travis.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,7 @@ language: node_js
 cache:
  - node_modules
 node_js:
-  - 6
+  - "lts/*"
 
 branches:
   only:
@@ -24,16 +24,16 @@ env:
     - BROWSER=chrome  BVER=stable
     - BROWSER=chrome  BVER=beta
     - BROWSER=chrome  BVER=unstable
-    - BROWSER=firefox BVER=stable
-    - BROWSER=firefox BVER=beta
-    - BROWSER=firefox BVER=nightly
+#    - BROWSER=firefox BVER=stable
+#    - BROWSER=firefox BVER=beta
+#    - BROWSER=firefox BVER=nightly
 
 matrix:
   fast_finish: true
 
   allow_failures:
     - env: BROWSER=chrome  BVER=unstable
-    - env: BROWSER=firefox BVER=nightly
+#    - env: BROWSER=firefox BVER=nightly
 
 before_install:
   - /sbin/start-stop-daemon --start --quiet --pidfile /tmp/cucumber_xvfb_99.pid --make-pidfile --background --exec /usr/bin/Xvfb -- :99 -ac -screen 0 1600x1400x16
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,5 @@
-dist: trusty
+#dist: trusty
+os: osx
 language: node_js
 cache:
  - node_modules
@@ -11,18 +12,19 @@ branches:
 
 addons:
   chrome: stable
-  apt:
-    packages:
-      - gcc
-      - g++
-      - libstdc++6
-      - pulseaudio
+  firefox: latest
+#  apt:
+#    packages:
+#      - gcc
+#      - g++
+#      - libstdc++6
+#      - pulseaudio
 
-env:
-  global:
-    - DISPLAY=:99.0
-  matrix:
-    - BROWSER=chrome  BVER=stable
+#env:
+#  global:
+#    - DISPLAY=:99.0
+#  matrix:
+#    - BROWSER=chrome  BVER=stable
 #    - BROWSER=chrome  BVER=beta
 #    - BROWSER=chrome  BVER=unstable
 #    - BROWSER=firefox BVER=stable
@@ -36,11 +38,11 @@ matrix:
 #    - env: BROWSER=chrome  BVER=unstable
 #    - env: BROWSER=firefox BVER=nightly
 
-before_install:
-  - /sbin/start-stop-daemon --start --quiet --pidfile /tmp/cucumber_xvfb_99.pid --make-pidfile --background --exec /usr/bin/Xvfb -- :99 -ac -screen 0 1600x1400x16
-
-before_script:
-  - pulseaudio --start
+#before_install:
+#  - /sbin/start-stop-daemon --start --quiet --pidfile /tmp/cucumber_xvfb_99.pid --make-pidfile --background --exec /usr/bin/Xvfb -- :99 -ac -screen 0 1600x1400x16
+#
+#before_script:
+#  - pulseaudio --start
 
 script:
   - npm test
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,6 +10,7 @@ branches:
     - gh-pages
 
 addons:
+  chrome: stable
   apt:
     packages:
       - gcc
@@ -22,8 +23,8 @@ env:
     - DISPLAY=:99.0
   matrix:
     - BROWSER=chrome  BVER=stable
-    - BROWSER=chrome  BVER=beta
-    - BROWSER=chrome  BVER=unstable
+#    - BROWSER=chrome  BVER=beta
+#    - BROWSER=chrome  BVER=unstable
 #    - BROWSER=firefox BVER=stable
 #    - BROWSER=firefox BVER=beta
 #    - BROWSER=firefox BVER=nightly
@@ -31,8 +32,8 @@ env:
 matrix:
   fast_finish: true
 
-  allow_failures:
-    - env: BROWSER=chrome  BVER=unstable
+#  allow_failures:
+#    - env: BROWSER=chrome  BVER=unstable
 #    - env: BROWSER=firefox BVER=nightly
 
 before_install:
--- a/src/content/getusermedia/resolution/js/resolution.spec.js
+++ b/src/content/getusermedia/resolution/js/resolution.spec.js
@@ -2,8 +2,8 @@ import ResolutionPage from './resolution.page';
 import assert from 'assert';
 
 describe('getUserMedia resolution', () => {
-  ResolutionPage.open();
   it('should be 320 wide at qvga', () => {
+    ResolutionPage.open();
     ResolutionPage.qvgaButton.click();
     browser.waitForVisible('#gum-res-local');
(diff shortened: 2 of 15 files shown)


ErikHellman:
Ping @samdutton and @henbos 

Let me know what you think of these new tests using WebDriver.IO. The purpose is to use something more common (WebDriver.IO seems to be the most commonly used now) and also make this run on more platforms (this now runs on Linux and macOS, previously the tests only could run on Linux).

wdio.conf.js
```diff
@@ -0,0 +1,272 @@
...
+  // https://docs.saucelabs.com/reference/platforms-configurator
+  //
+  capabilities: [
+    {
+      browserName: 'chrome',
+      chromeOptions: {
+        args: [
+          '--headless',
+          '--allow-file-access-from-files',
+          '--disable-gesture-requirement-for-media-playback',
```
fippo, line 48: this isn't listed in https://peter.sh/experiments/chromium-command-line-switches/
Did you mean --autoplay-policy=no-user-gesture-required?

.travis.yml
```diff
@@ -1,45 +1,48 @@
...
-  global:
-    - DISPLAY=:99.0
-  matrix:
-    - BROWSER=chrome  BVER=stable
-    - BROWSER=chrome  BVER=beta
-    - BROWSER=chrome  BVER=unstable
-    - BROWSER=firefox BVER=stable
-    - BROWSER=firefox BVER=beta
-    - BROWSER=firefox BVER=nightly
+  chrome: stable
```
fippo, line 14: you might want to check if wdio supports using a specific binary similar to how karma works with CHROME_BIN/FIREFOX_BIN (see [here](https://github.com/webrtc/adapter/blob/master/test/karma.conf.js)). Then integrating this with travis-multirunner would be much easier.

src/content/getusermedia/resolution/js/resolution.spec.js
```diff
@@ -0,0 +1,32 @@
+import ResolutionPage from './resolution.page';
+import assert from 'assert';
+
+describe('getUserMedia resolution', () => {
+  it('should be 320 wide at qvga', () => {
+    ResolutionPage.open();
+    ResolutionPage.qvgaButton.click();
+    browser.waitForVisible('#gum-res-local');
+    const videoLocal = browser.element('#gum-res-local');
+    const width = browser.getAttribute('#gum-res-local', 'videoWidth');
```
fippo, line 10: this is potentially flaky. Maybe not so in practice but i've seen too many flaky tests because of this.
What happens if GUM takes longer and videoWidth is still 0? Might be possible to test by omitting the use-fake-ui-for-media-stream flag

src/js/page.js
```diff
@@ -0,0 +1,9 @@
+export default class Page {
```
fippo, line 1: this should go into test/?

fippo (commented):
if you go by popularity, protractor is much more popular than wdio. I like neither but that is just me.

You are trying a bunch of things at the same time.
1) switch the test runner

The samples are currently running a travis matrix of six browsers/versions (and I wonder why we no longer run ESR... probably because of marionette/selenium woes) and i've also been using the tests in Edge as a post-release test for adapter

Running on mac is not that big of an issue, its mostly that nobody ever volunteered or complained enough.

2) switching the test framework to mocha

I <3 mocha but I don't see many use-cases for a test suite, the resolution test is one of the few that tests more than a single case.

3) rewriting the tests using page objects

which is generally fine but I think its overkill given the low complexity of the samples.
What do your QA folks think -- I always assumed I could get them to write some tests to automate their tasks but that never happened.

 4) adding html elements to avoid executeScript?

there is too much stuff in webrtc that is only testable with executeScript.

ErikHellman:
All good feedback. I choose webdriver.io after I tested a bunch of other tools. None is great, but this one had the least issues setting up.

I agree that some things are a bit overkill, but my idea here is to have complete samples. People asking "How do I test my app that uses WebRTC?" can look at these. This is also the main reason for switching the testrunner. If the samples uses a custom runner, it would deter people from properly testing their WebRTC code. 

Regarding nr. 4: I agree that testing WebRTC by itself would need `executeScript()`, but since all the tests are UI tests it makes more sense to have samples that can be fully tested as UI tests. These should after all be samples that shows how to do stuff in real world.

src/js/page.js
ErikHellman: Sound good.

fippo:
this is getting a bit meta so... grab a :coffee: or :tea: 

> I agree that some things are a bit overkill, but my idea here is to have complete samples.

A bit of history. The samples started in the Chrome tree, along with adapter and apprtc.
This was in the early days of WebRTC (2012) and it was a green field. Thanks to the early
adapter.js shimming the prefix differences they worked in both Chrome and Firefox without much effort.

Stuff was moved to github in 2014, first to the googlechrome organisation, then to the webrtc
organisation. @samdutton explains this [here](https://www.youtube.com/watch?time_continue=5&v=DvzDzIXoncg). Good old times.

In 2015, Edge shipped getUserMedia. This didn't work out of the box as Edge was supportіng srcObject
on the video element to attach a stream to it. Now we had getUserMedia
[working in three browsers](https://blog.andyet.com/2015/06/25/webrtc-on-the-edge) so yay convergence. Shimming RTCPeerConnection followed a bit later and the samples defined "good enough".

The purpose of tests in this repository has not been "show people how to test webrtc" so far.
See [here](https://blog.andyet.com/2015/07/28/selenium-travis-webrtc) and in particular #564. Given that @kaptenjansson is a QA guy the samples also serve as a baseline for what QA (:wave: @rantonysamy) is testing manually because automatic tests are ok but QA folks are much more... creative in finding weird stuff.

Last but not least a big source of inspiration for the samples have been questions asked on discuss-webrtc.
I always enjoyed saying "there is a sample for that". In the last year or two I could do the same on stackoverflow.

Likewise there is the reason (apart from 'please webrtc, understand that you need
developer advocate, this is not my job') the samples have remained written
in ES5 and plain HTML. If you look around on [stackoverflow](https://stackoverflow.com/search?tab=newest&q=webrtc) you will see very little ES6 usage in the questions. Lets give it a try though.

We did use things like promises and addTrack though because we wanted to specifically drive adoption
among WebRTC developers. We have specifically not adopted the style of @jan-ivar's jsfiddles which,
while highly elegant, are much harder to understand for the target audience.


>  If the samples uses a custom runner, it would deter people from properly testing their WebRTC code.

Much of what is used here has its roots in [the rtc.io testing process](http://rtc.io/testing-process.html)
and was written in early 2013 (I think). If the samples uses wdio, this might deter people using protractor, cypress, nightwatch or 'raw' selenium.
Further, using something in the samples might be seen as a recommendation ('Google uses it').
And I do not think there is a broad consensus beyond "we use travis matrix builds" to do so.


Further, the samples, which are intentionally very minimalistic, are not the right place to demonstrate
the testing process either. AppRTC, which is a complete application, is probably a better place to do that.
A good first bug to fix there would be https://github.com/webrtc/apprtc/issues/551 (@henbos can probably help figure out why) and then write a test to ensure this doesn't break again.

ErikHellman:
> this is getting a bit meta so... grab a ☕️ or 🍵

Way ahead of you... ;)

> The purpose of tests in this repository has not been "show people how to test webrtc" so far.

I think this approach should be changed. We shouldn't just show how WebRTC can be used, but also how to write tests for your own use of WebRTC APIs. 

Also, if we're having tests in the samples, those should be up-to-date and use something the rest of the community is familiar with.

You've convinced me that WebDriver.IO might not be the best option here. However, we should find something that doesn't require a custom runner. Protractor seems very angular specific, so I think I want to avoid that for that reason. Any other suggstion?

fippo:
> I think this approach should be changed. We shouldn't just show how WebRTC can be used, but also how to write tests for your own use of WebRTC APIs.

The main problem is that tests are typically tightly integrated into the CI process. And how this integration is done changes the requirements for things like a runner. For example, if you use something like gocd.io (I assume the same is true for the new jenkins dashboard) which allows you to define a task per demo, the "single nodejs script" which tests one sample you don't need the thing we have here as test/run-tests.js. Also it means your tests are much easier to run in parallel on different nodes in your CI cluster. I have a blog post in the queue about that for next month.

Travis-CI or similar cloud-CI services have the advantage that they have a very minimal configuration, but they also dictate a different model -- run all the tests in sequence.

What can be shown is the techniques for testing WebRTC. The main building blocks are
* checking the ice connection state and waiting for connected or completed.
* checking the video element readyState. iceConnectionState is not ѕufficient as it only includes the DTLS layer in Chrome -- its the connectionState. In Firefox it is the actual iceConnectionState. I don't think this is shown anywhere but it will only work if video is actually playing. Alternatively check whether videoWidth/videoHeight are set, see the upgrade test.
* looking at the video element webkitDecodedFrameCount or mozPaintedFrames. Highly useful to verify frames are drawn but prefixed.
* use getStats more to make assertions. The main reason this isn't done is that the tests predate spec-stats, even though the webdriver utils have a helper. Even without an output UI as shown in https://testrtc.com/houseparty-testrtc-testimonial/ just measuring the bitrate and making an assertion is useful. Problem is that the results of these tests are generally more unstable (due to more complexity).
* grab frames via canvas and reason about them. Done in the testrtc repo in this organisation, in particular the videoframechecker. This is useful for testing things like "if you click the video mute button, the other side should receive only black frames".
* Similar things can be done for audio with WebAudio, see the mictest in testrtc.

Most of this exists but is not well documented due to lack of time (whereas 'well-documented' is defined as 'one blog post per technique'). For real-life testing add "how to test that no camera is present" and "simulate a delay between calling getUserMedia and the result and check for a visual hint".

The other important thing is "test in more than a single browser and include beta/canary/nightly versions". We do that already.

> Also, if we're having tests in the samples, those should be up-to-date and use something the rest of the community is familiar with.

I think we can agree that writing selenium tests in java is meh (unless you come from java)

> Any other suggstion?

cypress.io maybe? Maybe @aullman has another suggestion, I'm mostly happy with node scripts -- but then I have a few dozen of them.

Issue #1065, opened 2018-07-12
Peer connection: Audio only sample is broken


rantonysamy:
## Browser affected
Version 68.0.3440.40 (Developer Build) (64-bit)
Also observed in M69 Dev Build

## Description
Peer connection: Audio only sample is broken and no graph been displayed 

## Steps to reproduce

Go to https://webrtc.github.io/samples/src/content/peerconnection/audio/
Select a codec and Click on the Call button
Observe the Bitrate and Packets sent per second graphs 

## Expected results
Packets sent per second and Bitrate Graph is been started and displaying graph

## Actual results
No graph been displayed

PR #1069, opened 2018-07-16
fix audio sample graphs

fippo:
even more fix for #1065 -- also uses Map.has for checking if a thing exists


src/content/peerconnection/audio/js/main.js
```diff
@@ -251,18 +251,19 @@ window.setInterval(() => {
       if (report.type === 'outbound-rtp') {
         bytes = report.bytesSent;
         packets = report.packetsSent;
-        const reportId = lastResult.get(report.id);
```
fippo, line 254: the semantics here was wrong, this is not a "reportId". Since @henbos reviewed and did not catch this he is now volunteered to explain the getStats snowball :-p

Issue #826, opened 2016-09-08
Add error logging to FileReader in file transfer sample


alvestrand:
**Browsers and versions affected**
Possibly Android. See https://crbugs.com/644878

**Description**
When something bad happens to a FileReader in the the "filetransfer" demo, no error will be reported;
the application seems to hang.

**Steps to reproduce**
See crbug above.

**Expected results**
File transfer should either complete or give an error message.

**Actual results**
File transfer does not appear to complete.

PR #1072, opened 2018-07-18
Error handling and refactoring of filetransfer sample

ErikHellman:
- Add error handling to FileReader (Fix for #826)
- Refactor sample and test to ES6 code


ErikHellman:
Ping @alvestrand. This is for issue #826 that you filed. Was this what you had in mind?

src/content/datachannel/filetransfer/js/main.js
```diff
@@ -70,14 +70,12 @@ function createConnection() {
 }
 
 function onCreateSessionDescriptionError(error) {
-  trace('Failed to create session description: ' + error.toString());
+  trace(`Failed to create session description: ${error.toString()}`);
```
kenchris: doesn't it call toString() automatically when embedded in a string?

src/content/datachannel/filetransfer/js/main.js
```diff
@@ -90,19 +88,20 @@ function sendData() {
...
-      return function(e) {
-        sendChannel.send(e.target.result);
-        if (file.size > offset + e.target.result.byteLength) {
-          window.setTimeout(sliceFile, 0, offset + chunkSize);
-        }
-        sendProgress.value = offset + e.target.result.byteLength;
-      };
+  const chunkSize = 16384;
+  const sliceFile = offset => {
+    const reader = new window.FileReader();
```
kenchris: window is global... and you don't override it locally... remove window.

src/content/datachannel/filetransfer/js/main.js
```diff
@@ -90,19 +88,20 @@ function sendData() {
...
+    reader.onerror = (error => {
```
kenchris: why the ( ) ?

src/content/datachannel/filetransfer/js/main.js
```diff
@@ -128,7 +127,8 @@ function closeDataChannels() {
 
 function gotDescription1(desc) {
   localConnection.setLocalDescription(desc);
-  trace('Offer from localConnection \n' + desc.sdp);
+  trace(`Offer from localConnection 
+${desc.sdp}`);
```
kenchris: why this change?

src/content/datachannel/filetransfer/js/main.js
```diff
@@ -148,29 +149,30 @@ function getOtherPc(pc) {
...
-      onAddIceCandidateSuccess(pc);
-    },
-    function(err) {
-      onAddIceCandidateError(pc, err);
-    }
-  );
-  trace(getName(pc) + ' ICE candidate: \n' + (event.candidate ?
-      event.candidate.candidate : '(null)'));
+    .then(
+      () => {
```
kenchris: or just `_ => onAddIceCandidateSuccess(pc);`

src/content/datachannel/filetransfer/js/main.js
```diff
@@ -200,21 +202,20 @@ function onReceiveMessageCallback(event) {
 
   // we are assuming that our signaling protocol told
   // about the expected file size (and name, hash, etc).
-  var file = fileInput.files[0];
+  const file = fileInput.files[0];
   if (receivedSize === file.size) {
-    var received = new window.Blob(receiveBuffer);
+    const received = new window.Blob(receiveBuffer);
```
kenchris: drop window.

src/content/datachannel/filetransfer/js/main.js
```diff
@@ -1,3 +1,4 @@
+/* eslint no-unused-expressions: 0 */
```
alvestrand, line 1: Nit: Normal instructions say "copyright at top". Is it OK to put eslint stuff after the copyright header?

src/content/datachannel/filetransfer/js/main.js
```diff
@@ -7,77 +8,76 @@
...
 
 fileInput.addEventListener('change', handleFileInputChange, false);
 
-function handleFileInputChange() {
-  var file = fileInput.files[0];
+async function handleFileInputChange() {
+  let file = fileInput.files[0];
   if (!file) {
-    trace('No file chosen');
+    console.log('No file chosen');
```
alvestrand: Why are you changing from trace to console.log?
I see that trace added timestamps. Timestamps can be useful in debugging. What's the argument for not providing them?

src/content/datachannel/filetransfer/js/main.js
```diff
@@ -90,91 +90,84 @@ function sendData() {
...
-
-function onAddIceCandidateError(error) {
-  trace('Failed to add Ice Candidate: ' + error.toString());
+async function onIceCandidate(pc, event) {
+  try {
+    const candidateText = event.candidate ? event.candidate.candidate : '(null)';
+    console.log(`${getName(pc)} ICE candidate: ${candidateText}`);
+    await getOtherPc(pc).addIceCandidate(event.candidate);
+    console.log('AddIceCandidate success.');
+  } catch (e) {
```
alvestrand: This try/catch will eat the error, so higher level functions will not see failure. Is that OK?

src/content/datachannel/filetransfer/js/main.js
```diff
@@ -226,71 +219,48 @@ function onReceiveMessageCallback(event) {
...
   }
 }
 
 // display bitrate statistics.
-function displayStats() {
-  var display = function(bitrate) {
-    bitrateDiv.innerHTML = '<strong>Current Bitrate:</strong> ' +
-        bitrate + ' kbits/sec';
+async function displayStats() {
+  const display = bitrate => {
```
alvestrand: having this as a function that is redefined every time you execute displayStats() was silly when I wrote it. Can you make it better?

src/content/datachannel/filetransfer/js/test.js
```diff
@@ -5,98 +5,92 @@
...
-      }, 2000);
-    }
-    // Wait for the download element to be displayed.
-    return driver.wait(webdriver.until.elementIsVisible(
+  const path = '/src/content/datachannel/filetransfer/index.html';
+  const url = `${process.env.BASEURL ? process.env.BASEURL : ('file://' + process.cwd())}${path}`;
+
+  return driver
+    .get(url)
+    .then(() => {
```
alvestrand, line 28: Is there a reason for not using await in this .then-chain?

alvestrand (commented):
Chief question is why you want to drop the timestamps in the console log, I think.
Rest are nits.

src/content/datachannel/filetransfer/js/main.js
ErikHellman: I'll try. I need to bump eslint tool as this is only needed because of a bug, but that requires rewriting the entire .eslinntrc. :/

src/content/datachannel/filetransfer/js/main.js
ErikHellman: While timestamps might be useful, I believe it is more useful to be able to track the line in the js file where this was printed. trace() is defined in a different file, so all these calls will look like it came from there.

Between timestamp and finding where the log was printed, I'd choose the later.

src/content/datachannel/filetransfer/js/main.js
ErikHellman: Can the caller do anything if this fails? If not, I think we're good here. 

Maybe we'd like to print the error in the HTML as well?

src/content/datachannel/filetransfer/js/test.js
ErikHellman: I'm waiting with updating the tests until we've decided on what we should use. 
This is just updating var to let/const (simple refactoring using WebStorm).

src/content/datachannel/filetransfer/js/main.js
alvestrand: Can we have our cake and eat it too?
ie change console.log(msg) to console.log(ts(), msg), where ts() returns a properly formatted timestamp?

src/content/datachannel/filetransfer/js/main.js
alvestrand: I'm mostly concerned with functions that expect to continue business-as-usual. When things go sideways, we want to stop and hoist the "read the logs, please!" banner, not continue waiting for things to clear up by themselves (which they won't).

src/content/datachannel/filetransfer/js/main.js
alvestrand: Probably worth leaving for another CL :-)

src/content/datachannel/filetransfer/js/main.js
```diff
@@ -61,16 +61,16 @@ async function createConnection() {
   };
   remoteConnection.ondatachannel = receiveChannelCallback;
 
-  await generateOffer(localConnection);
+  try {
+    const offer = await localConnection.createOffer();
+    await gotLocalDescription(offer);
+  } catch (e) {
+    onCreateSessionDescriptionError(e);
```
alvestrand: the whole on...Error thing is from when createOffer() etc took callbacks. Can you just inline the log line?

src/content/datachannel/filetransfer/js/main.js
alvestrand: onIceCandidate is called twice, both by functions that are event handlers, and the only point of having the function is to get the reference to the pc that sent the event, which seems like the event should have info about (does it?)

Can we rewrite this function so that it can be directly assigned as the event handler?

It should call out to some global "something went wrong" routine that splashes a banner on top of the page and stops trying to do more things until user intervenes.

alvestrand (commented):
Getting close.

src/content/datachannel/filetransfer/js/main.js
ErikHellman: That would make things like; `console.log('Got this thing: ', anObject)` difficult to do. 

We could redefine `console.log`, but IMO the timestamp is not really needed. Everything happens on a single thread anyway, so it will be logged in chronological order. Not sure what the timestamp actually helps with in this context.

src/content/datachannel/filetransfer/js/main.js
ErikHellman: How about having a `<div id="errorMessages"></div>` where we append errors? We use a similar (but simpler) thing in other samples. Maybe expand on that? 

I agree about the problems with `onIceCandidate` but I choose not to change that now. Let me investigate if we can get the pc from the event and maybe eliminate this function as well.

alvestrand (approved):

Issue #959, opened 2017-11-30
Adding Local Cache to Store the Data


rantonysamy:
**Browsers and versions affected**
All

**Description**
Adding Local Cache to Store the Data, So while connecting again, I do not need to enter the data  (Even better, make a AppRTC Call and find out the Credentials and list that down) 

https://webrtc.github.io/samples/src/content/peerconnection/trickle-ice/

KaptenJansson:
We should not do automatic fetching but using local storage is fine for storing submitted servers. Feel free to create a PR.

PR #1073, opened 2018-07-18
Stores the ICE servers configuration in local storage (see #959).

ErikHellman:
**Description**
Storing the configured ICE servers in localStorage to make testing easier. 
Added a reset button to revert to the original configuration.

See #959

### DIFF
--- a/src/content/peerconnection/trickle-ice/js/main.js
+++ b/src/content/peerconnection/trickle-ice/js/main.js
@@ -13,6 +13,7 @@ const candidateTBody = document.querySelector('tbody#candidatesBody');
 const gatherButton = document.querySelector('button#gather');
 const passwordInput = document.querySelector('input#password');
 const removeButton = document.querySelector('button#remove');
+const resetButton = document.querySelector('button#reset');
 const servers = document.querySelector('select#servers');
 const urlInput = document.querySelector('input#url');
 const usernameInput = document.querySelector('input#username');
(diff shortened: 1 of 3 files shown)


src/content/peerconnection/trickle-ice/js/main.js
```diff
@@ -32,6 +41,35 @@ let begin;
...
+
+function readServersFromLocalStorage() {
+  document.querySelectorAll('select#servers option').forEach(option => option.remove());
+  const serversSelect = document.querySelector('select#servers');
+  const storedServers = window.localStorage.getItem(allServersKey);
+
+  if (storedServers === null || storedServers === '') {
+    const o = document.createElement('option');
+    o.value = '{"urls":["stun:stun.l.google.com:19302"]}';
+    o.text = 'stun:stun.l.google.com:19302';
```
alvestrand: You don't want all these specific values in two places. Make a function or a constant for "setDefaultServers" and call it from both places.

src/content/peerconnection/trickle-ice/js/main.js
```diff
@@ -32,6 +41,35 @@ let begin;
...
+    serversSelect.add(o);
+  } else {
+    JSON.parse(storedServers).forEach((server, key) => {
+      const o = document.createElement('option');
+      o.value = JSON.stringify(server);
+      o.text = server.urls[0];
+      o.ondblclick = selectServer;
```
alvestrand, line 68: Not in this CL (it was there already), but how discoverable is a dobule-click interface?
What's it supposed to do?

alvestrand (approved):
Looks good, with one nit.

src/content/peerconnection/trickle-ice/js/main.js
ErikHellman: Agree. It seems it was added to quickly add new servers. I'll remove this for now as it is just confusing.

PR #1074, opened 2018-07-18
Fix for Web Audio input not working in Safari.

ErikHellman:
See #925

### DIFF
--- a/src/content/peerconnection/webaudio-input/js/main.js
+++ b/src/content/peerconnection/webaudio-input/js/main.js
@@ -67,10 +67,10 @@ function handleSuccess(stream) {
     console.log('Got one audio track:', audioTracks);
     const filteredStream = webAudio.applyFilter(stream);
     const servers = null;
-    pc1 = new webkitRTCPeerConnection(servers); // eslint-disable-line new-cap
+    pc1 = new RTCPeerConnection(servers); // eslint-disable-line new-cap
     console.log('Created local peer connection object pc1');
     pc1.onicecandidate = e => onIceCandidate(pc1, e);
-    pc2 = new webkitRTCPeerConnection(servers); // eslint-disable-line new-cap
+    pc2 = new RTCPeerConnection(servers); // eslint-disable-line new-cap
     console.log('Created remote peer connection object pc2');
     pc2.onicecandidate = e => onIceCandidate(pc2, e);
     pc2.ontrack = gotRemoteStream;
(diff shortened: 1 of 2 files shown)


src/content/peerconnection/webaudio-input/js/main.js
```diff
@@ -67,10 +67,10 @@ function handleSuccess(stream) {
     console.log('Got one audio track:', audioTracks);
     const filteredStream = webAudio.applyFilter(stream);
     const servers = null;
-    pc1 = new webkitRTCPeerConnection(servers); // eslint-disable-line new-cap
+    pc1 = new RTCPeerConnection(servers); // eslint-disable-line new-cap
```
alvestrand: Since you're now using a caps constructor, I think you can remove the eslint thingy.

src/content/peerconnection/webaudio-input/js/main.js
```diff
@@ -85,15 +85,15 @@ function handleSuccess(stream) {
 
     localStream = stream;
   } else {
-    alert('The media stream contains an invalid number of audio tracks.');
+    console.log('The media stream contains an invalid number of audio tracks.');
```
alvestrand: console.log is totally invisible to normal users of a sample page. Do you really want to switch from alert() to console.log()?

alvestrand (commented):
How does this fix audio input not working in safari?
the changes seem irrelevant to me.

src/content/peerconnection/webaudio-input/js/main.js
ErikHellman: `alert()` is really annoying and we shouldn't use it for the samples. It doesn't help with understanding. Our samples refer to the JS console in many other cases and I vote to move all logging and error message there. Also printing error messages in a separate `<div>` would work. I'll add that as well.

ErikHellman:
So `webkitRTCPeerConnection` causes a `ReferenceError` in Safari and Firefox, but it works in Chrome. I changed this and now it works in both. :)

src/content/peerconnection/webaudio-input/js/main.js
fippo: this code is probably cruft handling Chrome potentially returning a stream without audio tracks when being asked for {audio:true, video: true}) and there is no microphone. And this would not even have worked because when just asking for audio and there is no microphone an error would have been thrown.

away with it!

henbos (approved):

PR #1077, opened 2018-07-20
Adds a new sample for datachannel

ErikHellman:
- Messaging between two peers
- Uses new ES6 module importing
- Uses lit-element for web components
- async/await instead of promises

If this looks ok, I'd like to move on to refactor the other samples in a similar style.


ErikHellman:
Ping @fippo. I couldn't add you as a reviewer, but you might be interested in this as well.

src/content/datachannel/messaging/index.html
```diff
@@ -0,0 +1,68 @@
+<!DOCTYPE html>
+<!--
+ *  Copyright (c) 2015 The WebRTC project authors. All Rights Reserved.
```
fippo: masquerarding as jansson (who is ooo currently): wrong year, also below

src/content/datachannel/messaging/main.js
```diff
@@ -0,0 +1,168 @@
...
+    this._root.querySelector('#disconnectButton').disabled = true;
+    this._localConnection.close();
+    this._remoteConnection.close();
+  }
+
+  async connect(e) {
+    try {
+      this._root.querySelector('#connectButton').disabled = true;
+      const dataChannelParams = {ordered: false};
+      let lpc = this._localConnection = new RTCPeerConnection(null);
```
fippo: new RTCPeerConnection()
-- passing null was just to workaround Chrome being picky about the lack of arguments in Chrome < 51

src/content/datachannel/messaging/main.js
```diff
@@ -0,0 +1,168 @@
...
+  disconnect(e) {
+    this._root.querySelector('#disconnectButton').disabled = true;
+    this._localConnection.close();
+    this._remoteConnection.close();
+  }
+
+  async connect(e) {
+    try {
+      this._root.querySelector('#connectButton').disabled = true;
+      const dataChannelParams = {ordered: false};
```
fippo: any reason for using unordered datachannels? It will be somewhat hard to show out-of-order delivery using this sample

src/content/extensions/multipleroutes/src/utils.js
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable */
```
fippo: drive-by? wfm

src/content/datachannel/messaging/index.html
```diff
@@ -0,0 +1,68 @@
...
+    <title>Send messages with datachannel</title>
+
+    <link rel="icon" sizes="192x192" href="../../../images/webrtc-icon-192x192.png">
+    <link href="//fonts.googleapis.com/css?family=Roboto:300,400,500,700" rel="stylesheet" type="text/css">
+    <link rel="stylesheet" href="../../../css/main.css">
+    <link rel="stylesheet" href="main.css"/>
+
+    <script src="https://unpkg.com/@webcomponents/webcomponentsjs/webcomponents-loader.js"></script>
+    <script src="https://webrtc.github.io/adapter/adapter-latest.js"></script>
+    <script src="main.js" type="module"></script>
```
fippo, line 32: this works in all browsers supported?

src/content/datachannel/messaging/main.js
```diff
@@ -0,0 +1,168 @@
...
+      console.log('addIceCandidate successful!');
+    } catch (e) {
+      console.log(`error on addIceCandidate: ${JSON.stringify(e)}`);
+    }
+  }
+
+  _localChannelOpen(event) {
+    console.log(`Local channel open: ${JSON.stringify(event)}`);
+    this._root.querySelector('#sendLocal').disabled = false;
+    this._root.querySelector('#sendRemote').disabled = false;
```
fippo: this doesn't listen for the remote channel open event. Now its pretty hard to get this to fail in this sample but still wrong semantics

fippo (changes requested):

src/content/datachannel/messaging/main.js
ErikHellman: Copy paste mistake. Thanks for noticing. :)

src/content/datachannel/messaging/index.html
ErikHellman: I've been told it does.

src/content/datachannel/messaging/main.js
ErikHellman: True. I'll add the same to the remote as well for clarity.

src/content/extensions/multipleroutes/src/utils.js
ErikHellman: This one kept failing on a lot of stuff when I enable the newer JS stuff. I'll enable it again once I get a chance to refactor it as well.

src/content/datachannel/messaging/index.html
fippo: i've tested in firefox esr (52) and that works too so.. i'm surprised!

src/content/datachannel/messaging/main.js
```diff
@@ -0,0 +1,169 @@
...
+import {LitElement, html} from 'https://unpkg.com/@polymer/lit-element?module';
+
+class MessagingSample extends LitElement {
+  constructor() {
+    super();
+    this._localConnection = this._remoteConnection = this._localChannel = this._remoteChannel = null;
+  }
+
+  disconnect(e) {
+    this._root.querySelector('#disconnectButton').disabled = true;
```
kenchris: this.shadowRoot

src/content/datachannel/messaging/main.js
```diff
@@ -0,0 +1,169 @@
...
+    this._localConnection.close();
+    this._remoteConnection.close();
+  }
+
+  async connect(e) {
+    try {
+      this._root.querySelector('#connectButton').disabled = true;
+      const dataChannelParams = {ordered: true};
+      let lpc = this._localConnection = new RTCPeerConnection();
+      lpc.addEventListener('icecandidate', (e) => this._onIceCandidate(e, this._remoteConnection));
```
kenchris: only one argument... `e => this...` works fine

src/content/datachannel/messaging/main.js
```diff
@@ -0,0 +1,169 @@
...
+  _localChannelOpen(event) {
+    console.log(`Local channel open: ${JSON.stringify(event)}`);
+    this._root.querySelector('#sendLocal').disabled = false;
+    this._root.querySelector('#sendRemote').disabled = false;
+    this._root.querySelector('#disconnectButton').disabled = false;
+  }
+
+  _channelClosed(event) {
+    console.log(`Channel closed: ${JSON.stringify(event)}`);
+    this._root.querySelector('#sendLocal').disabled = true;
```
kenchris: with lit-html you might want to do these the otehr way around... it have local variable that you change and just make the elements depend on that. ie don't manually modify that dom in multiple goes

ErikHellman:
Ping @samdutton 

What's your opinion on having this format for the samples? My argument for using polymer is that all real web projects uses some kind of framework, and showing WebRTC in relation to that makes it easier to relate to.

src/content/datachannel/messaging/main.js
```diff
@@ -0,0 +1,180 @@
...
+    console.log(`Local message received by remote: ${event.data}`);
+    this.shadowRoot.querySelector('#remoteIncoming').value += event.data + '\n';
+  }
+
+  static get properties() {
+    return {
+      connected: {
+        type: Boolean,
+        notify: true,
+        reflectToAttribute: true
```
kenchris: Why reflecting to attribute?

kenchris: It means that your MessagingSample (ie <message-sample> can have a connected attribute like:)

`<message-sample connected>`

samdutton:
(Hi @ErikHellman — I wrote a reply a couple of days ago, but I had to restart Chrome at some point, and didn't realise my comment hadn't been posted :/.)

> What's your opinion on having this format for the samples? My argument for using polymer is that all real web projects uses some kind of framework, and showing WebRTC in relation to that makes it easier to relate to.

I take you point about making WebRTC easier to relate to, but personally I prefer the barebones approach.

Unless you're demonstrating a complete app (such as appr.tc) I think it's easier to understand what's going on when you can focus on the core code and not be distracted by scaffolding. 

Using a framework also means more to maintain, more to go wrong and more potential for obsolescence. For example the [YouTube Data API samples](https://developers.google.com/youtube/v3/code_samples/javascript) were written using jQuery, which is now generally avoided for new projects. I also suspect that many WebRTC developers may not be familiar with Polymer.

However, if other WebRTC folk feel differently, I'm not *that* bothered :). As long as it's possible to know from the code what's important, I think that's fine.

src/content/datachannel/messaging/main.js
kenchris: Also do you need `notify`? isnt it not only for callbacks? like observing when it changes from the outside

kenchris:
If you look at the patch, I would actually say that the logic is nicely separated from the UI parts - more so that normally and there is very little scaffolding.

If I am to learn WebRTC that seems perfect, as I don't care much about how the UI is updated but more about the Web RTC logic.

This is using lit-html and not really Polymer, so it is basically only used for stamping out the HTML, which would be much more work manually and result in more irrelevant code, IMHO

src/content/datachannel/messaging/main.js
```diff
@@ -0,0 +1,180 @@
...
+            <label for="remoteIncoming">Remote incoming messages:</label>
+            <textarea class="message" id="remoteIncoming" disabled
+                      placeholder="Remote incoming messages arrive here."></textarea>
+        </div>
+
+    </section>`;
+  }
+
+  _firstRendered() {
+    this.shadowRoot.querySelector('#connectButton').addEventListener('click', () => this.connect());
```
kenchris: You can add these in the HTML.

`<button disabled="${connected}" @click={() => this.connect()}>Connect</button>`

kenchris: actually you probably want

`<button ?disabled=${connected} @click={() => this.connect()}>Connect</button>`

src/content/datachannel/messaging/main.js
```diff
@@ -0,0 +1,180 @@
...
+    } catch (e) {
+      console.log(e);
+    }
+  }
+
+  async _onIceCandidate(e, connection) {
+    try {
+      console.log(`onIceCandidate: ${JSON.stringify(e)}`);
+      // eslint-disable-next-line no-unused-vars
+      const result = await connection.addIceCandidate(e.candidate);
```
kenchris: not using result

src/content/datachannel/messaging/main.js
```diff
@@ -0,0 +1,163 @@
...
+
+  disconnect() {
+    this._localConnection.close();
+    this._remoteConnection.close();
+  }
+
+  async connect() {
+    try {
+      const dataChannelParams = {ordered: true};
+      let lpc = this._localConnection = new RTCPeerConnection();
```
alvestrand: style: abbreviations are bad. lpc -> localConnection

src/content/datachannel/messaging/main.js
```diff
@@ -0,0 +1,163 @@
...
+    this.shadowRoot.querySelector('#remoteIncoming').value += event.data + '\n';
+  }
+
+  static get properties() {
+    return {
+      connected: Boolean
+    };
+  }
+
+  _render({connected}) {
```
alvestrand: Here's where I get confused about the point of all this frameworking.
First you use a declarative syntax to draw up the document.
Then you use an imperative syntax to declare an object with properties.
Then you use the imperative syntax to push out declarative syntax that seems to be exactly the same declarative syntax you would write up, but with two layers of extra quoting, and putting your declarative syntax inside the imperatives-syntax file rather than keeping it with the rest of the HTML.
What's the point?

(on a more mundane level: what's the rules for when _render gets called and/or re-called, and who sends parameters to it?)

src/content/extensions/multipleroutes/src/utils.js
```diff
@@ -10,8 +10,8 @@
 
 'use strict';
 
-var pn = chrome.privacy.network;
-var pi = null;
+const pn = chrome.privacy.network;
+const pi = chrome.privacy.IPHandlingPolicy;
```
alvestrand: Don't abbreviate. It obscures.

src/content/extensions/multipleroutes/src/utils.js
```diff
@@ -10,8 +10,8 @@
 
 'use strict';
 
-var pn = chrome.privacy.network;
-var pi = null;
+const pn = chrome.privacy.network;
```
alvestrand: Was this just pulled along? It seems a bit irrelevant to the main point of the example.

src/content/datachannel/messaging/main.js
```diff
@@ -0,0 +1,163 @@
...
+            <label for="remoteIncoming">Remote incoming messages:</label>
+            <textarea class="message" id="remoteIncoming" disabled
+                      placeholder="Remote incoming messages arrive here."></textarea>
+        </div>
+
+    </section>`;
+  }
+
+  _sendLocalMessage() {
+    const message = this.shadowRoot.querySelector('#localOutgoing').value;
```
alvestrand: where did the shadowRoot enter the picture?
Is this too part of the framework?

alvestrand (commented):
As we talked about on VC, I need to understand why this way is considered "better" than the barebones approach. So far, I don't see what's better about it. Can you try to explain that again?

src/content/datachannel/messaging/main.js
fippo: we should try to get some consistency. Other samples as well as lots of wpt use pc1 and pc2

src/content/datachannel/messaging/main.js
ErikHellman: `shadowRoot` is part of the whole Web Components thing. Basically, it's how we access the Shadow DOM.

src/content/extensions/multipleroutes/src/utils.js
ErikHellman: II need to make a separate PR that cleans up that file. Will fix.

src/content/extensions/multipleroutes/src/utils.js
ErikHellman: Agree. :)

src/content/datachannel/messaging/main.js
ErikHellman: I agree. I'll remove the abbreviations here.

src/content/datachannel/messaging/main.js
ErikHellman: Regarding the `_render()` function, the API docs are found here: https://github.com/Polymer/lit-element/blob/master/README.md

About the mix of declarative and imperative, I can understand the confusion. However, this is what a lot of the other frameworks, like React, are doing already, so most web developers will be familiar with it. Even on Android we're doing something similar. For me, the most obvious advantage is data-binding.

ErikHellman: Also, @kenchris can probably give a better answer about the design of this framework. :)

ErikHellman:
The framework I've used here is basically just a thin wrapper on top of the standard Web Components APIs (templates, Shadow DOM and custom elements). An example of how it would look like without lit-elements and only using these standard APIs can be seen here (works only in Chrome at the moment); https://codepen.io/ErikHellman/pen/BPVaZX

Since this is standard APIs that are coming to all browsers, my opinion is that samples for web APIs should embrace this way of creating UI as well. Nobody builds a real application the way the samples are designed today (barebone), and it serves no purpose in samples as they lack context. 

From a developer relations perspective this makes more sense. Samples should be contextual, or lots of developers will have a hard time understanding how to use the APIs in their projects.

src/content/datachannel/messaging/main.js
```diff
@@ -0,0 +1,164 @@
...
+ */
+/* eslint no-unused-expressions: 0 */
+
+'use strict';
+import {LitElement, html} from 'https://unpkg.com/@polymer/lit-element?module';
+
+class MessagingSample extends LitElement {
+  constructor() {
+    super();
+    this._localConnection = this._remoteConnection = this._localChannel = this._remoteChannel = null;
```
kenchris: If you never check if these are null (they are undefined by default) why do this?

kenchris: Anyway they look nice on each on a separate line

src/content/datachannel/messaging/main.js
```diff
@@ -0,0 +1,164 @@
...
+      });
+      this._localChannel.addEventListener('close', () => {
+        console.log('Local channel closed!');
+        this.connected = false;
+      });
+      this._localChannel.addEventListener('message', e => this._onLocalMessageReceived(e));
+
+      this._remoteConnection.addEventListener('datachannel', e => this._onRemoteDataChannel(e));
+
+      const localOffer = await this._localConnection.createOffer();
```
kenchris: Could you do the remoteOffer in parallel instead of waiting? ie 

```
const initLocalOffer = async () => {
   ...
}

const initRemoteOffer = async () => {
   ...
}

await Promise.all([initLocalOffer, initRemoteOffer);
console.log("All initialized!")
```

ErikHellman: Tested this and that doesn't work. The result from `createOffer()` must be set before we can call `createAnswer()`.

src/content/datachannel/messaging/main.js
```diff
@@ -0,0 +1,163 @@
...
+      console.log(`Got local offer ${JSON.stringify(localOffer)}`);
+      await this._localConnection.setLocalDescription(localOffer);
+      await this._remoteConnection.setRemoteDescription(localOffer);
+
+      const remoteOffer = await this._remoteConnection.createAnswer();
```
fippo: s/Offer/Answer/

src/content/datachannel/messaging/main.js
kenchris: Then do

```
const initLocalOffer = async () => {
   ...
}

const initRemoteOffer = async () => {
   ...
}

await initLocalOffer();
await initRemoveOffer();
```

At least that makes it clear what code belongs where

ErikHellman:
@samdutton @alvestrand 
What is your opinion on this approach? The solution will be similar to what many other framework uses and this will still be based on something that is available as Web APIs in all browsers.

src/content/datachannel/messaging/main.js
```diff
@@ -0,0 +1,164 @@
...
+      console.log('Not sending empty message!');
+      return;
+    }
+    console.log('Sending remote message: ', value);
+    channel.send(value);
+    textarea.value = '';
+  }
+}
+
+customElements.define('messaging-sample', MessagingSample);
```
henbos: nit: newline?

src/content/datachannel/messaging/main.js
```diff
@@ -0,0 +1,164 @@
...
+    this._remoteChannel.addEventListener('message', e => this._onRemoteMessageReceived(e));
+    this._remoteChannel.addEventListener('close', () => {
+      console.log('Remote channel closed!');
+      this.connected = false;
+    });
+  }
+
+  _onRemoteMessageReceived(event) {
+    console.log(`Local message received by remote: ${event.data}`);
+    this.remoteMessages += event.data + '\n';
```
henbos, line 94: Can we have a unittest that clicks the button and verifies the message is received?

ErikHellman: I will add a UI test once we've switched to Nightwatch. Don't want to write a test using the old fw before then.

samdutton:
> Nobody builds a real application the way the samples are designed today (barebone)

That's been deliberate: the samples have been designed to show off WebRTC APIs and methods with as little possible  other code  (within reason!)

> Samples should be contextual, or lots of developers will have a hard time understanding how to use the APIs in their projects.

Ideally, I think you have both: barebones code (to enable devs to understand core APIs and methods) and contextualised code (like appr.tc).

However, as @kenchris says, I don't think it matters that much — and contextualising code does have advantages.

henbos:
> The framework I've used here is basically just a thin wrapper on top of the standard Web Components APIs (templates, Shadow DOM and custom elements). An example of how it would look like without lit-elements and only using these standard APIs can be seen here (works only in Chrome at the moment); https://codepen.io/ErikHellman/pen/BPVaZX
> 
> Since this is standard APIs that are coming to all browsers, my opinion is that samples for web APIs should embrace this way of creating UI as well. Nobody builds a real application the way the samples are designed today (barebone), and it serves no purpose in samples as they lack context.
> 
> From a developer relations perspective this makes more sense. Samples should be contextual, or lots of developers will have a hard time understanding how to use the APIs in their projects.

I'm afraid I'm lacking the web guru expertise to have an opinion about Lit since I've never heard of it (but your argument sounds reasonable), I can only serve as an RTCPeerConnection expert in my reviewing. If you need more reviewer's opinions we need to find other reviewers than me since my experience with JavaScript is mostly writing tests for Chrome (using https://cs.chromium.org/chromium/src/third_party/WebKit/LayoutTests/resources/testharness.js).

henbos (approved):

PR #1078, opened 2018-07-20
Revert "Fix button state change on quickly clicking restart button"

fippo:
Reverts webrtc/samples#1075 which is now broken in Firefox


alvestrand:
@ErikHellman Checking for "completed" is a known bogon. Don't do it.

fippo:
it is? Works in Chrome -- on the sender side at least.

Issue #1045, opened 2018-05-04
Regression: Broken logo is displayed instead of the actual video's preview [Canvas]


vkkkumar:
## Browser affected
Chrome Version: 67.0.3388.0 (Dev)
OS: Chrome

## Steps to reproduce
What steps will reproduce the problem?
(1) Go to https://webrtc.github.io/samples/src/content/capture/canvas-record/
(2) Click on "Start recording"
(3) Click on "Stop recording" after a while 
(4) Observe the Canvas on the right

Note: Not available in M65.

## Expected results
Video Preview or blank canvas

## Actual results
Video Broken logo has been displayed 

Original Bug: https://bugs.chromium.org/p/chromium/issues/detail?id=837519&desc=2

vkkkumar:
This issue is reproducible in M68 as well. @rantonysamy

ErikHellman:
Setting the video.src attribute when we stop the recording solves this problem as well. See referenced PR.

PR #1080, opened 2018-07-23
Sets the video source on stop recording.

ErikHellman:
- Move the call to `video.src = ...` to `handleStop()`.

See #1045

### DIFF
--- a/src/content/capture/canvas-record/js/main.js
+++ b/src/content/capture/canvas-record/js/main.js
@@ -31,7 +31,8 @@ recordButton.onclick = toggleRecording;
 playButton.onclick = play;
 downloadButton.onclick = download;
 
-document.addEventListener('DOMContentLoaded', main);
+// Start the GL teapot on the canvas
+main();
 
 const stream = canvas.captureStream(); // frames per second
 console.log('Started stream capture from canvas element: ', stream);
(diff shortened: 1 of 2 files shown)


src/content/capture/canvas-record/js/main.js
```diff
@@ -31,15 +31,7 @@ recordButton.onclick = toggleRecording;
...
-// Call main() in demo.js to init teapot
-main();
-
-// window.isSecureContext could be used for Chrome
-let isSecureOrigin = location.protocol === 'https:' || location.hostname === 'localhost';
-if (!isSecureOrigin) {
-  alert('getUserMedia() must be run from a secure origin: HTTPS or localhost.\n\nChanging protocol to HTTPS');
-  location.protocol = 'HTTPS';
-}
+document.addEventListener('DOMContentLoaded', main);
```
samdutton: Is this necessary, given that the script element is at the bottom of the page?

(Or am I missing something...)

ErikHellman: Not really. I think this ended up here because I was doing something similar in a different place and my brain is bad at multi-tasking. :)

ErikHellman: Removed it.

samdutton (approved):
LGTM

PR #1081, opened 2018-08-01
reactivate ice restart test

fippo:
also fixes the cause of the flakyness

@ErikHellman @alvestrand PTAL

### DIFF
--- a/src/content/peerconnection/restart-ice/js/main.js
+++ b/src/content/peerconnection/restart-ice/js/main.js
@@ -105,10 +105,10 @@ function call() {
     console.log(`Using audio device: ${audioTracks[0].label}`);
   }
   const servers = null;
-  pc1 = new RTCPeerConnection(servers);
+  pc1 = window.pc1 = new RTCPeerConnection(servers);
   console.log('Created local peer connection object pc1');
   pc1.onicecandidate = e => onIceCandidate(pc1, e);
-  pc2 = new RTCPeerConnection(servers);
+  pc2 = window.pc2 = new RTCPeerConnection(servers);
   console.log('Created remote peer connection object pc2');
   pc2.onicecandidate = e => onIceCandidate(pc2, e);
   pc1.oniceconnectionstatechange = e => {
--- a/src/content/peerconnection/restart-ice/js/test.js
+++ b/src/content/peerconnection/restart-ice/js/test.js
@@ -12,20 +12,13 @@
 // The test script language comes from tape.
 const test = require('tape');
 
-/* Firefox TODO once ice restarts are implemented
- * https://bugzilla.mozilla.org/show_bug.cgi?id=906986
- * 1) re-enable test
- * 2) fix getStats, ideally using spec-stats
- *
- * TODO: once onselectedcandidatepairchange is supported this test gets simpler.
- */
-
 function getTransportIds(stats) {
   let localId;
   let remoteId;
+  // TODO: figuring out the currently active candidate pair is tricky cross-browser.
+  // https://github.com/w3c/webrtc-stats/issues/358
   stats.forEach(report => {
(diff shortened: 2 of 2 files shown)


src/content/peerconnection/restart-ice/js/main.js
```diff
@@ -105,10 +105,10 @@ function call() {
     trace(`Using audio device: ${audioTracks[0].label}`);
   }
   const servers = null;
-  pc1 = new RTCPeerConnection(servers);
+  pc1 = window.pc1 = new RTCPeerConnection(servers);
```
fippo: this is ugly but it allows accessing pc1 from marionette

src/content/peerconnection/restart-ice/js/test.js
```diff
@@ -52,12 +50,10 @@ test('PeerConnection restart ICE sample', {skip: true}, t => {
...
       return driver.wait(() => driver.findElement(webdriver.By.id('callButton')).isEnabled(), 30 * 1000);
     })
+    .then(() => driver.findElement(webdriver.By.id('callButton')).click())
     .then(() => {
-      driver.findElement(webdriver.By.id('callButton')).click();
       t.pass('Pressed callButton');
-      return driver.wait(() => {
-        return driver.executeScript('return pc1 && pc1.iceConnectionState === \'completed\' || \'connected\';');
-      }, 30 * 1000);
+      return driver.wait(() => driver.executeScript(() => window.pc1 && (pc1.iceConnectionState === 'completed' || pc1.iceConnectionState === 'connected')))
```
fippo: the string executeScripts were really ugly -- and potentially hiding js errors from a linter

src/content/peerconnection/restart-ice/js/test.js
```diff
@@ -68,18 +64,16 @@ test('PeerConnection restart ICE sample', {skip: true}, t => {
...
+      // listen for iceconnectionstatechange and store events.
+      return driver.executeScript(() => {
+        window.icestates = [];
+        pc1.addEventListener('iceconnectionstatechange', () => window.icestates.push(pc1.iceConnectionState));
+      });
     })
+    .then(() => driver.findElement(webdriver.By.id('restartButton')).click())
     .then(() => {
       t.pass('ICE restart triggered');
-      driver.manage().timeouts().setScriptTimeout(150000);
```
fippo, line 75: I think this was flaky as it could add the event listener after the ice restart had gone to connected/completed. Adding the listener earlier and then checking for the checking->connected transition seems better.

fippo: oh lovely. chrome doesn't go back to checking in this case. So this will have to listen for iegatheringstatechange instead...

fippo:
even more lovely... chrome gives me spec-stats now (yay) but it also gives me the *old* selected candidate pair. I tried checking writable but that was true on both the old and the new pair.

@alvestrand how do i tell the generation of a candidate pair from the stats?

alvestrand:
@fippo of course you use the usernameFragment attribute, that's what the webrtc spec says..... no such stats attribute? that has to be a bug - please file! ;-)

fippo:
if we figure out a good cross-browser way to determine whether the ice restart is done we can also fix https://github.com/webrtc/samples/issues/1036

ErikHellman (approved):
LGTM

I think this is good enough to merge. There are more improvement we can make, but let's do it in small increments.

fippo:
I think the bullet to bite here is writing browser specific code...
* checking the offerer goes to connected and then completed in Chrome (and safari)
* checking the offerer goes to checking and then connected in Firefox

will update.

fippo:
lets hope this is not flaky anymore...

PR #1083, opened 2018-08-03
Updates our eslint configuration to Google standard

ErikHellman:
- Disabled rules that require major changes and have little impact
- Excluding third_party directory


henbos (approved):
LGTM, though I don't know anything about the eslint stuff so that bit is stampy

PR #1087, opened 2018-08-07
make getStats code better in FF65+

fippo:
the code currently accesses a stats report.timestamp which
triggers the warnings described in
  https://blog.mozilla.org/webrtc/getstats-isremote-65/

Also adds the isRemote checks. While remote outbound-rtp
can not happen when using sender.getStats() people might
copy this technique and apply to pc.getStats() so it is
better to show

from discussion with @jan-ivar


jan-ivar (approved):

ErikHellman:
Ping @fippo 

Can you rebase this?

fippo:
i was about to say "pong" :-)

PR #1089, opened 2018-08-08
Calculate the correct size for video and canvas elements

### LINKED ISSUE #1088: css video cropping issues
it looks like the recent changes to use more of the page width lead to ugly black boxes (I think its called letterboxing?) around the video. Which was present before (see e.g. https://fippo.github.io/webrtc/src/content/peerconnection/pc1/) but much less visible. 

wdyt of removing the height attribute? That would let the fixed aspect ratio kick in.

ErikHellman:
Size is based on width and has a 4:3 aspect ratio.

Fixes #1088 

CSS is fun!

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -7,12 +7,12 @@ module.exports = function(grunt) {
   grunt.initConfig({
     // make node configurations available
     pkg: grunt.file.readJSON('package.json'),
-    csslint: {
-      options: {
-        csslintrc: '.csslintrc'
-      },
-      src: ['src/content/**/*.css', '!**/third_party/*.css' ]
-    },
+    // csslint: {
+    //   options: {
+    //     csslintrc: '.csslintrc'
+    //   },
+    //   src: ['src/content/**/*.css', '!**/third_party/*.css' ]
+    // },
     eslint: {
       options: {
         configFile: '.eslintrc',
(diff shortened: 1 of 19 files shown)


guidou (approved):
rubberstamp

Gruntfile.js
```diff
@@ -7,12 +7,12 @@ module.exports = function(grunt) {
   grunt.initConfig({
     // make node configurations available
     pkg: grunt.file.readJSON('package.json'),
-    csslint: {
-      options: {
-        csslintrc: '.csslintrc'
-      },
-      src: ['src/content/**/*.css', '!**/third_party/*.css' ]
-    },
+    // csslint: {
```
guidou, line 10: We generally prefer to remove commented-out code.

guidou (commented):
lgtm % comment

Gruntfile.js
```diff
@@ -55,13 +55,13 @@ module.exports = function(grunt) {
 
   // enable plugins
   grunt.loadNpmTasks('grunt-contrib-compress');
-  grunt.loadNpmTasks('grunt-contrib-csslint');
+  // grunt.loadNpmTasks('grunt-contrib-csslint');
```
guidou, line 58: remove commented out code here too

samdutton: Why is CSS linting being removed?

src/content/capture/canvas-pc/css/main.css
```diff
@@ -8,36 +8,17 @@
...
+  height: calc(var(--width) * 0.75);
+  margin: 1em;
   vertical-align: top;
-  width: calc(50% - 12px);
 }
 
 video {
-  height: 231px;
-  margin: 0 0 20px 0;
+  --width: calc(45%);
```
samdutton, line 19: If you want to define the width only once, you could put the variable definition in the `html` block (though that might be less clear!)

src/content/capture/canvas-record/css/main.css
```diff
@@ -5,60 +5,36 @@
  *  that can be found in the LICENSE file in the root of the source
  *  tree.
  */
- button {
-  margin: 0 3px 10px 0;
-  padding-left: 2px;
-  padding-right: 2px;
-  width: 99px;
+button {
+    margin: 0 3px 10px 0;
```
samdutton, line 9: Indentation should be two spaces, not four.

Gruntfile.js
ErikHellman: There is a bug in CSS lint in that it can't handle custom vars. Custom vars is needed if we want to be able to define an aspect ratio for the video or canvas. :/

src/content/capture/canvas-pc/css/main.css
ErikHellman: I was considering that at first, but the value would be different in the samples. This makes it a bit repetitive, but at least the samples are more isolated.

samdutton:
Ah — see my comment about Stylelint. That solves the problem :).

samdutton:
Yeah — swings and roundabouts!

PR #1090, opened 2018-08-08
Fix broken for

sampaiodiego:
**Description**
I think latest changes to ES6 broke it:

image

**Purpose**
Make it work again 😄


ErikHellman:
Thanks for noticing! This looks good. Please sign the CLA as described above so I can merge it.

ErikHellman:
I've fixed this bug in a different commit. Thanks for noticing!

sampaiodiego:
thanks @ErikHellman .. I've signed the CLA, I wonder if the reason it didn't got it was because I used the github UI to do the change.

anyways, thanks for fixing it.

PR #1091, opened 2018-08-08
Update main.js

BurhanBozkurt:
**Description**
A bug caused by the invalid variable declaration, broke the script.

**Purpose**
To fix the bug.


ErikHellman:
Thanks for noticing. The bug is fixed in a different commit.

BurhanBozkurt:
Great!
Keep up the good work ;)

PR #1095, opened 2018-08-13
Migrate to gulp

ErikHellman:
Because 2018.

Seems like gulp is much preferred over grunt these days. The extensions are updated more frequently and have less security issues.


package.json
```diff
@@ -14,25 +14,27 @@
...
-    "grunt-htmlhint": ">=0.9.12",
+    "gulp-csslint": "^1.0.1",
+    "gulp-eslint": "^5.0.0",
+    "gulp-htmlhint": "^2.1.1",
+    "gulp-zip": "^4.2.0",
     "pem": "^1.9.4",
     "webrtc-utilities": "^1.0.8"
+  },
+  "dependencies": {
+    "gulp": "^3.9.1"
```
fippo, line 38: nit: devDependency. This isn't published to npm so it does not really matter but... :-)

fippo (approved):
i don't like gulp but... E_TOO_MANY_TOOLS

ErikHellman:
I've kidn of given up on grunt as I keep getting security audit warnings from the extensions. Checked around and "everyone" seems to migrate to gulp. It is also something that Google seems to recommend themselves: https://developers.google.com/web/ilt/pwa/introduction-to-gulp

PR #1096, opened 2018-08-13
Fix/migrate to nightwatchjs

ErikHellman:
**Description**
We should get rid of the custom test runner in webrtc-uttilities and use something the rest of the community would be more familiar with. After some investigation I've decided that [Nightwatch.JS](https://nightwatchjs.org) is a suitable candidate.

**Purpose**
This change introduces NightwatchJS and removes our custom testrunner. It also demonstrates how to create custom commands and assertions suitable for testing WebRTC.


ErikHellman:
This is a WIP. I'm migrating one test at a time.

@fippo Most of the UI test frameworks seems to allow some form of extensions, but this was the one where it was the easiest. Let me know what you think of this approach.

.travis.yml
```diff
@@ -1,46 +1,17 @@
...
+os: osx
 language: node_js
 cache:
  - node_modules
 node_js:
-  - 6
+  - lts/*
+addons:
+  chrome: stable
+  firefox: latest
```
fippo: IIUC that will run only in chrome and firefox stable? This would be a significant step backward.

nightwatch/customCommands/waitForClientConnected.js
```diff
@@ -0,0 +1,61 @@
...
+      }, 1000);
+    } else {
+      cb(false);
+    }
+  };
+
+  this.api.execute(function(selector) {
+    try {
+      const element = document.querySelector(selector);
+      return element && (element.readyState === 4);
```
fippo: this waits for the video element readyState? I am much in favor of that, it is a better check than the iceConnectionState check (which in Firefox is only that and ignores any DTLS or codec errors).

src/content/datachannel/basic/js/test.js
```diff
@@ -1,49 +1,31 @@
...
-test('Basic datachannel sample', {skip: process.env.BROWSER === 'firefox'},
-  t => {
-    const webdriver = require('selenium-webdriver');
-    const seleniumHelpers = require('webrtc-utilities').seleniumLib;
-    const driver = seleniumHelpers.buildDriver();
-    const path = '/src/content/datachannel/basic/index.html';
-    const url = (process.env.BASEURL ? process.env.BASEURL : ('file://' + process.cwd())) + path;
+    browser
+      .click('#closeButton')
+      .expect.element('#sendButton').to.not.be.enabled.before(50);
```
fippo: this removes the webrtc specifics of this test?

src/content/datachannel/datatransfer/js/test.js
```diff
@@ -1,43 +1,20 @@
...
-    .then(null, err => {
-      t.fail(err);
-      t.end();
-    });
-});
+    browser
+      .url(url)
+      .click('#sendTheData')
+      .pause(1000)
+      .assert.value('#receiveProgress', '16777200')
```
fippo: magic number? :-)

src/content/devices/input-output/js/test.js
```diff
@@ -1,143 +1,20 @@
...
-});
+export default {
+  'It should have select elements for each media device': (browser) => {
+    const path = '/src/content/devices/input-output/index.html';
+    const url = 'file://' + process.cwd() + path;
+
+    browser
+      .url(url)
+      .waitForElementVisible('#audioSource option:nth-of-type(1)', 1000, 'Check that there is at least one audio source')
+      .waitForElementVisible('#audioOutput option:nth-of-type(1)', 1000, 'Check that there is at least one audio output')
```
fippo: this won't pass in Firefox as it does not implement output devices yet.

src/content/getusermedia/gum/js/test.js
```diff
@@ -1,84 +1,19 @@
...
-    })
-    .then(null, err => {
-      t.fail(err);
-      t.end();
-    });
-});
+    browser
+      .url(url)
+      .waitForElementVisible('video', 5000)
+      .waitForClientConnected('video', 5000)
```
fippo: this has the wrong name. It waits for video to be shown, not for anything to be connected

src/content/getusermedia/resolution/js/test.js
```diff
@@ -1,70 +1,29 @@
...
-  resolutionTest(t, 'fourK', 4096);
-});
-*/
+export default {
+  'It should have a video element with specific width': (browser) => {
+    const path = '/src/content/getusermedia/resolution/index.html';
+    const url = 'file://' + process.cwd() + path;
+
+    browser
+      .url(url)
```
fippo: reloading had the advantage that the device was more likely to be released. The code releases the old stream before calling GUM again though so this might not be required.

src/content/getusermedia/resolution/js/test.js
```diff
@@ -1,70 +1,29 @@
...
+      .click('button#qvga')
+      .waitForElementVisible('video', 5000)
```
fippo: probably earlier: make this timeout a constant - or maybe there is a global nightwatch setting?

src/content/peerconnection/audio/js/test.js
```diff
@@ -1,68 +1,20 @@
...
-  const driver = seleniumHelpers.buildDriver();
-  if (process.env.BROWSER === 'firefox') {
-    t.pass('Firefox not supported yet');
-    t.end();
-    return;
+    // TODO Test all codecs?
+    browser
+      .url(url)
+      .click('#callButton')
+      .waitForClientConnected('#audio2', 1000, 'Receiving remote audio.')
```
fippo: this removes the test for the codec that actually gets used

fippo (commented):
I am a bit concerned that this reduces both the browser matrix tested as well as no longer shows useful testing techniques :-/

@tsahilevi can probably say more about the Nightwatch bits since testrtc.com uses nightwatch.

.travis.yml
fippo: https://github.com/nightwatchjs/nightwatch/wiki/Chrome-Setup suggests it should be possible to set the binary path, I assume also for firefox.

src/content/datachannel/datatransfer/js/test.js
ErikHellman: Based on the default size in the index.html :)

src/content/datachannel/basic/js/test.js
ErikHellman: Yes. The point here is to show how to do a UI test for WebRTC. We let the WebRTC API control the enabled state and the test will just verify that.

src/content/devices/input-output/js/test.js
ErikHellman: Good to know. Currently fighting with the geckodriver, so I guess I would run into this once that happens. :)

src/content/getusermedia/gum/js/test.js
ErikHellman: Agree. The name of the function is not optimal. Will rename.

.travis.yml
ErikHellman: Yes, I started with just chrome stable, but will expand it to firefox and safari as well. Currently struggling with getting geckodriver working properly

ErikHellman:
I'm currently adding support for firefox and safari as well. I'll also re-enable the test matrix if I notice that running the tests takes a long time, or if we need it for some other reason. Nightwatch does support parallell execution of different browsers.

.travis.yml
fippo: wait until you meet the old geckodriver that is required for Firefox ESR... :-)

src/content/datachannel/basic/js/test.js
fippo: so how does this teach people how to test webrtc then?

fippo:
parallel execution is probably not a good idea. The travis machines are not very powerful so running two browsers in parallel is pushing it usually. Matrix builds ftw

fippo:
wdyt about hooking up nightwatch and travis-multirunner? Roughly like in this patch:
```
diff --git a/nightwatch.conf.js b/nightwatch.conf.js
index 2945d13..35735fa 100644
--- a/nightwatch.conf.js
+++ b/nightwatch.conf.js
@@ -1,3 +1,4 @@
-require('babel-core/register')
-
-module.exports = require('./nightwatch.json')
+require('babel-core/register');
+const config = require('./nightwatch.json');
+config.test_settings.chrome.desiredCapabilities.chromeOptions.binary = process.cwd() + '/browsers/bin/chrome-' + process.env.BVER; 
+module.exports = config;
diff --git a/nightwatch.json b/nightwatch.json
index 8c609ef..0a92f50 100644
--- a/nightwatch.json
+++ b/nightwatch.json
@@ -67,7 +67,8 @@
             "--disable-translate",
             "--no-process-singleton-dialog",
             "--mute-audio"
-          ]
+          ],
+          "binary": "to be replaced with environment variable"
         }
       }
     },
@@ -110,4 +111,4 @@
       }
     }
   }
-}
\ No newline at end of file
+}
```
You'll need to have installed travis-multirunner, set BROWSER=chrome and BVER=unstable and have run node_modules/travis-multirunner/setup.sh before.

Might be a good idea to just drop the json and write the config as js. Easier to evaluate other things like os.platform then.

ErikHellman:
Yep. Good idea. Right now I'm fighting with geckodriver to get Firefox test running. 

This is way harder than it should be. :)

ErikHellman:
@fippo I managed to get the test matrix working without the travis-multirunner (See the latest `.travis.yml` in this PR). Still having issues with Firefox and geckodriver, but getting close now! :)

fippo:
what is wrong with travis-multirunner?
I don't think using osx images (which are probably 'expensive' for travis) when linux images can do the job is good. Seems like matrix builds on multiple OS are a pain though...

ErikHellman:
Regarding the travis-multirunner; we really don't need it. The CI matrix on travis can be configured entirely in the config file, even with different OS.
Regarding OSX: More people will be using OSX than Linux desktop. Also, I want to add Safari to the test matrix.

I can run Firefox and Chrome on Linux, but the config for those is actually more complicated than just running it on OSX.

nightwatch/customAsserts/videoHeight.js
```diff
@@ -0,0 +1,19 @@
+exports.assertion = function(selector, videoHeight, msg) {
+  this.message = msg || `Testing if element <${selector}> has videoHeight ${videoHeight}`;
+  this.expected = videoHeight;
+
+  this.pass = function(value) {
+    return value === videoHeight;
+  };
```
henbos: nit: Should we one-line it if it's super small?
```
this.pass = function(value) { return value === videoHeight; };
```

nightwatch/customAsserts/videoWidth.js
```diff
@@ -0,0 +1,19 @@
+exports.assertion = function(selector, videoWidth, msg) {
```
henbos, line 1: Any particular reason for this file convention? At first I thought it was silly to put videoHeight and videoWidth in separate files since they are essentially the same thing, but now I see the folder is "customAsserts". That's fine I'm just curious why not a "customAsserts.js"?

nightwatch/customCommands/waitForClientConnected.js
```diff
@@ -0,0 +1,41 @@
+import EventEmitter from 'events';
+import assert from 'assert';
+
+class WaitForClientConnected extends EventEmitter {
+  constructor() {
+    super();
+
```
henbos: nit: Remove empty line

nightwatch/customCommands/waitForClientConnected.js
```diff
@@ -0,0 +1,41 @@
...
+    return this;
+  }
+
+  check(selector, timeout, message) {
+    const self = this;
+    const executeArgs = [selector];
+    this.api.execute(function(selector) {
+      const element = document.querySelector(selector);
+      return element && (element.readyState === 4);
```
henbos: 4? Is element a MediaStreamTrack or something else?
Can you document what "selector" is and shouldn't this be a string comparison if its a readyState enum?

nightwatch/customCommands/waitForClientConnected.js
```diff
@@ -0,0 +1,41 @@
+import EventEmitter from 'events';
+import assert from 'assert';
+
+class WaitForClientConnected extends EventEmitter {
```
henbos: Should this be WaitForHTMLMediaElementIsReady?
The check() is not really designed to whether or not the clients are connected. RTCPeerConnection-wise, that would probably be the signalingState or iceConnectionState 'connected'.

src/content/datachannel/datatransfer/test/test.js
```diff
@@ -0,0 +1,20 @@
...
+    const url = 'file://' + process.cwd() + path;
+
+    browser
+      .url(url)
+      .click('#sendTheData')
+      .pause(1000)
+      .assert.value('#receiveProgress', '16777200')
+      .end();
+  }
+};
```
henbos: nit: No newline at the end of file

src/content/getusermedia/resolution/test/test.js
```diff
@@ -0,0 +1,32 @@
...
+export default {
+  'It should have a video element with specific width': (browser) => {
+    const path = '/src/content/getusermedia/resolution/index.html';
+    const url = 'file://' + process.cwd() + path;
+
+    browser
+      .url(url)
+      .click('button#qvga')
+      .pause(500)
+      .waitForElementVisible('video', 5000)
```
henbos, line 17: Sometimes 1000, sometimes 5000... Is there a default const kTimeout = x; that we should stick too?

src/content/peerconnection/dtmf/test/test.js
```diff
@@ -0,0 +1,28 @@
...
+      .click('/html/body/div/div[@id=\'dialPad\']/div[1]/button[1]') // 1
+      .click('/html/body/div/div[@id=\'dialPad\']/div[3]/button[1]') // 9
+      .click('/html/body/div/div[@id=\'dialPad\']/div[3]/button[4]') // #
+      .click('/html/body/div/div[@id=\'dialPad\']/div[4]/button[1]') // A
+      .useCss()
+      .assert.value('input#sentTones', '1 9 # A ')
+      .click('#hangupButton')
+      .end();
+  }
+};
```
henbos: nit: Missing newline at the end.

henbos (changes requested):
I know nothing about which testing library to use so the Nightwatch.js and json files being updated here I'm mostly rubber-stamping, but I love the new tests they're super short and easy to read.

There are several tests that are only renamed but not updated. Are these running the old framework or are they disabled?

nightwatch/customAsserts/videoHeight.js
ErikHellman: In that case I would do:
```
this.pass = (value) => value === videoHeight;
```

nightwatch/customAsserts/videoWidth.js
ErikHellman: The assert function is named after the file :)

nightwatch/customCommands/waitForClientConnected.js
ErikHellman: 4 is basically 'client connected', which in this case means that media is playing.

I'll document it better.

nightwatch/customCommands/waitForClientConnected.js
ErikHellman: I based this on what readyState === 4 is described as, but you're right. Fixing.

src/content/getusermedia/resolution/test/test.js
ErikHellman: There is a default and I should probably clean it up in the tests. However, media stuff needs longer waiting than other elements..

ErikHellman:
The main purpose of chosing Nightwatch is not so much because this specific framework, but to get rid of the custom WebRTC stuff in WebRTC utilities. Less to maintain here and pick something that others are more likely to be using.

nightwatch/customCommands/waitForClientConnected.js
fippo: HTMLVideoElement.HAVE_ENOUGH_DATA

fippo:
multiple OS seem less bad than I thought, https://github.com/DamonOehlman/travis-multirunner/blob/master/.travis.yml shows this actually.

henbos (approved):

nils-ohlmeier:
So tests are more readable now, but they are only checked in Chrome?
Given that it feels like Firefox and Chrome soon could finally expose the same API's I'm not convinced this is a step in the right direction.

ErikHellman:
@nils-ohlmeier Yeah, it's a temporary problem with the Geckodriver. Apparently I'm not alone, so I'm investigating how to solve it in a nice way. 

If I exclude the custom assertions and command I've created, tests runs fine. But that means we can't properly test WebRTC stuff.

I'm not about to merge this until I have it working for at least 2 browsers on TravisCV. Currently I almost got it working for Safari as well as Chrome.

lgrahl:
I would prefer it if Chrome and Firefox are always tested since Safari and Chrome share pretty much the same code base.

ErikHellman:
One thing we should remember here is that these are samples showing how web developers how to do things. My intention is to show both how the API is used as well as how to do UI testing for WebRTC applications. 

The tests here should not be part of the QA for WebRTC. It doesn't matter which browsers we're testing on since it is only for showing _how to test_. I'd prefer to have Chrome, Firefox AND Safari, but as there seems to be some problems with the geckodriver, I would be satisfied with Chrome and Safari.

fippo:
But reducing the test matrix from
* chrome stable/beta/unstable
* firefox stable/beta/nightly

to
* chrome stable/beta
* safari stable

is counterproductive.

It was a long and bloody fight to get people to stop shipping webrtc stuff chrome-only (and I am not looking just at you hangouts). These samples (and the 'google recommends' that is implied) were important in delivering a "all browsers" message.

lgrahl:
I share @fippo's concern. If this is about showing people how to test WebRTC and if it doesn't work with Firefox at the moment, then people should stay away from it until it works. And so should the examples.

Perhaps we can nudge the people who are working on geckodriver regarding that particular issue. @ErikHellman is there a Bugzilla entry for the problem?

ErikHellman:
I'm merging this PR now as I got the tests working on both Firefox and Chrome using the selenium-standalone server. 

While we're not having all the browser versions as we did before, we got a much more flexible solution in place. I'll be extending this with more tests and additional browsers in future PRs.

fippo:
I think that is a bad decision. Dropped my commit bit, ha det bra :wave:

Issue #973, opened 2017-12-11
Pause button is not working while playing a recorded video.


vkkkumar:
**Browsers and versions affected**
M64 Dev ToT
10172.0.0 / 64.0.3282.11
(Chrome)

**Description**
Pause button is not working. 

**Steps to reproduce**
0. Browse to https://webrtc.github.io/samples/src/content/getusermedia/record/
1. Open the javascript console (alt+cmd+j or ctr+shift+j) and watch for any errors
2. Enable "Enable experimental Web Platform features" flag in chrome://flags.
3. Click on "Start Recording" and record audio/video for a min or so
4. Click on "Stop Recording" 
5. Click on "Play" to playback the media stream recorded
6. Pause the video by clicking puase button

**Expected results**
Video needs to be paused. 

**Actual results**
Video cannot be paused.

**Tip**
Seems like an html issue. Under Video tag, Instead of "loop controls", "controls loop" needs to be added for looping the video. Probably this is blocking the feature of pausing.

KaptenJansson:
Please submit a fix in a PR if you think this is important.

RpCoding:
Any updates on this? I can't get the pause button to work and could need some help.

vkkkumar:
Hi RpCoding, Haven't started the work on it. Soon will update on this issue. Thanks!

PR #1108, opened 2018-08-29
Fix for recording user media sample - #973

ErikHellman:
Fix for recording user media sample - #973


src/content/getusermedia/record/js/main.js
```diff
@@ -40,23 +45,11 @@ recordButton.addEventListener('click', () => {
 const playButton = document.querySelector('button#play');
 playButton.addEventListener('click', () => {
   const superBuffer = new Blob(recordedBlobs, {type: 'video/webm'});
+  recordedVideo.src = null;
+  recordedVideo.srcObject = null;
   recordedVideo.src = window.URL.createObjectURL(superBuffer);
-  // workaround for non-seekable video taken from
```
fippo, line 44: non-seekable videos are still a problem i think...

src/content/getusermedia/record/js/main.js
```diff
@@ -99,37 +80,35 @@ function handleDataAvailable(event) {
...
-function handleStop(event) {
-  console.log('Recorder stopped: ', event);
-}
-
 function startRecording() {
   recordedBlobs = [];
   let options = {mimeType: 'video/webm;codecs=vp9'};
   if (!MediaRecorder.isTypeSupported(options.mimeType)) {
-    console.log(options.mimeType + ' is not Supported');
+    console.log(options.mimeType, ' is not Supported');
```
fippo: running https://github.com/cpojer/js-codemod/blob/master/transforms/template-literals.js on the whole codebase might be easier :-)

src/content/getusermedia/record/js/main.js
```diff
@@ -138,20 +117,31 @@ function startRecording() {
...
+    const stream = await navigator.mediaDevices.getUserMedia(constraints);
+    handleSuccess(stream);
+  } catch (e) {
+    console.error('navigator.getUserMedia error:', e);
+  }
 }
 
-navigator.mediaDevices.getUserMedia(constraints).then(handleSuccess).catch(handleError);
+// window.isSecureContext could be used for Chrome
+let isSecureOrigin = location.protocol === 'https:' || location.hostname === 'localhost';
```
fippo: we are requiring https on github pages so I don't think we need to handle this error

src/content/getusermedia/record/js/main.js
ErikHellman: Yep. But it doesn't affect this sample as you cannot seek in the remote video.

src/content/getusermedia/record/js/main.js
ErikHellman: 👍

src/content/getusermedia/record/js/main.js
ErikHellman: Yeah, I use both since I sometimes want the object displayed and sometimes just a string. This should probably be a template literal..

src/content/getusermedia/record/js/main.js
fippo: right, because you removed the controls. It would be still good to have this as a place to point people to if they have questions about seeking. Unless we can get @yellowdoge to write up a lengthy blog post instead :-)

ErikHellman: Since it isn't directly concerning WebRTC I'm hesitant to keep it there. I would also like to avoid having ugly workarounds in the samples as far as possible. Better then to have a sample that we "disable" until it works.

fippo: it is much more effective to keep this here and have people find it in context than having to find this hidden in some chrome bug. mediarecorder files won't be as easily seekable by design.

Reality is messy. WebRTC in particular so... and its getting worse actually.

ErikHellman: I'm leaving the comment out for now and I'm adding it to my list of "workaround snippets" that we can hopefully publish somewhere.

fippo: sounds like a hack... webrtchacks is always open for that :-)

src/content/getusermedia/record/css/main.css
```diff
@@ -22,13 +22,11 @@ p.borderBelow {
 }
 
 video {
-  /*height: 232px;*/
-  margin: 0 12px 20px 0;
+  /*margin: 0 12px 20px 0;*/
```
alvestrand: If you don't need it, delete it.

src/content/getusermedia/record/js/main.js
```diff
@@ -21,8 +27,7 @@ let sourceBuffer;
 
 const recordedVideo = document.querySelector('video#recorded');
 recordedVideo.addEventListener('error', function(ev) {
-  console.error('MediaRecording.recordedMedia.error()');
-  alert(`Your browser can not play ${recordedVideo.src} media clip. event: ${JSON.stringify(ev)}`);
+  console.error('MediaRecording.recordedMedia.error()', ev);
```
alvestrand: the error log is invisible when running the demo if you don't open the console.
Can you update some status in the page too? (agree that alert() needs to Die Die Die....)

src/content/getusermedia/record/js/main.js
```diff
@@ -40,23 +45,11 @@ recordButton.addEventListener('click', () => {
 const playButton = document.querySelector('button#play');
 playButton.addEventListener('click', () => {
   const superBuffer = new Blob(recordedBlobs, {type: 'video/webm'});
+  recordedVideo.src = null;
+  recordedVideo.srcObject = null;
   recordedVideo.src = window.URL.createObjectURL(superBuffer);
```
alvestrand, line 47: Does it work to set srcObject instead here, given orphis' quest to destroy createObjectUrl?
(I know createObjectURL(blob) is not being removed, but there's no reason to recommend createObjectURL in any way, shape or form.)

src/content/getusermedia/record/js/main.js
```diff
@@ -99,37 +80,35 @@ function handleDataAvailable(event) {
...
-function handleStop(event) {
-  console.log('Recorder stopped: ', event);
-}
-
 function startRecording() {
   recordedBlobs = [];
   let options = {mimeType: 'video/webm;codecs=vp9'};
   if (!MediaRecorder.isTypeSupported(options.mimeType)) {
-    console.log(options.mimeType + ' is not Supported');
+    console.log(`${options.mimeType} is not Supported`);
```
alvestrand: Same comment about showing something on the screen here.
In this case, it could be OK to show the MIME type you end up with on the screen.

src/content/getusermedia/record/js/main.js
```diff
@@ -99,37 +80,35 @@ function handleDataAvailable(event) {
...
         options = {mimeType: ''};
       }
     }
   }
   try {
     mediaRecorder = new MediaRecorder(window.stream, options);
   } catch (e) {
-    console.error(`Exception while creating MediaRecorder: ${e}`);
-    alert(`Exception while creating MediaRecorder: ${e}. mimeType: ${options.mimeType}`);
+    console.error('Exception while creating MediaRecorder:', e);
```
alvestrand, line 102: And here....

alvestrand (commented):
Mostly looks good. But replace console.error() with a function that updates an on-page status and logs console.error() - silent errors are the bane of demos.

src/content/getusermedia/record/js/main.js
ErikHellman: Unfortunately not, it causes a `TypeError` when setting a blob on the srcObject.

src/content/getusermedia/record/js/main.js
ErikHellman: Removing this error as it didn't make sense.

src/content/getusermedia/record/js/main.js
```diff
@@ -84,21 +81,26 @@ function startRecording() {
   recordedBlobs = [];
   let options = {mimeType: 'video/webm;codecs=vp9'};
   if (!MediaRecorder.isTypeSupported(options.mimeType)) {
-    console.log(`${options.mimeType} is not Supported`);
+    console.error(`${options.mimeType} is not Supported`);
+    errorMsgElement.innerHTML = `${options.mimeType} is not Supported`;
```
alvestrand: Still think that those 2 lines could be wrapped into a helper, but I'll let that pass.

alvestrand (approved):

Issue #1102, opened 2018-08-27
Chromium failed to generate and transfer data on RTCDataChannel


cpu82:
## Browser affected

Chromium 68.0.3440.106

## Description

It fails trying to generate and send data.

## Steps to reproduce

Go to https://webrtc.github.io/samples/src/content/datachannel/datatransfer/
Click on "Generate and send data"
(Ordered mode checkbox has been checked)

## Expected results

All the data has been sent.

## Actual results

The data has not been sent.

![screenshot at 2018-08-27 01-58-32](https://user-images.githubusercontent.com/4080230/44634722-acde2780-a99d-11e8-974b-8b91b9879e51.png)

ErikHellman:
Thanks for report! I'm investigating an issue with bufferedAmountLow event not being fired. That would explain why this doesn't work anymore.

cpu82:
Let me know if you need anything else to fix the issue.

Thanks

ErikHellman:
@cpu82 If you are able to trigger a `bufferedAmountLow` event in your own project with a RTCDataChannel, please let me know. At the moment I'm not able to trigger it, which is why the sample fails.

cpu82:
No, I'm using Chromium on FreeBSD. Is FreeBSD a supported OS? Can it be related to the problem?

ErikHellman:
Probably not. I'm filed an issue no Chrome that you can track for this: https://bugs.chromium.org/p/chromium/issues/detail?id=878682

I'll update this sample to only use a polling solution until the event is emitted proprely.

cpu82:
OK, I'll do some tests when you update the datatransfer sample.

cpu82:
I get this result by running these [tests](http://w3c-test.org/webrtc/RTCDataChannel-bufferedAmount.html).

![screenshot at 2018-08-29 18-51-19](https://user-images.githubusercontent.com/4080230/44802887-23a53b80-abbd-11e8-8645-15d0a968c4a2.png)

sumnulu:
```javascript
 while (sendProgress.value < sendProgress.max) {
      if (sendChannel.bufferedAmount > bufferFullThreshold) {
        if (usePolling) {
          setTimeout(sendAllData, 250);
        } else {
          sendChannel.addEventListener('bufferedamountlow', listener);
        }
      // ---- Missing break; ----
      } else {
        sendProgress.value += chunkSize;
        sendChannel.send(stringToSendRepeatedly);
      }
    }
```
[...]

PR #1110, opened 2018-08-30
Fixes the problem with bbufferedamoutnlow event not being fired

ErikHellman:
(Temporary) fix for #1102

The bufferedamountlow event is not being fired in any of the bbrowsers at the moment. Until that is ficxed, we need to use setTimeout() to send data instead.

See https://bugs.chromium.org/p/chromium/issues/detail?id=878682


src/content/datachannel/datatransfer/js/main.js
```diff
@@ -41,97 +43,78 @@ megsToSend.addEventListener('change', function(e) {
   }
 });
 
-function createConnection() {
+async function createConnection() {
   sendButton.disabled = true;
   megsToSend.disabled = true;
+  // Safari needs to trigger the media permission for RTCDataChannel to work
```
fippo: use adapter.browserDetails.browser. Also s/for RTCDataChannel to work/for the ice connection to work due to ice candidate restrictions/

src/content/datachannel/datatransfer/js/main.js
```diff
@@ -41,97 +43,78 @@ megsToSend.addEventListener('change', function(e) {
...
+    // console.log('sendChannel.bufferedAmountLowThreshold:', sendChannel.bufferedAmountLowThreshold);
+    // console.log('Buffered amount before sending:', sendChannel.bufferedAmount);
+    sendChannel.send(DATA_STRING);
+    sendProgress.value += CHUNK_SIZE;
+    // console.log('Buffered amount after sending:', sendChannel.bufferedAmount);
+    // console.log(`Sent another ${CHUNK_SIZE} bytes!`);
+
+    // We need to use setTimeout() at the moment due to a bug with bufferedamountlow event.
+    // See https://bugs.chromium.org/p/chromium/issues/detail?id=878682 for details
+    // sendChannel.addEventListener('bufferedamountlow', sendData, {'once': true});
```
fippo: i don't think it is as bad... the old strategy from 4c06b2ebf40f46d7969cb7b035f4dfa71731baf9 works afaics. Also this was originally written by bemasc who implemented bufferedamountlow. Main problem is that this behaviour might be according to an old version of the spec which has probably flipflopped.

ErikHellman: The comment is still valid though. Once it's fixed, we can change to this behavior.

fippo: the strategy from 4c06b2ebf40f46d7969cb7b035f4dfa71731baf9 works (add a console log in the listener function) and is probably much more performant than setTimeout. It would benefit from the {once: true}.
Might be worth removing the pure polling fallback.

ErikHellman:
I changed the workaround to use a loop instead. Thanks @sumnulu for coming up with this. :)

Ping @fippo. What do you think, should we use this for now until the ground cause is fixed? I can see a problem here that the while-loop will block the main thread.

fippo:
From @lgrahl's comment in https://github.com/w3c/webrtc-pc/issues/1979#issuecomment-417613095 the old code showed the correct way to do things -- send until bufferedAmount >= myHighWaterMark

Of course this would be easier if Chrome supported sending blobs but this is just the most-starred webrtc.org issue with no movement since 2016 :-)

src/content/datachannel/datatransfer/js/main.js
```diff
@@ -41,97 +43,81 @@ megsToSend.addEventListener('change', function(e) {
...
 
-function createConnection() {
+async function createConnection() {
   sendButton.disabled = true;
   megsToSend.disabled = true;
+  // Safari needs to trigger the media permission for RTCDataChannel to work
+  if (adapter.browserDetails.browser === 'safari') {
+    try {
+      console.log('Call getUserMedia() to trigger media permission request.');
+      const stream = await navigator.mediaDevices.getUserMedia({'audio': true, 'video': true});
```
lgrahl: I don't see why this would be needed. The [associated issue has been marked as resolved](https://bugs.webkit.org/show_bug.cgi?id=174500) and this is not cross-browser. If it doesn't work, it's a bug.

fippo: the samples don't use a stun server. Not sure about the state of MDNS in Safari...

lgrahl: The linked issue has been resolved by the addition of mDNS. :man_shrugging:  Can anyone confirm that this has been deployed?

ErikHellman: If there is a different workaround that fixes the problem in safari, please make a PR with that (or send me the details and I'll implement it). Thanks! :)

lgrahl: What I'm saying is: There should be no need for a workaround if mDNS has been deployed in Safari. And if it hasn't yet, I would be strongly against adding that ~~workaround~~ behaviour because we shouldn't demonstrate bad practices in official WebRTC examples.

Edit: To clarify, this is a data only example. If it doesn't work in Safari without media permissions, then it doesn't work.

ErikHellman: We got tons of workarounds in the code today. I looked for other ways to solve this but couldn't find any. The alternative is that this sample won't work in Safari.

lgrahl: That is what I'm suggesting. Let it fail.

People will copy the sample code and believe it's fine to request media permissions for data only use cases because the official examples have done so. But it is definitely not fine. We should not encourage bad practices.

ErikHellman: Ok, let's do it like that. I'll update this PR tomorrow.

alvestrand (approved):
Looks good. I assume that it Just Works.

src/content/datachannel/datatransfer/js/main.js
```diff
@@ -41,97 +44,78 @@ megsToSend.addEventListener('change', function(e) {
...
-    result += String.fromCharCode(33 + Math.random() * 93);
+function sendData(e) {
+  console.log('BufferedAmountLow event:', e);
+  if (sendProgress.value < sendProgress.max) {
+    transferStatus.innerHTML = 'Sending data...';
+    sendChannel.addEventListener('bufferedamountlow', sendData, {once: true});
+
+    // The following is a workaround due to the problem with bufferedamountlow event not being fired on every
+    // call to send(), despite the amount of data exceeding the bufferedAmountLowThreshold.
+    // If the event would be firing correctly, we could exclude the while loop below.
```
lgrahl, line 96: The loop could only go if you really only want to enqueue a single chunk. Adding more than one chunk at once may still be a good idea.

src/content/datachannel/datatransfer/js/main.js
```diff
@@ -41,97 +44,78 @@ megsToSend.addEventListener('change', function(e) {
...
+    let count = 0;
+    while (sendProgress.value < sendProgress.max && sendChannel.bufferedAmount < CHUNK_SIZE) {
```
lgrahl, line 98: ~~The high water mark needs to be higher to allow for more than one chunk at once. With the current code, you will have the effect that one chunk is in the send pipe and no further chunk can be enqueued before the send pipe is completely dry (because of the `bufferedAmount < CHUNK_SIZE` check). This will result in bad throughput.~~ I misinterpreted - the check is fine.

Additionally, the loop should `setTimeout(..., 0)` after a few chunks to prevent a race condition with the underlying transport (this is a bug in all implementations at the moment, see [this comment](https://bugs.chromium.org/p/chromium/issues/detail?id=878682#c24)).

src/content/datachannel/datatransfer/js/main.js
```diff
@@ -7,6 +7,8 @@
  */
 
 'use strict';
+const CHUNK_SIZE = 16384;
```
lgrahl, line 10: 65536 is now safe (unless you want to factor in Firefox ESR 52).

lgrahl: Or actually, forget about that! These examples are using adapter, so let's please use `pc.sctp.maxMessageSize`.

ErikHellman: That would be an implementation detail of WebRTC and not really relevant to the sample itself.

src/content/datachannel/datatransfer/js/main.js
ErikHellman: Not sure what you want me to change here?

Should I remove the loop and do a fixed number of calls to `send()` or should I change the comment?

src/content/datachannel/datatransfer/js/main.js
ErikHellman: We don't want too many calls to `send()` as that will block the main thread and stall the UI. What we could do is to figure out the largest possible chunk and then just send that and wait for `bufferedlowamount` to trigger. 

I also don't want to add a workaround like `setTimeout()` for this, as we decided on not doing that earlier.

On a higher level, this sample should probably be removed and replaced with something more relevant (the filetransfer is one example that works better). Currently we run into weird scenarios because of a very unrealistic use of the API.

src/content/datachannel/datatransfer/js/main.js
lgrahl: Huh? `maxMessageSize` is part of the specification. I'm not saying to blindly accept `maxMessageSize` but something along the line of `Math.min(pc.sctp.maxMessageSize, 2**18)` to keep the chunks relatively small. This would be good practice because it removes a magic number (which people will simply copy and not understand) and instead relies on a detection mechanism that will eventually be implemented in the browsers themselves.

src/content/datachannel/datatransfer/js/main.js
lgrahl: The comment could be misinterpreted but I don't have a good suggestion on how to change it for now.

src/content/datachannel/datatransfer/js/main.js
ErikHellman: Ok, `pc.sctp.maxMessageSize` could work for this sample. I'll update that.

src/content/datachannel/datatransfer/js/main.js
lgrahl: `setTimeout` works differently here. This is not about polling, this is about not racing with the underlying transport as that will send in parallel and `bufferedAmount` reflects the live value (which is a bug). In the worst case, this can lead to blocking the UI for the whole send operation (e.g. it will effectively send all chunks at once in a single loop).

I agree that the example is artificial but the required send mechanism would not change much.

ErikHellman: Do you mind making a PR with the suggested change? I'm not entirely sure how you would like this too look like and I feel it is better if I review the change instead of trying to guess what you want. ;)

lgrahl: Probably makes sense.

src/content/datachannel/datatransfer/js/main.js
lgrahl: I'll include that in the PR too then.

PR #1112, opened 2018-08-31
Fixes bufferedamoutnlow event not being fired (#1102)

sumnulu:
Missing break statement causing busy wait.

**ref**
https://github.com/webrtc/samples/issues/1102 and https://github.com/webrtc/samples/pull/1110

### DIFF
--- a/src/content/datachannel/datatransfer/js/main.js
+++ b/src/content/datachannel/datatransfer/js/main.js
@@ -100,6 +100,7 @@ function sendGeneratedData() {
         } else {
           sendChannel.addEventListener('bufferedamountlow', listener);
         }
+        break;
       } else {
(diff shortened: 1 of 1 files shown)


sumnulu:
I signed it!

ErikHellman:
Thanks for the PR!

While it does seem to fix the immediate issue, the problem still remains that we need a while-loop in order to send data. If the event would be firing correctly we shouldn't need a loop but only have need to register a listener for `bufferedamountlow` before we call `send()`, and have the function be triggered recursively through the callback.

If we use a loop, we risk blocking the main thread until the event is fired, which is not behaviour we want to encourage. 

For now, the only workaround when sending data like this and not blocking the main thread is to use `setTimeout()` as I do in #1110

PR #1117, opened 2018-09-04
Datatransfer example workarounds & maxMessageSize

lgrahl:
Updates the datatransfer example to use `maxMessageSize`, a couple of bug fixes and adds a workaround to prevent racing with the underlying transport.

(As discussed in #1110)


lgrahl:
The other examples should probably be updated too but I don't have time for that atm.

ErikHellman (approved):
Looks good. Let's start with this and hopefully we can remove the `setTimeout()` once the bug is fixed. :)

Thanks!

Issue #1125, opened 2018-09-19
Number of audio tracks is empty after changing the value in create-offer sample


KaptenJansson:
# Please read first!
Please use [discuss-webrtc](https://groups.google.com/forum/#!forum/discuss-webrtc) for general technical discussions and questions.

- [x] I have provided steps to reproduce
- [x] I have provided browser name and version
- [x] I have provided a link to the sample here or a modified version thereof

**Note: If the checkboxes above are not checked (which you do after the issue is posted), the issue will be closed.**

## Browser affected
Chrome version 69.0.3497.92 (Official Build) (64-bit)

## Description
Number of audio tracks is empty after changing the value

## Steps to reproduce
Browse to https://webrtc.github.io/samples/src/content/peerconnection/create-offer/
Move the Number of audio tracks slider

## Expected results
Display the number of audio tracks selected

## Actual results
Changes from the default 1 to empty string, the span element is empty.

https://github.com/webrtc/samples/blob/gh-pages/src/content/peerconnection/create-offer/js/main.js#L23 this in `this.value` is the wrong scope, its window rather than the event object from the onchange event.

PR #1127, opened 2018-09-20
Fix for #1125

ErikHellman:
- Use innerText instead of textContent
- Use async/await instead of promises
- Fix potential adding of more tracks than requested


samdutton (approved):
LGTM

Issue #1134, opened 2018-10-08
Failed to execute 'addTrack' on 'RTCPeerConnection'


vkkkumar:
Pre-Condition:
–-enable-blink-features=RTCUnifiedPlanByDefault to Chrome.
Add atleast 2 video and 2 audio devices to the device under test

Steps to reproduce:
1.Open https://webrtc.github.io/samples/src/content/peerconnection/create-offer/
2.Set number of audio devices to 2
3.Click on "Offer to receive audio"
4.Click on "Create Offer"

## Browser affected
71.0.3569.0 Macbook pro

## Expected results
Unified plan SDP is available in the logs (at least two values with m=) [Without the Flag set in the pre-condition, we should get Plan B SDP, only one "m" value]

## Actual results

No values are listed in the text box below. Its always empty when we choose audio track =2. 
Upon opening the console there are errors shown. 

Here comes the error:
Uncaught (in promise) DOMException: Failed to execute 'addTrack' on 'RTCPeerConnection': A sender already exists for the track.
    at RTCPeerConnection.window.RTCPeerConnection.addTrack (https://webrtc.github.io/adapter/adapter-latest.js:3161:33)
    at HTMLButtonElement.createOffer (https://webrtc.github.io/samples/src/content/peerconnection/create-offer/js/main.js:34:20)

<img width="540" alt="screen shot 2018-10-08 at 2 23 13 pm" src="https://user-images.githubusercontent.com/34342420/46609188-6ab11580-cb07-11e8-8d5d-20e8618fbf1c.png">
<img width="1217" alt="screen shot 2018-10-08 at 2 14 27 pm" src="https://user-images.githubusercontent.com/34342420/46609189-6ab11580-cb07-11e8-85bc-b2748c33d53e.png">
<img width="1221" alt="screen shot 2018-10-08 at 2 14 16 pm" src="https://user-images.githubusercontent.com/34342420/46609190-6b49ac00-cb07-11e8-8555-3a19cf2f9170.png">

vkkkumar:
@KaptenJansson @rantonysamy

Orphis:
The sample is trying to add the same track multiple times to the same PeerConnection object, which is not possible as per the spec: https://w3c.github.io/webrtc-pc/#dom-rtcpeerconnection-addtrack

> 6. Let senders be the result of executing the CollectSenders algorithm. If an RTCRtpSender for track already exists in senders, throw an InvalidAccessError.

[...]

PR #1137, opened 2018-10-11
Fix createOffer() sample when using multiple audio tracks

ErikHellman:
Fix for #1134

Ping @vkkkumar

### DIFF
--- a/src/content/peerconnection/create-offer/js/main.js
+++ b/src/content/peerconnection/create-offer/js/main.js
@@ -23,13 +23,14 @@ numAudioTracksInput.addEventListener('change', e => numAudioTracksDisplay.innerT
 
 async function createOffer() {
   outputTextarea.value = '';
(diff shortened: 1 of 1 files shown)


src/content/peerconnection/create-offer/js/main.js
```diff
@@ -53,10 +52,11 @@ function createOffer() {
...
 
-  pc.createOffer(offerOptions)
-    .then(desc => {
-      pc.setLocalDescription(desc);
-      outputTextarea.value = desc.sdp;
-    })
-    .catch(error => outputTextarea.value = `Failed to createOffer: ${error}`);
+  try {
+    const desc = await pc.createOffer(offerOptions);
+    pc.setLocalDescription(desc);
```
Orphis: setLocalDescription() returns a promise. Shouldn't you await it as well or does it not matter for this sample?

Orphis (approved):

PR #1144, opened 2018-10-30
Updated screen capturing sample with Chrome support.

ErikHellman:
This updates the screen sharing sample to support both Chrome and Firefox. 
It will also record the capture stream for later viewing.

### DIFF
--- a/src/content/getusermedia/getdisplaymedia/index.html
+++ b/src/content/getusermedia/getdisplaymedia/index.html
@@ -25,25 +25,26 @@
   <link rel="icon" sizes="192x192" href="../../../images/webrtc-icon-192x192.png">
   <link href="//fonts.googleapis.com/css?family=Roboto:300,400,500,700" rel="stylesheet" type="text/css">
   <link rel="stylesheet" href="../../../css/main.css">
+  <link rel="stylesheet" href="css/styles.css">
+  <script src="https://unpkg.com/@webcomponents/webcomponentsjs/webcomponents-loader.js"></script>
 
 </head>
 
 <body>
 
   <div id="container">
-    <h1><a href="//webrtc.github.io/samples/" title="WebRTC samples homepage">WebRTC samples</a> <span>getDisplayMedia</span></h1>
+    <h1><a href="//webrtc.github.io/samples/" title="WebRTC samples homepage">WebRTC samples</a> <span>Demo of getDisplayMedia and screen recording</span></h1>
 
-    <video id="gum-local" autoplay playsinline></video>
+    <h4>Screen capturing is currently an experimental feature which is onyl supported by latest Chrome and Firefox!</h4>
 
-    <div id="errorMsg"></div>
+    <screen-sharing></screen-sharing>
 
-    <p>Display the screensharing stream from <code>getDisplayMedia()</code> in a video element.</p>
+    <p>Display the screensharing stream from <code>getDisplayMedia()</code> in a video element and record the stream.</p>
 
     <a href="https://github.com/webrtc/samples/tree/gh-pages/src/content/getusermedia/getdisplaymedia" title="View source for this page on GitHub" id="viewSource">View source on GitHub</a>
   </div>
 
-  <script src="https://webrtc.github.io/adapter/adapter-latest.js"></script>
-  <script src="js/main.js"></script>
+  <script src="js/main.js" type="module"></script>
 
   <script src="../../../js/lib/ga.js"></script>
 
--- a/src/content/getusermedia/getdisplaymedia/js/main.js
+++ b/src/content/getusermedia/getdisplaymedia/js/main.js
@@ -25,7 +25,6 @@ class ScreenSharing extends LitElement {
       enableStartCapture: Boolean,
       enableStopCapture: Boolean,
       enableDownloadRecording: Boolean,
-      buttons: Object,
       stream: {
         type: {
           fromAttribute: input => input
--- a/src/content/getusermedia/getdisplaymedia/js/main.js
+++ b/src/content/getusermedia/getdisplaymedia/js/main.js
@@ -67,9 +67,9 @@ video {
 
   async _startCapturing(e) {
     console.log('Start capturing.');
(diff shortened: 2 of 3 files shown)


src/content/getusermedia/getdisplaymedia/js/main.js
```diff
@@ -6,40 +6,127 @@
...
+    this.recording = [];
+    this.mediaRecorder = null;
+  }
 
-function handleSuccess(stream) {
-  const video = document.querySelector('video');
-  video.srcObject = stream;
+  static get properties() {
+    return {
+      buttons: Object,
```
kenchris: this wont work in Lit... has to use type: Object... but then Object is a useless serializer

src/content/getusermedia/getdisplaymedia/js/main.js
```diff
@@ -6,40 +6,127 @@
...
+  display: block;
+  padding: 10px;
+  width: 100%;
+  height: 100%;
 }
-
-function handleError(error) {
-  errorMsg(`getDisplayMedia error: ${error.name}`, error);
+video {
+    --width: 100%;
```
kenchris: maybe find a better name that just --width

src/content/getusermedia/getdisplaymedia/js/main.js
```diff
@@ -6,40 +6,127 @@
...
 
-function errorMsg(msg, error) {
-  const errorElement = document.querySelector('#errorMsg');
-  errorElement.innerHTML += `<p>${msg}</p>`;
-  if (typeof error !== 'undefined') {
-    console.error(error);
+</style>
+<video playsinline autoplay muted .srcObject="${this.stream}"></video>
+<div>
+<button ?disabled="${!this.buttons.startCapturing}" @click="${async e => this._startCapturing(e)}">Start screen capture</button>
```
kenchris: why async?

src/content/getusermedia/getdisplaymedia/js/main.js
ErikHellman: Yeah, this was the best I could come up with. I think I'll clean it up with three separate booleans instead.

src/content/getusermedia/getdisplaymedia/index.html
```diff
@@ -25,25 +25,26 @@
   <link rel="icon" sizes="192x192" href="../../../images/webrtc-icon-192x192.png">
   <link href="//fonts.googleapis.com/css?family=Roboto:300,400,500,700" rel="stylesheet" type="text/css">
   <link rel="stylesheet" href="../../../css/main.css">
+  <link rel="stylesheet" href="css/styles.css">
+  <script src="https://unpkg.com/@webcomponents/webcomponentsjs/webcomponents-loader.js"></script>
```
jan-ivar, line 29: This seems like an unnecessary dependency that shouldn't have been merged IMHO. Doesn't even load on Edge anymore.

I'll file an issue to revert it.

src/content/getusermedia/getdisplaymedia/index.html
```diff
@@ -25,25 +25,26 @@
...
 </head>
 
 <body>
 
   <div id="container">
-    <h1><a href="//webrtc.github.io/samples/" title="WebRTC samples homepage">WebRTC samples</a> <span>getDisplayMedia</span></h1>
+    <h1><a href="//webrtc.github.io/samples/" title="WebRTC samples homepage">WebRTC samples</a> <span>Demo of getDisplayMedia and screen recording</span></h1>
 
-    <video id="gum-local" autoplay playsinline></video>
+    <h4>Screen capturing is currently an experimental feature which is onyl supported by latest Chrome and Firefox!</h4>
```
jan-ivar, line 38: Typo and factually incorrect, since Edge was the [first browser to support getDisplayMedia](https://blog.mozilla.org/webrtc/getdisplaymedia-now-available-in-adapter-js/). Safari is also developing it.

ErikHellman: Thanks. Will fix!

src/content/getusermedia/getdisplaymedia/index.html
ErikHellman: We've had several disucssions around barebone vs contexualised samples before. The feedback I got from other web develoeprs were that the contexualised samples are more appreciated, as it shows how to use it in a real application. 

If it doesn't work in Edge we'll fix that. It's the only browser I haven't been able to test in.

fippo:
with the recent changes in canary this sample open the camera in Chrome (possibly safari as well).
There was a reason this sample (or any sample) is using adapter.js *and* its browser detection.

src/content/getusermedia/getdisplaymedia/js/main.js
```diff
@@ -6,40 +6,127 @@
...
+      this.stream.getTracks().forEach(track => track.stop());
+      this.stream = null;
+    });
+    this.mediaRecorder = new MediaRecorder(this.stream, {mimeType: 'video/webm'});
+    this.mediaRecorder.addEventListener('dataavailable', event => {
+      if (event.data && event.data.size > 0) {
+        this.recording.push(event.data);
+      }
+    });
+    this.mediaRecorder.start(10);
```
Pehrsons, line 98: What's the rationale behind a 10ms timeslice?

Your use of the timeslice, namely storing the blobs in memory, is functionally equivalent to not having a timeslice at all.

Not having a timeslice also adds some benefits. It lets the UA optimize how it gathers data into the blobs, e.g., by storing the data in a blob on disk instead of in memory. It also simplifies your code since you don't need the `recording` array and the bookkeeping that comes with it.

ErikHellman: That part is a copy from another sample. Since this sample is going away (from the massive negative feedback I got so far) I won't update this for now, but please have a look at the MediaRecorder sample and suggest an improvement there. Thanks!

Pehrsons: Gotcha. Filed issue #1153.

PR #1177, opened 2019-02-06
Add error.message and error.name to a few samples

KaptenJansson:
**Description**
Add more details about the gUM error.

I did not do all samples due to them having all different formats etc. An overhaul in general would be nice and perhaps a shared error handler rather than having it in each separate sample.

**Purpose**
Easier to debug.


KaptenJansson:
Merging due to minor fixes.

PR #1206, opened 2019-07-23
Fix sample

dsanders11:
Firefox (as of July 2019) will throw an error if `captureStream()` is called on a canvas without the context being touched.

Chrome (as of July 2019) seems to have page load hang indefinitely if `captureStream()` is called before the `window.onload` event.

I was not able to get this sample to work in either Chrome or Firefox without these changes.

@alvestrand

### DIFF
--- a/src/content/capture/canvas-filter/js/main.js
+++ b/src/content/capture/canvas-filter/js/main.js
@@ -8,42 +8,43 @@
 
 'use strict';
 
-const source = document.querySelector('#source');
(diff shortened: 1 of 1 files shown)


alvestrand:
The main track and the patch seem to have diverged (or perhaps converged) to the point where I can't see how to apply this any more - some changes seem to be differences of style, like using window.onload rather than {}(), so I'm not sure which changes are good changes any more.

Closing this for now - if the error still exists, please open a bug.

PR #1211, opened 2019-07-30
trickle-ice: use native or polyfilled attributes and ignore empty candidate

fippo:
Uses the native (or polyfilled) candidate attributes instead of parsing the candidate string and ignores the '' ice candidate emitted by Firefox.


fippo:
@juberti PTAL since this removes a lot of code you wrote five summers ago :-)

juberti (approved):

PR #1225, opened 2019-09-22
Bit Rate Control Panel on videos

rdxmiles:
**Description**
Bit Rate Control Panel on videos is used to control speed of videos that are running remotely at the same time.
**Purpose**
This control will help the user to control video speed to be shown remotely like user can increase and decrease speed and window on the right side will clearly see the changes made on left window i.e will when a user wants to slow down the video speed for another user at the same time.


rdxmiles:
@googlebot I signed it!

PR #1229, opened 2019-10-01
Video bitrate control

rdxmiles:
Description
Bit Rate Control Panel on videos helps to control video speed on peer connections at the same time.

Purpose
Bit Rate Control Panel helps to control video speed so that user can manage on what speed the video is to be shown to peer connections like when user wants to slow speed of video at some instance the changes can directly be shown to peer connections.


alvestrand:
This feature seems to show off a feature of video playback, and not a feature of WebRTC.
Playback at a slower rate than the sending rate is not a feature WebRTC is supposed to support out of the box; I'm not sure what is supposed to happen here.

Closing PR.

PR #1238, opened 2019-11-20
recording sample: add logging

fippo:
Adds logging to make https://bugs.chromium.org/p/chromium/issues/detail?id=1024120 more visible and logs the recorded blobs at the right time.

Drive-by: eslint --fix as separate commit


fippo:
the lint failures look pretty unrelated

@guidou @armax00 PTAL. Makes it a bit easier to bisect ;-)

guidou (approved):

PR #1244, opened 2019-11-27
connectionstatechange->loadedmetadata delay

fippo:
STR:
- checkout the branch
- run the pc1 sample
- observe the console logs for "Time between connectionStatechange" (which should be close to the bottom

My expectation is that this is minor but it turns out to be in the order of ~150ms and the RTT should be close to 0. playoutDelayHint seems to have a small impact on that but even setting it to 0 only reduces the delay between the events to ~100ms.

Arguably we should be able to listen to track.onunmute but... ;-)

### DIFF
--- a/src/content/peerconnection/pc1/js/main.js
+++ b/src/content/peerconnection/pc1/js/main.js
@@ -25,8 +25,10 @@ localVideo.addEventListener('loadedmetadata', function() {
   console.log(`Local video videoWidth: ${this.videoWidth}px,  videoHeight: ${this.videoHeight}px`);
 });
 
+let connectedTime;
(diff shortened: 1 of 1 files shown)


fippo:
@alvestrand @henbos any idea?

alvestrand:
Yes, it takes a helluvalot of time. The first video packets have to flow before you get loadedmetadata, and the number of layers of abstraction involved makes my head hurt.
Would be nice to make this delay a metric and track it; that requires filling a chromium bug, I think.

alvestrand:
Closing because more appropriate as a Chromium bug. Nice example.

PR #1246, opened 2019-12-09
munge-sdp: restore CRLF

fippo:
restores the CRLF in munge-sdp which apparently gets lost when putting the text into the textarea. Found while testing
  https://chromium-review.googlesource.com/c/chromium/src/+/1953635
this will make it easier to detect actual changes (if that ever happens)


alvestrand:
would like to have a test on this.
note - if there is ever a textarea that preserves \r, the result will be \r\r\n; this may fail.

fippo:
updated to be more robust which is a good thing to show anyway.
This is covered by a test fortunately

PR #1259, opened 2020-02-18
audio/bandwidth: show headerBytes sent

fippo:
which visualizes the overhead the RTP header has on audio packets.
Screenshot:
image
@henbos as master-of-stats PTAL

cc @eladalon1983 -- shows *why* one wants to adapt the ptime to reduce the overhead


henbos (approved):
Sure!

PR #1266, opened 2020-03-10
support ES6 in Internet Explorer

shivangi-deshpande:
**Description**

**Purpose**


alvestrand:
Closing because of no CLA. Including files from other sites is generally not a good idea either.

PR #1268, opened 2020-03-27
Add an insertable streams demonstrator for encryption.

alvestrand:
**Description**
Demonstrate JS-layer encryption of a video call.

This consists of sending a video track to a receiver and a
"middlebox". The receiver has the key, the middlebox doesn't.

The receiver can see the video; the middlebox can only see it
if the key is blank.

The encryption itself is silly.

**Purpose**
Demonstrate how to set up an Insertable Streams usage.

### DIFF
--- a/src/content/peerconnection/endtoend-encryption/index.html
+++ b/src/content/peerconnection/endtoend-encryption/index.html
@@ -1,0 +1,75 @@
+<!DOCTYPE html>
+<!--
+ *  Copyright (c) 2015 The WebRTC project authors. All Rights Reserved.
+ *
+ *  Use of this source code is governed by a BSD-style license
+ *  that can be found in the LICENSE file in the root of the source
+ *  tree.
+-->
+<html>
+<head>
+
+    <meta charset="utf-8">
+    <meta name="description" content="WebRTC code samples">
+    <meta name="viewport" content="width=device-width, user-scalable=yes, initial-scale=1, maximum-scale=1">
+    <meta itemprop="description" content="Client-side WebRTC code samples">
+    <meta itemprop="image" content="../../../images/webrtc-icon-192x192.png">
+    <meta itemprop="name" content="WebRTC code samples">
+    <meta name="mobile-web-app-capable" content="yes">
+    <meta id="theme-color" name="theme-color" content="#ffffff">
+
+    <base target="_blank">
+
+    <title>Peer connection end to end encryption</title>
+
+    <link rel="icon" sizes="192x192" href="../../../images/webrtc-icon-192x192.png">
+    <link href="//fonts.googleapis.com/css?family=Roboto:300,400,500,700" rel="stylesheet" type="text/css">
+    <link rel="stylesheet" href="../../../css/main.css"/>
+    <link rel="stylesheet" href="css/main.css"/>
+
+</head>
+
+<body>
+
+<div id="container">
+
+    <h1><a href="//webrtc.github.io/samples/" title="WebRTC samples homepage">WebRTC samples</a> <span>Peer connection end to end encryption</span>
+    </h1>
+    <span id="banner"></span>
+    <p>
+      <h2>Sender and receiver</h2>
+      <div id="videos">
+        Sender and receiver<br>
+        <video id="video1" playsinline autoplay muted></video>
+        <video id="video2" playsinline autoplay></video>
+      </div>
+      <br>
+      Crypto key: <input type="text" id="crypto-key">
+    </p>
+    <p>
+      <h2>Middlebox</h2>
+      <div id="monitor">
+        <video id="video-monitor" playsinline autoplay></video>
+      </div>
+    </p>
+
+    <div id="buttons">
+        <button id="start">Start</button>
+        <button id="call" disabled>Call</button>
+        <button id="hangup" disabled>Hang Up</button>
+    </div>
+    <div id="status"></div>
+
+    <a href="https://github.com/webrtc/samples/tree/gh-pages/src/content/peerconnection/multiple-relay"
+       title="View source for this page on GitHub" id="viewSource">View source on GitHub</a>
+
+</div>
+
+<script src="https://webrtc.github.io/adapter/adapter-latest.js"></script>
+<script src="js/videopipe.js"></script>
+<script src="js/main.js" async></script>
+
+<script src="../../../js/lib/ga.js"></script>
+
+</body>
+</html>
--- a/src/content/peerconnection/endtoend-encryption/js/main.js
+++ b/src/content/peerconnection/endtoend-encryption/js/main.js
@@ -102,7 +102,8 @@ function encodeFunction(chunk, controller) {
     const newView = new DataView(newData);
 
     for (let i = 0; i < chunk.data.byteLength; ++i) {
-      newView.setInt8(i, ~view.getInt8(i));
+      const keyByte = currentCryptoKey.charCodeAt(i % currentCryptoKey.length);
+      newView.setInt8(i, view.getInt8(i) ^ keyByte);
     }
     // Append checksum
     newView.setUint32(chunk.data.byteLength, 0xDEADBEEF);
@@ -123,7 +124,8 @@ function decodeFunction(chunk, controller) {
     const newData = new ArrayBuffer(chunk.data.byteLength - 4);
     const newView = new DataView(newData);
     for (let i = 0; i < chunk.data.byteLength - 4; ++i) {
-      newView.setInt8(i, ~view.getInt8(i));
+      const keyByte = currentCryptoKey.charCodeAt(i % currentCryptoKey.length);
+      newView.setInt8(i, view.getInt8(i) ^ keyByte);
     }
     chunk.data = newData;
   }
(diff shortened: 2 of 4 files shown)


src/content/peerconnection/endtoend-encryption/js/videopipe.js
```diff
@@ -0,0 +1,102 @@
...
+    }
+  };
+  pc2.ontrack = function(e) {
+    if (receiveTransform) {
+      const transform = new TransformStream({
+        start() {},
+        flush() {},
+        transform: receiveTransform
+      });
+      const receiverStreams = pc2.getReceivers()[0].createEncodedVideoStreams();
```
murillo128: wouldn't it be better to just use the receiver from the event?
```
const receiverStreams= e.receiver.createEncodedVideoStreams();
```

src/content/peerconnection/endtoend-encryption/js/videopipe.js
```diff
@@ -0,0 +1,102 @@
...
+
+  const sender = pc1.addTrack(stream.getVideoTracks()[0], stream);
+  if (sendTransform) {
+    const senderStreams = sender.createEncodedVideoStreams();
+    const senderTransformStream = new TransformStream({
+      start() {},
+      flush() {},
+      transform: sendTransform
+    });
+    senderStreams.readableStream
```
murillo128, line 40: would it be interesting to do the encrypt/decrypt on a worker thread by transfering the streams to it?

src/content/peerconnection/endtoend-encryption/js/main.js
```diff
@@ -0,0 +1,136 @@
...
+    const view = new DataView(chunk.data);
+    const checksum = view.getUint32(chunk.data.byteLength - 4);
+    if (checksum != 0xDEADBEEF) {
+      console.log('Corrupted frame received');
+      console.log(checksum.toString(16));
+    }
+    const newData = new ArrayBuffer(chunk.data.byteLength - 4);
+    const newView = new DataView(newData);
+    for (let i = 0; i < chunk.data.byteLength - 4; ++i) {
+      newView.setInt8(i, ~view.getInt8(i));
```
murillo128: same here:
```
newView.setInt8(i, view.getInt8(i) ^ currentCryptoKey.charAt(i % currentCryptoKey.length));
```

src/content/peerconnection/endtoend-encryption/js/main.js
```diff
@@ -0,0 +1,136 @@
...
+
+function encodeFunction(chunk, controller) {
+  if (currentCryptoKey) {
+    const view = new DataView(chunk.data);
+    // Any length that is needed can be used for the new buffer.
+    const newData = new ArrayBuffer(chunk.data.byteLength + 4);
+    const newView = new DataView(newData);
+
+    for (let i = 0; i < chunk.data.byteLength; ++i) {
+      newView.setInt8(i, ~view.getInt8(i));
```
murillo128: how about xor'ing with the key?
```
newView.setInt8(i, view.getInt8(i) ^ currentCryptoKey.charAt(i % currentCryptoKey.length));
```

murillo128 (commented):
Nice work! Just some comments for improvement.

src/content/peerconnection/endtoend-encryption/js/videopipe.js
alvestrand: Yep. That's a later demo.
This doesn't work yet (the CL for making frames transferable is still in review), so I didn't include it at this time.

src/content/peerconnection/endtoend-encryption/js/main.js
alvestrand: Done!

src/content/peerconnection/endtoend-encryption/js/main.js
alvestrand: Done!

src/content/peerconnection/endtoend-encryption/js/videopipe.js
alvestrand: Done!

src/content/peerconnection/endtoend-encryption/index.html
```diff
@@ -0,0 +1,75 @@
+<!DOCTYPE html>
+<!--
+ *  Copyright (c) 2015 The WebRTC project authors. All Rights Reserved.
```
fippo, line 3: nit: update the year

src/content/peerconnection/endtoend-encryption/index.html
```diff
@@ -0,0 +1,75 @@
...
+    </p>
+
+    <div id="buttons">
+        <button id="start">Start</button>
+        <button id="call" disabled>Call</button>
+        <button id="hangup" disabled>Hang Up</button>
+    </div>
+    <div id="status"></div>
+
+    <a href="https://github.com/webrtc/samples/tree/gh-pages/src/content/peerconnection/multiple-relay"
```
fippo, line 63: nit: fix the url

lgrahl:
I really like this. Can you point me to the spec? I have some suggestions.

Nits: `start` and `flush` of `TransformStream` are optional.

If I may add a suggestion: We know that even if there's a warning in caps that people will use this example in real world projects. So, how about we drop in [tweetnacl-js](https://github.com/dchest/tweetnacl-js) and do non-silly encryption? :slightly_smiling_face:

murillo128:
version with iv + aes gcm using web crypto:

```
	let counter = 0;
	const roomKey = await getRoomKey(roomId,secret);
	async function encrypt(chunk, controller) {
		try {
			//Get iv
			const iv = new ArrayBuffer(4);
			//Create view, inc counter and set it
			new DataView(iv).setUint32(0,counter <65535 ? counter++ : counter=0);
			//Encrypt
			const ciphertext = await window.crypto.subtle.encrypt(
				{
					name: "AES-GCM",
					iv: iv
				},
				roomKey,
				chunk.data
			);
			//Crate new encoded data and allocate size for iv
			chunk.data = new ArrayBuffer(ciphertext.byteLength + 4);
			//Get uint view
			const data = new Uint8Array(chunk.data);
			//Copy iv
			data.set(new Uint8Array(iv),0);
			//Copy encruptedframe
			data.set(new Uint8Array(ciphertext),4);
			//Write
			controller.enqueue(chunk);
		} catch(e) {
		}
	}

	async function decrypt(chunk, controller) {
		try {
			//decrypt
			chunk.data =  await window.crypto.subtle.decrypt(
				{
				  name: "AES-GCM",
				  iv: new Uint8Array(chunk.data,0,4)
				},
				roomKey,
				new Uint8Array(chunk.data,4,chunk.data.byteLength - 4)
			);
			//Write
			controller.enqueue(chunk);
		} catch(e) {
		}
	}
```

lgrahl:
Now I'm even more scared. :sweat_smile: I'm not well versed with AES-GCM, so, for the record, I have no idea if this is safe in the end but the IV part (which I think is the nonce in AES-GCM) doesn't look very random or like a safe nonce counter to me. Please, no unsafe examples.

murillo128:
i think you should review aes-gcm then or how srtp works.

lgrahl:
That is exactly **not** the answer I want to have in an example. Please, provide comments and references instead of "RTFM".

Please explain the ternary for the counter. IIRC, the nonce is 12 bytes, so the full 12 bytes should be used when counting. Adding a random part is probably an option, too, if the key may be reused.

murillo128:
It's under work, just wait a bit ;)

alvestrand:
lgrahl: https://github.com/alvestrand/webrtc-media-streams/blob/master/explainer.md is the most up to date spec (the spec in that repo is an older version).

alvestrand:
I think I'll leave the "encryption" as totally bogus in this iteration. A subtly broken crypto is more dangerous than a totally bogus one - and I'm not competent to make sure something we cook up for the occasion isn't subtly broken.

guidou:
lgtm

lgrahl:
> A JS example is available in the webrtc/sample repository. A better example, closer to the SFrame implementation was provided by Sergio, but eventually not added, because an invited expert does not understand it, did not want to check against the specs, and could not wait for this blog post. oh well.

http://webrtcbydralex.com/index.php/2020/03/30/secure-frames-sframes-end-to-end-media-encryption-with-webrtc-now-in-chrome/

Hi Alex, I'll say it out loud: You're poison to the whole WebRTC community and one of the reasons why people stay the f*** away from it.

agouaillard-cosmo:
Lenart. The intend was not to upset your anyone, and was thought to be factual. As a courtesy and sign of good faith, I have removed the part you judged offensive, like I would have if you had privately and directly contacted me with a request to do so. Now the only written trace left is your comment above.

However entitled you are to have an opinion, I would humbly request that you respect the code of conduct in the place where you choose to post. In this thread this is it: https://webrtc.org/support/code-of-conduct

I would also like to bring to your attention the severity of defamation laws in many countries, especially when it is not slander but libel. I would advice caution and control in the future.

lgrahl:
Updated to concur with defamation laws (probably). I have to say, **you** of all persons reminding me of the CoC... wow.

PR #1269, opened 2020-03-30
endtoend-encryption: clean up and modernize code

fippo:
also use 320px wide video

### DIFF
--- a/src/content/peerconnection/endtoend-encryption/js/videopipe.js
+++ b/src/content/peerconnection/endtoend-encryption/js/videopipe.js
@@ -21,32 +21,15 @@
 //
 'use strict';
 
-function errorHandler(context) {
-  return function(error) {
-    trace('Failure in ' + context + ': ' + error.toString);
-  };
-}
-
-// eslint-disable-next-line no-unused-vars
-function successHandler(context) {
-  return function() {
-    trace('Success in ' + context);
-  };
-}
-
-function noAction() {
-}
-
-
 function VideoPipe(stream, sendTransform, receiveTransform, handler) {
-  const pc1 = new RTCPeerConnection({
+  this.pc1 = new RTCPeerConnection({
     forceEncodedVideoInsertableStreams: !!sendTransform
   });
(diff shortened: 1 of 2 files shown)


src/content/peerconnection/endtoend-encryption/js/videopipe.js
```diff
@@ -58,44 +41,38 @@ function VideoPipe(stream, sendTransform, receiveTransform, handler) {
...
+  this.pc2.ontrack = e => {
     if (receiveTransform) {
       const transform = new TransformStream({
         start() {},
         flush() {},
         transform: receiveTransform
       });
-      const receiverStreams = e.receiver.createEncodedVideoStreams();
+      const {receiver} = e.transceiver;
+      const receiverStreams = receiver.createEncodedVideoStreams();
```
alvestrand: any particular reason why you want to go via the transceiver?

src/content/peerconnection/endtoend-encryption/js/videopipe.js
```diff
@@ -58,44 +41,38 @@ function VideoPipe(stream, sendTransform, receiveTransform, handler) {
...
-    if (event.candidate) {
-      pc1.addIceCandidate(new RTCIceCandidate(event.candidate),
-          noAction, errorHandler('AddIceCandidate'));
-    }
-  };
-  pc2.ontrack = function(e) {
+  this.pc2.ontrack = e => {
     if (receiveTransform) {
       const transform = new TransformStream({
         start() {},
```
lgrahl, line 47: Nit: These are optional (`start` and `flush`)

fippo: wrong PR for nitting on that :-)

src/content/peerconnection/endtoend-encryption/js/videopipe.js
fippo: in general because the transceiver is better for matching things (e.g. with mid) than the receiver but that isn't relevant to this example so removed it.

PR #1270, opened 2020-03-30
make it work when there is an audio track + fix style nits

fippo:
i assume you're going to squash the PR before merging so don't care about the commit messages :-)


fippo:
argh...

PR #1272, opened 2020-03-30
e2ee: fix minor issues

fippo:
see individual commits

### DIFF
--- a/src/content/peerconnection/endtoend-encryption/index.html
+++ b/src/content/peerconnection/endtoend-encryption/index.html
@@ -1,6 +1,6 @@
 <!DOCTYPE html>
 <!--
- *  Copyright (c) 2015 The WebRTC project authors. All Rights Reserved.
+ *  Copyright (c) 2020 The WebRTC project authors. All Rights Reserved.
  *
  *  Use of this source code is governed by a BSD-style license
  *  that can be found in the LICENSE file in the root of the source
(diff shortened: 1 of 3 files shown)


src/content/peerconnection/endtoend-encryption/css/main.css
```diff
@@ -1,5 +1,5 @@
 /*
- *  Copyright (c) 2015 The WebRTC project authors. All Rights Reserved.
+ *  Copyright (c) 2020 The WebRTC project authors. All Rights Reserved.
```
alvestrand: the main.css file is copied from another source, so 2015 is right

src/content/peerconnection/endtoend-encryption/index.html
```diff
@@ -1,6 +1,6 @@
 <!DOCTYPE html>
 <!--
- *  Copyright (c) 2015 The WebRTC project authors. All Rights Reserved.
+ *  Copyright (c) 2020 The WebRTC project authors. All Rights Reserved.
```
alvestrand, line 3: this one is changed enough that it's reasonable to call it new, even though it's also a branch from another file.

alvestrand (commented):
the fork history of the 2 major-changed files can be left behind, but the css is unchanged from its origin.

src/content/peerconnection/endtoend-encryption/css/main.css
fippo: we should just bring those common things into main.css. Some day...

fippo: i was tempted to change the css not to include audio but then I assume we'll add audio shortly :-)

alvestrand:
can you remove the CSS file change, so that I can merge the rest of the year fixes?

src/content/peerconnection/endtoend-encryption/css/main.css
fippo: this is obsolete but github did not auto-resolve

fippo:
it was already removed, github didn't autoresolve the conversation :-|

PR #1274, opened 2020-03-31
e2ee: make fun of the middlebox

fippo:
the visual results of not encrypting the first 10 bytes of the VP8 payload
  https://tools.ietf.org/html/rfc6386#section-9.1
are much more interesting since the decoder gets enough stuff to allow it
to try decoding a bit, resulting in weird images for the middlebox.

### DIFF
--- a/src/content/peerconnection/endtoend-encryption/js/main.js
+++ b/src/content/peerconnection/endtoend-encryption/js/main.js
@@ -30,6 +30,7 @@ cryptoKey.addEventListener('change', setCryptoKey);
 let startToMiddle;
 let startToEnd;
 let currentCryptoKey;
+let currentKeyIdentifier = 0;
 
 let localStream;
 // eslint-disable-next-line no-unused-vars
@@ -98,7 +99,7 @@ function encodeFunction(chunk, controller) {
   if (currentCryptoKey) {
     const view = new DataView(chunk.data);
     // Any length that is needed can be used for the new buffer.
-    const newData = new ArrayBuffer(chunk.data.byteLength + 4);
+    const newData = new ArrayBuffer(chunk.data.byteLength + 5);
     const newView = new DataView(newData);
 
     // Do not encrypt the first 10 bytes of the payload. For VP8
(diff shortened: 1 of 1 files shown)


fippo:
@alvestrand for your amusement ;-)

fippo:
this sometimes results in pixelation on the receiving end after stopping to decrypt. I suspect what happens there is that a frame which was encrypted and has a checksum is, because the currentCryptoKey is empty, no longer run through the correct parts of the decode function.

Arguably a minor issue but maybe the decode function should look for the "checksum" instead?

alvestrand:
I've seen that changing the key sometimes leads to pixelation.
I suspect that what's happening is that there's a frame that's in flight while I change the crypto key, leading to the frame being decrypted with the wrong key.

In a real cryptosystem the solution to this is a) to make sure the checksum is wrong when decrypted with the wrong key, so that it's just frame loss, and b) to synchronize the changing of the key to the frame.

alvestrand:
This change should also make the example work without the GFD, right?

fippo:
I don't see how its broken without GFD. Do you have STR? ;-)

Added a second commit to deal a bit better with in-flight frames by dropping them. Still a super simplistic way of dealing with things but the whole scenario of changing keys on the fly is a bit unrealistic anyway. It does avoid pixelation and just causes short freezes which are then resolved by nack/pli (presumably)

src/content/peerconnection/endtoend-encryption/js/main.js
```diff
@@ -114,20 +120,26 @@ function encodeFunction(chunk, controller) {
 }
 
 function decodeFunction(chunk, controller) {
+  const view = new DataView(chunk.data);
+  const checksum = view.getUint32(chunk.data.byteLength - 4);
   if (currentCryptoKey) {
-    const view = new DataView(chunk.data);
-    const checksum = view.getUint32(chunk.data.byteLength - 4);
-    if (checksum != 0xDEADBEEF) {
+    if (checksum !== 0xDEADBEEF) {
```
alvestrand, line 129: To make it more realistic: Have a counter "currentCryptoKeyGeneration" that is incremented every time you call setCryptoKey and add that to the checksum (0xDEADBEEF + currentCryptoKeyGeneration).
That way, frames encrypted with the wrong key will be discarded reliably, and changes from encrypt-with-key-A to encrypt-with-key-B will be handled, instead of just changes from encrypt-with-key-A to no-encrypt and vice versa.

alvestrand:
actually changing keys on the fly is not only realistic, it's a requirement of group calling :-(

thanks for the update!

If you don't have GFD, and encrypt the whole frame, and set the key before pressing the "call" button, the receiver will see a black frame.

src/content/peerconnection/endtoend-encryption/js/main.js
fippo: added as a separate commit. Single byte instead of the four bytes MKI from SRTP ;-)
The console.log shows up nicely when changing the key

alvestrand:
Merging this. I'm going to add a toggle to encrypt from the start or not, so that we can verify that the GFD matters for keyframe detection.

PR #1276, opened 2020-03-31
Video analyzer

alvestrand:
**Description**

Adds an example of using insertable streams to perform analysis of incoming video.

**Purpose**

Demonstrates an usage of Insertable Streams that isn't encryption.

### DIFF
--- a/src/content/peerconnection/video-analyzer/js/main.js
+++ b/src/content/peerconnection/video-analyzer/js/main.js
@@ -37,13 +37,6 @@ localVideo.addEventListener('loadedmetadata', function() {
 });
 
 remoteVideo.addEventListener('loadedmetadata', function() {
-  console.log(`Remote video videoWidth: ${this.videoWidth}px,  videoHeight: ${this.videoHeight}px`);
-});
-
-remoteVideo.addEventListener('resize', () => {
-  console.log(`Remote video size changed to ${remoteVideo.videoWidth}x${remoteVideo.videoHeight}`);
-  // We'll use the first onsize callback as an indication that video has started
-  // playing out.
   if (startTime) {
     const elapsedTime = window.performance.now() - startTime;
     console.log('Setup time: ' + elapsedTime.toFixed(3) + 'ms');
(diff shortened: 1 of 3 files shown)


src/content/peerconnection/video-analyzer/js/main.js
```diff
@@ -0,0 +1,232 @@
...
+  console.log(`Local video videoWidth: ${this.videoWidth}px,  videoHeight: ${this.videoHeight}px`);
+});
+
+remoteVideo.addEventListener('loadedmetadata', function() {
+  console.log(`Remote video videoWidth: ${this.videoWidth}px,  videoHeight: ${this.videoHeight}px`);
+});
+
+remoteVideo.addEventListener('resize', () => {
+  console.log(`Remote video size changed to ${remoteVideo.videoWidth}x${remoteVideo.videoHeight}`);
+  // We'll use the first onsize callback as an indication that video has started
```
fippo: that is actually wrong, loadedmetadata is more correct. But I can clean that up everywhere in a different PR

src/content/peerconnection/video-analyzer/js/main.js
```diff
@@ -0,0 +1,232 @@
...
+const keyFrameSizeDisplay = document.querySelector('#keyframe-size');
+const interFrameCountDisplay = document.querySelector('#interframe-count');
+const interFrameSizeDisplay = document.querySelector('#interframe-size');
+let keyFrameCount = 0;
+let interFrameCount = 0;
+
+function videoAnalyzer(chunk, controller) {
+  const view = new DataView(chunk.data);
+  // The lowest value bit in the first byte is the keyframe indicator.
+  const keyframeBit = view.getUint8(0) & 0x01;
```
fippo, line 247: add a reference to https://tools.ietf.org/html/rfc6386#section-9.1 please or we forget where this comes from :-)
API Q: so this only works for VP8... how do we know what the codec is here?

src/content/peerconnection/video-analyzer/js/main.js
```diff
@@ -0,0 +1,232 @@
...
+  // console.log(view.getUint8(0).toString(16));
+  if (keyframeBit === 0) {
+    keyFrameCount++;
+    keyFrameCountDisplay.innerText = keyFrameCount;
+    keyFrameSizeDisplay.innerText = chunk.data.byteLength;
+  } else {
+    interFrameCount++;
+    interFrameCountDisplay.innerText = interFrameCount;
```
fippo: this is going to update *very* rapidly. Increment and update the display once a second?

fippo (approved):
LGTM with comments

src/content/peerconnection/video-analyzer/js/main.js
alvestrand: moved stuff around - I want the size for a different reason.

src/content/peerconnection/video-analyzer/js/main.js
alvestrand: good point. TODO added (and reference).

src/content/peerconnection/video-analyzer/js/main.js
alvestrand: setTimeout()? Once a second is a bit too slow, it feels like. Trying every half second.

leaysgur: `requestAnimationFrame` is a better choice?

fippo (approved):
:cake:

src/content/peerconnection/video-analyzer/js/main.js
alvestrand: @leader22 requestAnimationFrame will probably fire faster than once-per-frame, so if the point is to make less flicker, setTimeout() seems good.

fippo:
Note: you can also use chunk.type - works on the sender at least, receiver still has minor issues: https://bugs.chromium.org/p/chromium/issues/detail?id=1068468

PR #1281, opened 2020-04-06
e2ee: audio support

fippo:
still unused until requesting with getUserMedia

@alvestrand @guidou PTAL. This uses the same sendTransform... which makes me wonder if it would make sense to pass the track kind, either from the browser or in JS

### DIFF
--- a/src/content/peerconnection/endtoend-encryption/js/videopipe.js
+++ b/src/content/peerconnection/endtoend-encryption/js/videopipe.js
@@ -23,25 +23,31 @@
 
 function VideoPipe(stream, sendTransform, receiveTransform, handler) {
   this.pc1 = new RTCPeerConnection({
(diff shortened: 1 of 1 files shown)


fippo:
Would probably be wise to have a checkbox for enabling audio output on the elements too. Because having both encrypted and unencrypted output is going to be very noisy.

@alvestrand while I like your approach of having a separate middlebox connection for demo purposes maybe it would be easier to just toggle decryption with a checkbox?

fippo:
Note: currently this has no audio output -- I assume due to https://bugs.chromium.org/p/chromium/issues/detail?id=1068125 so don't merge yet

fippo:
can confirm that this is fixed in ToT so this is safe to merge. Would do the checkbox for toggling audio between unencrypted and middlebox in a different PR. The encrypted audio is *very* noisy

PR #1282, opened 2020-04-08
e2ee: audio support

fippo:
see individual commits, first one adds a checkbox to switch between proper output and middlebox garbage.
The second commit updates the no-encryption scheme to support audio (where chunk.type is undefined; I would prefer to have a chunk.kind but this works) and enables audio in getUserMedia

### DIFF
--- a/src/content/peerconnection/endtoend-encryption/js/main.js
+++ b/src/content/peerconnection/endtoend-encryption/js/main.js
@@ -164,7 +164,7 @@ function decodeFunction(chunk, controller) {
     dump(chunk, 'recv');
   }
   const view = new DataView(chunk.data);
-  const checksum = view.getUint32(chunk.data.byteLength - 4);
+  const checksum = chunk.data.byteLength > 4 ? view.getUint32(chunk.data.byteLength - 4) : false;
   if (currentCryptoKey) {
     if (checksum !== 0xDEADBEEF) {
       console.log('Corrupted frame received, checksum ' +
--- a/src/content/peerconnection/endtoend-encryption/js/main.js
+++ b/src/content/peerconnection/endtoend-encryption/js/main.js
@@ -69,7 +69,7 @@ function gotremoteStream(stream) {
 function start() {
   console.log('Requesting local stream');
   startButton.disabled = true;
-  const options = {audio: false, video: true};
+  const options = {audio: true, video: true};
   navigator.mediaDevices
       .getUserMedia(options)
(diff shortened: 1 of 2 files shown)


fippo:
one of the interesting things here is that this shows the weakness of the current (intentionally stupid) crypto scheme -- you can still kinda hear the bip-bop from the fake device even with encryption for short keys.

src/content/peerconnection/endtoend-encryption/js/main.js
```diff
@@ -110,6 +112,19 @@ function dump(chunk, direction, max = 16) {
   console.log(performance.now().toFixed(2), direction, bytes.trim(), chunk.data.byteLength, (data[0] & 0x1) === 0);
 }
 
+// If using crypto offset:
+// Do not encrypt the first bytes of the payload. For VP8
+// this is the content described in
+//   https://tools.ietf.org/html/rfc6386#section-9.1
+// which is 10 bytes for key frames and 3 bytes for delta frames.
+// For opus (where chunk.type is not set) this is the TOC byte from
+//   https://tools.ietf.org/html/rfc6716#section-3.1
```
alvestrand, line 122: Add a description of the purpose, something like:

This allows the middlebox to decode some packets without crashing, making for more interesting pictures and sounds.

src/content/peerconnection/endtoend-encryption/js/main.js
```diff
@@ -110,6 +112,23 @@ function dump(chunk, direction, max = 16) {
...
+//
+// It makes the (encrypted) video and audio much more fun to watch and listen to
+// as the decoder does not immediately throw a fatal error.
+const frameTypeToCryptoOffset = {
+  key: 10,
+  delta: 3,
+  undefined: 1,
```
fippo, line 129: I would very much prefer being able to determine audio/video in a way that differs from "does the frame have a type"

fippo:
there is a fun caveat here. If you set track.enabled to false and don't have a crypto key, the opus silence frames will be three bytes. Which leads to an out-of-bounds trying to read the checksum. Will fix...

fippo:
Additionally the byteLength can be 0 with DTX when there is silence. Handled correctly :-)

src/content/peerconnection/endtoend-encryption/js/main.js
alvestrand: You can determine it when installing the function - at that time you know the track type.

PR #1283, opened 2020-04-08
e2ee: add worker support

fippo:
and just get rid of the other variant. The beauty of this is that worker.js now contains all our "cryptography" (silly, for demo purposes) bits and is trivial to inspect in devtools

### DIFF
--- a/src/content/peerconnection/endtoend-encryption/js/main.js
+++ b/src/content/peerconnection/endtoend-encryption/js/main.js
@@ -33,9 +33,6 @@ muteMiddleBox.addEventListener('change', toggleMute);
 
 let startToMiddle;
 let startToEnd;
-let currentCryptoKey;
-let useCryptoOffset;
-let currentKeyIdentifier = 0;
 
 let localStream;
 // eslint-disable-next-line no-unused-vars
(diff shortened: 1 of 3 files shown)


fippo:
(will rebase after the other two have landed)

alvestrand:
Hm. I'm not too happy with getting rid of the other variant.
Doing stuff in workers and doing stuff in the main thread is kind of different things - if people don't understand workers, they'll have trouble grokking this exampe, while people looking for how to use workers may need an introduction to inserting streams first.

Would it be better to fork the example?

fippo:
one thing I liked on my playground project was putting all the stuff related to the encode function into a separate file (insertable.js) and then it was just a matter of adding a onmessage.
I'll try that, then we can have both

fippo:
doing both turned out to be less readable. Added docs and said "if you want to do it here that is fine too" with the essential bits of the old code.

alvestrand:
what I meant by "clone" was doing one example that uses a worker and one example that doesn't, not trying to mash both into one example - that's indeed not readable.

I'll review this code .... it may be easier to back-clone the worker-based one than to change this into a clone.

src/content/peerconnection/endtoend-encryption/js/main.js
```diff
@@ -79,6 +75,41 @@ function start() {
...
+// for basic concepts.
+const worker = new Worker('./js/worker.js');
+function setupSenderTransform(sender) {
+  const senderStreams = sender.track.kind === 'video' ? sender.createEncodedVideoStreams() : sender.createEncodedAudioStreams();
+  // Instead of creating the transform stream here, we do a postMessage to the worker. The first
+  // argument is an object defined by us, the sceond a list of variables that will be transferred to
+  // the worker. See
+  //   https://developer.mozilla.org/en-US/docs/Web/API/Worker/postMessage
+  /*
+  const transformStream = new TransformStream({
```
alvestrand, line 92: Missing: If you don't want to use a worker, you can instead do:

As written, it looks like a random block of commented-out code.

src/content/peerconnection/endtoend-encryption/js/main.js
```diff
@@ -79,6 +75,41 @@ function start() {
...
+    readableStream: senderStreams.readableStream,
+    writableStream: senderStreams.writableStream,
+  }, [senderStreams.readableStream, senderStreams.writableStream]);
+}
+
+function setupReceiverTransform(receiver) {
+  const receiverStreams = receiver.track.kind === 'video' ? receiver.createEncodedVideoStreams() : receiver.createEncodedAudioStreams();
+  worker.postMessage({
+    type: 'decode',
+    readableStream: receiverStreams.readableStream,
```
alvestrand, line 110: Suggest having a field called "operation" with the value "decode" and a field "type" with the value receiver.track.kind.

src/content/peerconnection/endtoend-encryption/js/worker.js
```diff
@@ -0,0 +1,119 @@
+let currentCryptoKey;
```
alvestrand, line 15: Copyright header and a few words about what this file is for.

src/content/peerconnection/endtoend-encryption/js/worker.js
```diff
@@ -0,0 +1,119 @@
...
+    return; // encrypted in-flight frame but we already forgot about the key.
+  }
+  controller.enqueue(chunk);
+}
+
+onmessage = async (event) => {
+  const {type} = event.data;
+  if (['encode', 'decode'].includes(type)) {
+    const {readableStream, writableStream} = event.data;
+    const transformStream = new TransformStream({
```
alvestrand, line 128: I think it's more readable if you don't do the double-thing and just do

if (encode) 
  transformStream: encodeFunction
  pipe
} else if (decode)
  transformStream: decodeFunction
  pipe
} else if (setCryptoKey)
  set crypto key
}

alvestrand (commented):
Still think it's better to copy this stuff into endtoend-with-worker instead of putting all the concepts here.
But will accept this eventually if you continue to disagree.

src/content/peerconnection/endtoend-encryption/js/main.js
fippo: why do we need to have the kind? (I think this relates to the other discussion we're having about having the kind on the frame)

alvestrand: you asked for it.... when complaining about having to use "undefined" as a key value when determining the kind of the track.

alvestrand:
OK, let's go with this version.

src/content/peerconnection/endtoend-encryption/js/main.js
fippo: right. but if you look at the current code we're not creating a bound audio transform or video transform. We could but I don't think this will be "normal" way to use it. We need more opinions of people using this :-)

PR #1301, opened 2020-05-12
add codec selection with setCodecPreferences

fippo:
because munging is bad and we didn't have any other example. One might
still have to fall back to munging but these samples are not a guide to
SDP munging

### DIFF
--- a/src/content/peerconnection/change-codecs/index.html
+++ b/src/content/peerconnection/change-codecs/index.html
@@ -1,0 +1,78 @@
+<!DOCTYPE html>
+<!--
+ *  Copyright (c) 2015 The WebRTC project authors. All Rights Reserved.
+ *
+ *  Use of this source code is governed by a BSD-style license
+ *  that can be found in the LICENSE file in the root of the source
+ *  tree.
+-->
+<html>
+<head>
+
+    <meta charset="utf-8">
+    <meta name="description" content="WebRTC code samples">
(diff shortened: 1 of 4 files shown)


alvestrand:
I'd prefer to clone this into a new demo rather than cluttering up pc1.
once we start cluttering, there's no end to the clutter we might end up with. One demo, one feature.

alvestrand:
we've started down the slippery slope with sdpSemantics; if you clone it to pc_bells_and_whistles, you can take that one along :-)

fippo:
now someone has to fix apprtc. I nominate @henbos :-p

src/content/peerconnection/change-codecs/index.html
```diff
@@ -0,0 +1,78 @@
...
+    <p>View the console to see logging. The <code>MediaStream</code> object <code>localStream</code>, and the <code>RTCPeerConnection</code>
+        objects <code>pc1</code> and <code>pc2</code> are in global scope, so you can inspect them in the console as
+        well.</p>
+
+    <p>For more information about RTCPeerConnection, see <a href="http://www.html5rocks.com/en/tutorials/webrtc/basics/"
+                                                            title="HTML5 Rocks article about WebRTC by Sam Dutton">Getting
+        Started With WebRTC</a>.</p>
+
+
+    <a href="https://github.com/webrtc/samples/tree/gh-pages/src/content/peerconnection/pc1"
```
alvestrand: Oops - fix the source link...

PR #1303, opened 2020-05-13
e2ee: update to latest chrome implementation

fippo:
to test
  https://bugs.chromium.org/p/chromium/issues/detail?id=1069295#c4


guidou (approved):
Thanks!

PR #1305, opened 2020-05-14
e2ee: polyfill RTCEncoded(Audio|Video)Frame.getMetadata()

fippo:
For the encoded video frame this is an incomplete polyfill since
the additional properties would have to be parsed from the additionalData
and depends on the codec used.

### DIFF
--- a/src/content/peerconnection/endtoend-encryption/js/worker.js
+++ b/src/content/peerconnection/endtoend-encryption/js/worker.js
@@ -12,6 +12,22 @@
  */
 
 'use strict';
+// Polyfill RTCEncoded(Audio|Video)Frame.getMetadata() (not available in M83, available M84+).
+// The polyfill can not be done on the prototype since its not exposed in workers. Instead,
+// it is done as another transformation to keep it separate.
+function polyFillEncodedFrameMetadata(encodedFrame, controller) {
+  if (!encodedFrame.getMetadata) {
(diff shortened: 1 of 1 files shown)


fippo:
will test in tomorrows canary. I haven't seen much of a need for additonalData yet so I am not particularly keen on polyfilling that one ;-)

fippo:
almost... `window` is not defined in a worker so one has to use `self`

guidou (approved):

fippo:
this doesn't work since RTCEncodedAudioFrame is not defined in the worker.

fippo:
@foolip: we need your IDL wisdom. Should RTCEncodedAudioFrame and RTCEncodedVideoFrame exist in the worker scope? They can clearly be used...

foolip:
https://github.com/alvestrand/webrtc-media-streams/blob/master/explainer.md#api is missing the require `Exposed` extended attribute, and I can't tell from the rest of the IDL what's intended here. @alvestrand should these interfaces be exposed to workers?

fippo:
even if we do that, we can't do the polyfill in M83. Pity but... we found a potential spec bug and doing the polyfill at runtime can be considered another transform.

fippo:
updated the code here and filed a (pre-)spec issue

alvestrand:
Also nicely illustrates that chaining transforms can be useful!

PR #1313, opened 2020-06-04
video-analyzer: remove RED

fippo:
to allow easier analysis in wireshark


fippo:
@alvestrand this should make https://bugs.chromium.org/p/webrtc/issues/detail?id=11647 easier. See https://webrtchacks.com/wireshark-debug-vp8/ for how to decode vp8 in wireshark

PR #1314, opened 2020-06-05
audio: allow enabling opus dtx

fippo:
from the js console. This is an expert option.

@alvestrand PTAL -- cc @minyuel

### DIFF
--- a/src/content/peerconnection/audio/js/main.js
+++ b/src/content/peerconnection/audio/js/main.js
@@ -36,6 +36,25 @@ const offerOptions = {
   voiceActivityDetection: false
 };
 
+// Enabling opus DTX is an expert option without GUI.
+// eslint-disable-next-line prefer-const
+let useDtx = false;
+
+// Change the ptime. For opus supported values are [10, 20, 40, 60].
+// Expert option without GUI.
+// eslint-disable-next-line no-unused-vars
+async function setPtime(ptime) {
+  const offer = await pc1.createOffer();
+  await pc1.setLocalDescription(offer);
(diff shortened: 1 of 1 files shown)


src/content/peerconnection/audio/js/main.js
```diff
@@ -36,6 +36,9 @@ const offerOptions = {
   voiceActivityDetection: false
 };
 
+// Enabling opus DTX is an expert option without GUI.
+let useDtx = false;
```
alvestrand, line 41: The linter is anal.
let useDtx; useDtx = false will presumably pass. Or find the magic "lint, lint go away" cantrip.

fippo: we've run into that before in bandwidth.js ;-)

fippo:
updated to also allow changing ptime. Can separate but...

src/content/peerconnection/audio/js/main.js
```diff
@@ -241,6 +248,20 @@ function setDefaultCodec(mLine, payload) {
   return newLine.join(' ');
 }
 
+// Change the ptime. For opus supported values are [10, 20, 40, 60].
+// eslint-disable-next-line no-unused-vars
```
alvestrand: can you add "For experts at the console only" here too? And add ptime to the description?

alvestrand (approved):

src/content/peerconnection/audio/js/main.js
fippo: done. moved to the top too to increase visibility.

PR #1326, opened 2020-07-03
fix: Don't try to add null ice candidate

abalkin:
**Description**

Add a null check before calling addIceCandidate in datachannel/basic sample.

**Purpose**

Avoid logging errors during normal operation.

### DIFF
--- a/src/content/datachannel/basic/js/main.js
+++ b/src/content/datachannel/basic/js/main.js
@@ -116,14 +116,14 @@ function getName(pc) {
   return (pc === localConnection) ? 'localPeerConnection' : 'remotePeerConnection';
 }
 
-function onIceCandidate(pc, event) {
-  getOtherPc(pc)
-      .addIceCandidate(event.candidate)
+function onIceCandidate(pc, {candidate}) {
+  candidate && getOtherPc(pc)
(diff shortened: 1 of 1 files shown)


abalkin:
@googlebot I signed it!

abalkin:
After reading [addIceCandidate](https://developer.mozilla.org/en-US/docs/Web/API/RTCPeerConnection/addIceCandidate) documentation more carefully, I realize that calling it with null argument is not an error, however when I use `datachannel/basic/js/main.js` in my own html page, I see:

```
localPeerConnection ICE candidate: (null)
Failed to add Ice Candidate: TypeError: Failed to execute 'addIceCandidate' on 'RTCPeerConnection': Candidate missing values for both sdpMid and sdpMLineIndex
remotePeerConnection ICE candidate: (null)
Failed to add Ice Candidate: TypeError: Failed to execute 'addIceCandidate' on 'RTCPeerConnection': Candidate missing values for both sdpMid and sdpMLineIndex
```

while original page does not have this issue.

abalkin:
I found the real cause of my problem: I did not include the following line in my .html file

```html
<script src="https://webrtc.github.io/adapter/adapter-latest.js"></script>
```

PR #1327, opened 2020-07-06
audio: prefer setCodecPreferences if available

fippo:
because sdp munging is not good.

Mostly the same code as for video but had to deal with some audio intricacies such as the clockRate


fippo:
@alvestrand @Orphis can you take a look?

alvestrand (approved):

PR #1329, opened 2020-07-10
getDisplayMedia: revert to old state

### LINKED ISSUE #1148: Don't use unnecessary polymer custom elements in WebRTC samples
https://github.com/webrtc/samples/pull/1144#pullrequestreview-170347632 introduced dependencies on polymer custom element stuff that shouldn't have been introduced in samples IMHO. The samples don't even load on Edge anymore.

Samples should be simple and accessible, without unnecessary dependencies, and these dependencies are clearly unnecessary. I suggest they be reverted.

### LINKED ISSUE #1152: Get non-standard stuff out of the WebRTC Samples
https://github.com/webrtc/samples/pull/1144 booted adapter.js for custom non-standard cruft to get `getDisplayMedia`-equivalent behavior in Firefox. We at Mozilla believe in promoting the spec, and do not want people copying this.

This is antithetical to what the purpose of webrtc samples has been up until now, which is to promote samples that we feel good about people copying.

Keeping non-standard stuff out of the webrtc samples was the original purpose of adapter.js. What's changed?

cc @alvestrand @ErikHellman

fippo:
reverts to the state of 21e9bc3e56bb8ec1e9e77036282a72c3af6cb8e7
and applies some fixes ontop of that such as the move of getDisplayMedia to navigator.mediaDevices
and that it has to be triggered from a gesture these days (in some browsers)

Fixes #1148
Fixes #1152

### DIFF
--- a/src/content/getusermedia/getdisplaymedia/css/styles.css
+++ b/src/content/getusermedia/getdisplaymedia/css/styles.css
@@ -1,4 +1,0 @@
-video#captured-media {
-  width: 100%;
-  height: calc(width * 16 / 9);
-}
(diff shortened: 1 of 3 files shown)


guidou (approved):

KaptenJansson:
Lets get it back to basics. Would have preferred to have this in separate PR's but will do.

PR #1333, opened 2020-07-13
audio: visualize audio level through getSynchronizationSources

fippo:
getSynchronizationSources is the fancy new way to do audio levels. However, we don't have samples of how to do it "properly". 
This sample (which is becoming a bit overloaded) collects the data using requestAnimationFrame and then takes the maximum every 500ms. This may be too slow for some purposes but... it's just a sample.


fippo:
The timestamp argument is unused. Linter does not complain :-1:

KaptenJansson:
Wow really cool. I think even if it's slow it's fine, it's a sample!

KaptenJansson (approved):

fippo:
hold on a bit with merging this...

Issue #1196, opened 2019-06-06
"npm test" is failing when run locally


alvestrand:
The command "npm test", when run in a local checkout, gives the error

hta@hta-bag:~/code/webrtc-samples$ npm test

> webrtc-samples@1.0.0 test /home/hta/code/webrtc-samples
> gulp

[13:32:19] Failed to load external module @babel/register
[13:32:19] Requiring external module babel-register
/home/hta/code/webrtc-samples/node_modules/stylelint/node_modules/micromatch/index.js:44
    let isMatch = picomatch(String(patterns[i]), { ...options, onResult }, true);
                                                   ^^^
SyntaxError: Unexpected token ...
    at createScript (vm.js:56:10)
    at Object.runInThisContext (vm.js:97:10)
    at Module._compile (module.js:542:28)
    ...

Since it runs on the travis CI, I assume that there's something missing in the local setup (I did do npm install).

fippo:
usual thing to try: `rm -rf node_modules; npm i`

but its even worse... nightwatch.json hardcodes selenium-server-standalone-3.14.0.jar but package.json does not pin to that version. And the latest version is 3.14.159 (I wonder if there is a reason for that?!). So gulp nightwatch bails out without doing any test. But as shown in https://travis-ci.org/webrtc/samples/jobs/614621276#L547 the return value is 0 which means the test is considered success.

May I suggest reverting to the old test system?

fippo:
nighwatch is now gone and npm test should work better again

PR #1334, opened 2020-07-13
cause a lint error

fippo:
image
(if it does the situation is even worse than the nightwatch failures in #1196)


fippo:
:disappointed_relieved: so we just have to deal with bad rules.

PR #1345, opened 2020-08-24
audio: visualize audio level through getSynchronizationSources

fippo:
reland of #1333 once https://bugs.chromium.org/p/chromium/issues/detail?id=1112952 is done


fippo:
activity in the bug is low enough that we won't interfere. Will give it a heads-up once merged

@KaptenJansson PTAL (no changes from last time)

src/content/peerconnection/audio/js/main.js
```diff
@@ -359,3 +367,33 @@ window.setInterval(() => {
     lastResult = res;
   });
 }, 1000);
+
+if (window.RTCRtpReceiver && ('getSynchronizationSources' in window.RTCRtpReceiver.prototype)) {
```
henbos, line 371: I believe getSynchronizationSources has been available on all major browsers for several years, do we really need to guard it at this point?

src/content/peerconnection/audio/js/main.js
```diff
@@ -359,3 +367,33 @@ window.setInterval(() => {
...
+      return;
+    }
+    const sources = receiver.getSynchronizationSources();
+    sources.forEach(source => {
+      audioLevels.push(source.audioLevel);
+    });
+    if (!lastTime) {
+      lastTime = timestamp;
+    } else if (timestamp - lastTime > 500 && audioLevels.length > 0) {
+      // Update graph every 500ms.
```
henbos, line 389: We're collecting data points for every animation frame (e.g. 30 fps?) but we're only updating the graph every 500 ms. It seems a bit strange to me to call getSSRC this often if we end up with a very rough 0.5 fps audio meter. Should we update it a bit more often or call getSSRCs less often? E.g. a regular setTimeout. No strong opinion here though

src/content/peerconnection/audio/js/main.js
```diff
@@ -359,3 +367,33 @@ window.setInterval(() => {
...
+      const maxAudioLevel = Math.max.apply(null, audioLevels);
+      audioLevelSeries.addPoint(Date.now(), maxAudioLevel);
+      audioLevelGraph.setDataSeries([audioLevelSeries]);
```
henbos, line 392: If animation frames are triggered at an irregular interval, would the graph still plot them out evenly, skewing the graph? Maybe it doesn't matter

src/content/peerconnection/audio/js/main.js
fippo: you should principally guard in real-world applications.

src/content/peerconnection/audio/js/main.js
fippo: the recommendation I heard was to use RAF together with this API.

src/content/peerconnection/audio/js/main.js
fippo: unless you can tell apart 10ms on the x axis it probably doesn't :-)

henbos (approved):
Ack, merge away

PR #1348, opened 2020-09-01
correct typo "fraemStreams"

luisfonsivevo:
Fixes a console error at the [video analyzer page](https://webrtc.github.io/samples/src/content/peerconnection/video-analyzer/).

PR #1351, opened 2020-09-08
fix lint issues

fippo:
as #1348 showed I slipped in a typo. The linter did not notice since the lint
rules do not include no-undef like they did before #1083

This change adds lint rules + globals to fix the lint process.

### DIFF
--- a/src/content/extensions/multipleroutes/src/options.js
+++ b/src/content/extensions/multipleroutes/src/options.js
@@ -34,7 +34,8 @@ if (!browserSupportsIPHandlingPolicy()) {
 // Saves options.
 function saveOptions() {
   const radios = document.getElementsByName('ip_policy_selection');
-  for (let i = 0; i < radios.length; i++) {
+  let i;
+  for (i = 0; i < radios.length; i++) {
     if (radios[i].checked) {
       break;
     }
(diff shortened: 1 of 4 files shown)


src/content/extensions/multipleroutes/src/options.js
```diff
@@ -34,7 +34,8 @@ if (!browserSupportsIPHandlingPolicy()) {
 // Saves options.
 function saveOptions() {
   const radios = document.getElementsByName('ip_policy_selection');
-  for (let i = 0; i < radios.length; i++) {
+  let i;
+  for (i = 0; i < radios.length; i++) {
```
fippo, line 38: `i` was used outside the loop here.

henbos: i is used below (after breaking out of the loop), line 46

henbos (approved):

src/content/extensions/multipleroutes/src/options.js
fippo: yeah. This was ok in the es5 var syntax (which essentially was function scope) but not with the es6 syntax (which does block scope). Now I doubt the extension was ever updated with the code (which tbh doesn't even belong into this repo but... solved the "how can we not have a new repo" process)

guidou (approved):

PR #1362, opened 2020-10-01
Fix: Make the media element lookup dynamic

NaveenPantra:
**Description**
In line `74` of `main.js` there is fix line `FIXME: Make the media element lookup dynamic.`

Fixing this by having `data` attribute on the `media` and `select` elements, when user changes the option we will fetch the `type` from the `dataset` of the `select` element and `querySelect` the `[data-type="<type>"]` to get the corresponding media element (i.e both `select` and `media` element have same `data-type` attribute but when we do `querySelector` the first matching element will be returned which is `media` element).

**Purpose**
`FIXME: Make the media element lookup dynamic.`


NaveenPantra:
@googlebot I signed it!

NaveenPantra:
@fippo can you review this? I'm not able to add reviewers.

PR #1363, opened 2020-10-02
Cleanup datachannel senders and make errors recieved upon or after datachannel logged with console.log rather than error

### LINKED ISSUE #1261: Error in sendChannel
- [ ] I have provided steps to reproduce
- [ ] I have provided browser name and version
- [ ] I have provided a link to the sample here or a modified version thereof

**Note: If the checkboxes above are not checked (which you do after the issue is posted), the issue will be closed.**

## Browser affected
 Chrome
##Tested on 82.0.4056.0(64 bit).
## Description
While sending the file,in java console getting error like 'Error in sendChannel'.Its coming for any file size as well as 0KB file.
Error coming for MAC,Windows,Linux

## Steps to reproduce
0. Browse to https://webrtc.github.io/samples/src/content/datachannel/filetransfer/
1. Open the javascript console (alt+cmd+j or ctr+shift+j) and watch for any errors
2. Click the "Choose File" button
3. Select a file of any size
4. Verify that the Average Bitrate is > 0
5. Click the "Click to download $filename" link
6. Download the file, verify that it has the same size as the originally uploaded file.
7. Click the "Choose File" button
8. Select a file of size 0 bytes
9. Verify that the message "File is empty, please select a non-empty file" appears

## Expected results
File should sent successfully without any javascript  console error.

## Actual results
Error showing in javascript console
test case 9 error
test case 9

KaptenJansson:
**Description**
Fix error sent after data channel closure

**Purpose**
Fix #1261


src/content/datachannel/filetransfer/js/main.js
```diff
@@ -180,56 +182,69 @@ function receiveChannelCallback(event) {
   }
 }
 
-function onReceiveMessageCallback(event) {
-  console.log(`Received Message ${event.data.byteLength}`);
-  receiveBuffer.push(event.data);
-  receivedSize += event.data.byteLength;
+async function onReceiveMessageCallback(event) {
+  if (sendChannel && receiveChannel) {
```
fippo: this is odd since you're using state only known at the sender on the receiving end?

KaptenJansson: Ah, good spot, I left this when trying to figure why onerror is fired after the datachannel is closed, the only error type that should be thrown at close time is network error which is not the case here (even though close transport is network related, it's not an error per see and also of the wrong type)

src/content/datachannel/filetransfer/js/main.js
```diff
@@ -215,21 +216,33 @@ function onReceiveMessageCallback(event) {
 }
 
 function onSendChannelStateChange() {
-  const readyState = sendChannel.readyState;
-  console.log(`Send channel state is: ${readyState}`);
-  if (readyState === 'open') {
-    sendData();
+  if (sendChannel) {
+    const readyState = sendChannel.readyState;
```
fippo, line 220: be fancy and do `const {readyState} = sendChannel;`

src/content/datachannel/filetransfer/js/main.js
```diff
@@ -180,11 +182,10 @@ function receiveChannelCallback(event) {
   }
 }
 
-function onReceiveMessageCallback(event) {
+async function onReceiveMessageCallback(event) {
```
fippo, line 185: not required?

fippo (approved):
LGTM with nits

fippo (approved):
still approving, can fix the nits later ;-)

alvestrand (approved):
Seconding fippo.

PR #1374, opened 2020-11-10
Fix duplicated codec infomation

dong-heun:
**Description**
When changing the priority of the codec, codec information is duplicated

**Purpose**
remove duplication of codec list.

### DIFF
--- a/src/content/insertable-streams/endtoend-encryption/js/videopipe.js
+++ b/src/content/insertable-streams/endtoend-encryption/js/videopipe.js
@@ -43,7 +43,7 @@ function VideoPipe(stream, forceSend, forceReceive, handler) {
     const {codecs} = RTCRtpSender.getCapabilities('video');
     const selectedCodecIndex = codecs.findIndex(c => c.mimeType === preferredVideoCodecMimeType);
     const selectedCodec = codecs[selectedCodecIndex];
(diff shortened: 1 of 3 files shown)


fippo (approved):
Not a typo, I mixed up slice (returns element, does not change array) and splice (removes element)

Thanks for catching this!

(@alvestrand should setParameters validate and throw? :-))

fippo:
"4. Remove any duplicate values in codecs." -- https://w3c.github.io/webrtc-pc/#dom-rtcrtptransceiver-setcodecpreferences
-- so not throwing is ok.

dong-heun:
> "4. Remove any duplicate values in codecs." -- https://w3c.github.io/webrtc-pc/#dom-rtcrtptransceiver-setcodecpreferences
> -- so not throwing is ok.

Thanks for your comments.

So I updated my patch and comments. Please check.

fippo:
ah... I still think its a good idea to actually splice and remove the element even if duplicates are handled. Can you revert that, the first version was fine.

dong-heun:
> ah... I still think its a good idea to actually splice and remove the element even if duplicates are handled. Can you revert that, the first version was fine.

Dear @fippo 
I revert the first version. Please check again.
thanks.

fippo (approved):

henbos (approved):

PR #1375, opened 2020-11-12
Fix event handler "onnegotiationeeded" has two Ns

yousefamar:
https://developer.mozilla.org/en-US/docs/Web/API/RTCPeerConnection/onnegotiationneeded


henbos (approved):

henbos:
Thanks!

PR #1379, opened 2020-11-18
Await the async transform function.

dogben:
I noticed a bug in https://github.com/webrtc/samples/pull/1378; the transform function is async, so it must be awaited.


dogben:
@KaptenJansson

PR #1381, opened 2020-11-20
Reorganize insertable stream samples

dogben:
Create a new heading in index.html for insertable streams. Move existing samples to the new group (and also move the files to the insertable-stream subdir). Add the video processing sample.

After discussing with @guidou and @alvestrand, it seems best to group these samples together. They use slightly different APIs, but the overall concept is similar. Indeed, the explainers for [WebRTC Insertable Streams](https://github.com/w3c/webrtc-insertable-streams/blob/master/explainer.md) and [MediaStream Insertable Streams](https://github.com/w3c/mediacapture-insertable-streams/blob/main/explainer.md) share a lot of verbiage.


KaptenJansson (approved):
LGTM

Yeah this makes sense.

PR #1386, opened 2020-11-27
In video processing sample, indicate when page is ready.

dogben:
For slow connections, there can be a long delay between when the page is visible (and the user can interact with the UI elements) and when all the scripts are loaded and the event handlers have been installed. This PR changes the controls to be disabled on page load and enabled when the handlers have been added to indicate when the page is ready.

### DIFF
--- a/src/content/insertable-streams/video-processing/index.html
+++ b/src/content/insertable-streams/video-processing/index.html
@@ -46,17 +46,17 @@
 
     <div class="box">
         <span>Source:</span>
-        <select id="sourceSelector">
+        <select id="sourceSelector" disabled>
             <option selected value="">(stopped)</option>
             <option value="camera">Camera</option>
             <option value="video">Video</option>
             <option value="pc">Peer connection (from camera)</option>
         </select>
-        <span>Add to page: <input type="checkbox" id="sourceVisible"></span>
+        <span>Add to page: <input type="checkbox" id="sourceVisible" disabled></span>
     </div>
     <div class="box">
         <span>Transform:</span>
-        <select id="transformSelector">
+        <select id="transformSelector" disabled>
             <option selected value="webgl">WebGL</option>
             <option value="canvas2d">Canvas2D</option>
             <option value="drop">Drop frames at random</option>
@@ -65,7 +65,7 @@
     </div>
     <div class="box">
         <span>Destination:</span>
-        <select id="sinkSelector">
+        <select id="sinkSelector" disabled>
             <option selected value="video">Video</option>
             <option value="pc">Peer connection</option>
         </select>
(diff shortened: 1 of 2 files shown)


src/content/insertable-streams/video-processing/js/main.js
```diff
@@ -165,14 +165,19 @@ function updatePipelineSourceIfSet() {
...
-  if (!sourceType) {
+  if (!sourceType || !pipeline) {
     initPipeline();
   } else {
     updatePipelineSourceIfSet();
   }
 }
 sourceSelector.oninput = updatePipelineSource;
-sourceVisibleCheckbox.oninput = () => {
+sourceSelector.disabled = false;
```
KaptenJansson: Can you use https://javascript.info/onload-ondomcontentloaded#window-onload instead or are do you also need to wait for devices to show up (I assume onload only waits for all resources to be (down)loaded and not that all requested hardware is loaded/available)?

KaptenJansson (changes requested):

src/content/insertable-streams/video-processing/js/main.js
dogben: Is https://github.com/webrtc/samples/pull/1386/commits/e8bb1b2fb73e3b63a3fe2805addab04a05f20bb2 along the lines of what you were thinking? It allows changing the scripts to async, which may help page load delay in some cases.

I also made https://github.com/webrtc/samples/pull/1386/commits/c0dcfc989accccac1642cbce58eeaece48a66800 to put the UI code in a class, but it makes the change harder to review and probably doesn't make things simpler.

src/content/insertable-streams/video-processing/index.html
```diff
@@ -85,18 +85,18 @@ <h1><a href="//webrtc.github.io/samples/" title="WebRTC samples homepage">WebRTC
...
-<script src="js/peer-connection-sink.js" defer></script>
-<script src="js/peer-connection-source.js" defer></script>
-<script src="js/pipeline.js" defer></script>
-<script src="js/simple-transforms.js" defer></script>
-<script src="js/video-mirror-helper.js" defer></script>
-<script src="js/video-sink.js" defer></script>
-<script src="js/video-source.js" defer></script>
-<script src="js/webgl-transform.js" defer></script>
-<script src="js/main.js" defer></script>
+<script src="js/camera-source.js" async></script>
```
KaptenJansson, line 88: Are any of these dependent on order and that the page has completely loaded(looks like everything is now triggered by the onload event thus I assume no but I just want to make sure)?

KaptenJansson (approved):
LGTM with a comment and readability is fine.

src/content/insertable-streams/video-processing/index.html
dogben: That's right. Now, the only non-declaration in any of the files is:
window.onload = initUI;

PR #1404, opened 2021-01-08
Add a Breakout Box transform using WebCodec

alvestrand:
This demonstrates that the Breakout Box API and WebCodec API are
compatible.
Also adds a "null" transform, which can be useful for performance
measurements.

### DIFF
--- a/src/content/insertable-streams/video-processing/index.html
+++ b/src/content/insertable-streams/video-processing/index.html
@@ -59,8 +59,10 @@
         <select id="transformSelector" disabled>
             <option selected value="webgl">WebGL</option>
             <option value="canvas2d">Canvas2D</option>
+            <option value="noop">Do nothing</option>
             <option value="drop">Drop frames at random</option>
             <option value="delay">Delay all frames by 100ms</option>
+            <option value="webcodec">Run frames through WebCodec</option>
         </select>
     </div>
     <div class="box">
@@ -95,6 +97,7 @@
 <script src="js/video-mirror-helper.js" async></script>
 <script src="js/video-sink.js" async></script>
 <script src="js/video-source.js" async></script>
+<script src="js/webcodec-transform.js" async></script>
 <script src="js/webgl-transform.js" async></script>
 <script src="js/main.js" async></script>
 
--- a/src/content/insertable-streams/video-processing/js/simple-transforms.js
+++ b/src/content/insertable-streams/video-processing/js/simple-transforms.js
@@ -9,6 +9,21 @@
 'use strict';
 
 /**
+ * Does nothing.
+ * @implements {FrameTransform} in pipeline.js
+ */
+class NullTransform { // eslint-disable-line no-unused-vars
+  /** @override */
+  async init() {}
+  /** @override */
+  async transform(frame, controller) {
+    controller.enqueue(frame);
+  }
+  /** @override */
+  destroy() {}
+}
+
+/**
  * Drops frames at random.
  * @implements {FrameTransform} in pipeline.js
  */
--- a/src/content/insertable-streams/video-processing/js/webcodec-transform.js
+++ b/src/content/insertable-streams/video-processing/js/webcodec-transform.js
@@ -17,17 +17,24 @@
 
 class WebCodecTransform { // eslint-disable-line no-unused-vars
   constructor() {
-    // All fields are initialized in init()
-    this.debugPath_ = 'debug.pipeline.frameTransform_';
+    // Encoder and decoder are initialized in init()
+    this.decoder_ = null;
+    this.encoder_ = null;
+    this.controller_ = null;
   }
   /** @override */
   async init() {
     console.log('[WebCodecTransform] Initializing encoder and decoder');
-    this.decoder_ = new VideoDecoder({output: this.enqueueDecoded.bind(this), error: this.error});
-    this.encoder_ = new VideoEncoder({output: this.decodeEncoded.bind(this), error: this.error});
+    this.decoder_ = new VideoDecoder({
+      output: frame => this.handleDecodedFrame(frame),
+      error: this.error
+    });
+    this.encoder_ = new VideoEncoder({
+      output: frame => this.handleEncodedFrame(frame),
+      error: this.error
+    });
     this.encoder_.configure({codec: 'vp8', width: 640, height: 480});
     this.decoder_.configure({codec: 'vp8', width: 640, height: 480});
-    this.controller_ = null;
   }
 
   /** @override */
@@ -44,16 +51,16 @@ class WebCodecTransform { // eslint-disable-line no-unused-vars
   destroy() {}
 
   /* Helper functions */
-  enqueueDecoded(videoFrame) {
+  handleEncodedFrame(encodedFrame) {
+    this.decoder_.decode(encodedFrame);
+  }
+
+  handleDecodedFrame(videoFrame) {
     if (!this.controller_) {
       videoFrame.destroy();
       return;
     }
     this.controller_.enqueue(videoFrame);
-  }
-
-  decodeEncoded(encodedFrame) {
-    this.decoder_.decode(encodedFrame);
   }
 
   error(e) {
--- a/src/content/insertable-streams/video-processing/js/webcodec-transform.js
+++ b/src/content/insertable-streams/video-processing/js/webcodec-transform.js
@@ -12,6 +12,9 @@
  * Encodes and decodes frames using the WebCodec API.
  * @implements {FrameTransform} in pipeline.js
  */
+
+/* global VideoEncoder, VideoDecoder */ // part of WebCodec specification
+
 class WebCodecTransform { // eslint-disable-line no-unused-vars
   constructor() {
     // All fields are initialized in init()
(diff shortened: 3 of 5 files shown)


alvestrand:
Calling on @dogben to review

src/content/insertable-streams/video-processing/js/webcodec-transform.js
```diff
@@ -0,0 +1,60 @@
...
+'use strict';
+
+/**
+ * Encodes and decodes frames using the WebCodec API.
+ * @implements {FrameTransform} in pipeline.js
+ */
+class WebCodecTransform { // eslint-disable-line no-unused-vars
+  constructor() {
+    // All fields are initialized in init()
+    this.debugPath_ = 'debug.pipeline.frameTransform_';
```
dogben: This field is unused.

The purpose of debugPath_ is to indicate how to inspect objects when someone opens the developer console. Take a look at some of the other files.

If you don't want to do that here, the field should be removed.

src/content/insertable-streams/video-processing/js/webcodec-transform.js
```diff
@@ -0,0 +1,60 @@
...
+  }
+  /** @override */
+  async init() {
+    console.log('[WebCodecTransform] Initializing encoder and decoder');
+    this.decoder_ = new VideoDecoder({ output: this.enqueueDecoded.bind(this), error: this.error });
```
dogben: The browser APIs that eslint doesn't recognize can be added to https://github.com/webrtc/samples/blob/gh-pages/.eslintrc.js

src/content/insertable-streams/video-processing/js/webcodec-transform.js
```diff
@@ -0,0 +1,60 @@
...
+    }
+    this.controller_ = controller;
+    this.encoder_.encode(frame);
+  }
+
+  /** @override */
+  destroy() {}
+
+  /* Helper functions */
+  enqueueDecoded(videoFrame) {
```
dogben: Maybe consider renaming to handleDecodedFrame

src/content/insertable-streams/video-processing/js/webcodec-transform.js
```diff
@@ -0,0 +1,60 @@
...
+    if (!this.controller_) {
+      videoFrame.destroy();
+      return;
+    }
+    this.controller_.enqueue(videoFrame);
+  }
+
+  decodeEncoded(encodedFrame) {
```
dogben: Maybe consider renaming to handleEncodedFrame

src/content/insertable-streams/video-processing/js/webcodec-transform.js
dogben: Maybe consider ordering the encoder-related code before the decoder-related code to follow the chronological order of events.

dogben: The Google style guide recommends using lambas instead of bind. I.e. change `this.enqueueDecoded.bind(this)` to `frame => this.enqueueDecoded(frame)`.

https://google.github.io/styleguide/jsguide.html#features-functions-arrow-functions

src/content/insertable-streams/video-processing/js/webcodec-transform.js
```diff
@@ -0,0 +1,60 @@
...
+class WebCodecTransform { // eslint-disable-line no-unused-vars
+  constructor() {
+    // All fields are initialized in init()
+    this.debugPath_ = 'debug.pipeline.frameTransform_';
+  }
+  /** @override */
+  async init() {
+    console.log('[WebCodecTransform] Initializing encoder and decoder');
+    this.decoder_ = new VideoDecoder({ output: this.enqueueDecoded.bind(this), error: this.error });
+    this.encoder_ = new VideoEncoder({ output: this.decodeEncoded.bind(this), error: this.error });
```
dogben: The Google style guide recommends using lambas instead of bind. I.e. change `this.decodeEncoded.bind(this)` to `frame => this.decodeEncoded(frame)`. Or perhaps even just inline it: `frame => this.decoder_.decode(frame)`.

https://google.github.io/styleguide/jsguide.html#features-functions-arrow-functions

src/content/insertable-streams/video-processing/js/webcodec-transform.js
dogben: The Google style guide recommends all fields are declared in the constructor.

https://google.github.io/styleguide/jsguide.html#features-classes-fields

src/content/insertable-streams/video-processing/js/webcodec-transform.js
dogben: The Google style guide recommends all fields are declared in the constructor.

https://google.github.io/styleguide/jsguide.html#features-classes-fields

src/content/insertable-streams/video-processing/js/webcodec-transform.js
```diff
@@ -0,0 +1,60 @@
...
+    this.encoder_.configure({codec: 'vp8', width: 640, height:480});
+    this.decoder_.configure({codec: 'vp8', width: 640, height:480});
+    this.controller_ = null;
```
dogben: The Google style guide recommends all fields are declared in the constructor.

https://google.github.io/styleguide/jsguide.html#features-classes-fields

dogben (commented):
You should resolve all the eslint errors. If you would like to run eslint locally to save time, I can give some help.

I'm not sure if this repo is supposed to use the Google style guide or not. However, since it's imported into the Chromium repo, I figured I should stick with Chromium standards. Probably not a big deal either way, but I mentioned a few things that don't match the style guide just in case.

src/content/insertable-streams/video-processing/js/webcodec-transform.js
dogben: (See "globals" section of that file.)

src/content/insertable-streams/video-processing/js/webcodec-transform.js
alvestrand: Any particular preference for declaration ( = null) vs actual initialization in the constructor?
The other examples seemed to prefer to do init work that could fail in init().

src/content/insertable-streams/video-processing/js/webcodec-transform.js
alvestrand: Removed.

src/content/insertable-streams/video-processing/js/webcodec-transform.js
alvestrand: Done.

src/content/insertable-streams/video-processing/js/webcodec-transform.js
alvestrand: Done

src/content/insertable-streams/video-processing/js/webcodec-transform.js
alvestrand: Done.

src/content/insertable-streams/video-processing/js/webcodec-transform.js
alvestrand: Done.

src/content/insertable-streams/video-processing/js/webcodec-transform.js
dogben: I'm not sure. Coming from C++ (minus exceptions), I have a habit of not doing much in the constructor.

The reason for having the init method was mainly to avoid instantiating WebGL on page load, since the WebGL transform is selected by default. In retrospect, I probably should have simplified things by avoiding an init method and using a callback/factory to instantiate things instead.

alvestrand (commented):
dogben PTAL.

src/content/insertable-streams/video-processing/js/simple-transforms.js
```diff
@@ -8,6 +8,21 @@
...
+  /** @override */
+  async transform(frame, controller) {
+    controller.enqueue(frame);
+  }
+  /** @override */
+  destroy() {}
+}
+
 /**
  * Drops frames at random.
```
dogben, line 27: We might consider removing DropTransform in favor of NullTransform, since they don't really demonstrate unique uses of the API. (I just thought NullTransform would be too dull, so I tried to make it more interesting; however, NullTransform is more interesting for testing and experimentation.)

src/content/insertable-streams/video-processing/js/webcodec-transform.js
```diff
@@ -0,0 +1,62 @@
...
+ *  that can be found in the LICENSE file in the root of the source
+ *  tree.
+ */
+
+'use strict';
+
+/**
+ * Encodes and decodes frames using the WebCodec API.
+ * @implements {FrameTransform} in pipeline.js
+ */
```
dogben, line 14: Doc comment should be immediately before the class.

src/content/insertable-streams/video-processing/index.html
```diff
@@ -59,8 +59,10 @@ <h1><a href="//webrtc.github.io/samples/" title="WebRTC samples homepage">WebRTC
         <select id="transformSelector" disabled>
             <option selected value="webgl">WebGL</option>
             <option value="canvas2d">Canvas2D</option>
+            <option value="noop">Do nothing</option>
             <option value="drop">Drop frames at random</option>
```
dogben, line 63: We might consider removing DropTransform in favor of NullTransform, since they don't really demonstrate unique uses of the API. (I just thought NullTransform would be too dull, so I tried to make it more interesting; however, NullTransform is more interesting for testing and experimentation.)

dogben (approved):

src/content/insertable-streams/video-processing/index.html
alvestrand: When we get some kind of performance monitoring in there, NullTransform is interesting because it sets the lowest possible benchmark. DropTransform is kind of interesting in that it gives a jagged movement look, and it's one example of user-provided adaptation (lower bitrate by frame decimation). So I'd keep both.

(One of the things we can extend with when we pass through the PeerConnection is to measure the resulting bandwidth - DropTransform should show less bandwidth; the same should happen if we applied a WebGL "smoothing" filter).

src/content/insertable-streams/video-processing/js/webcodec-transform.js
alvestrand: And the intervening lines are no longer needed...

PR #1408, opened 2021-01-14
add requestVideoFrameCallback

fippo:
which is very nice and even has WebRTC extensions:
  https://web.dev/requestvideoframecallback-rvfc/

### DIFF
--- a/index.html
+++ b/index.html
@@ -154,6 +154,8 @@
 
             <li><a href="src/content/peerconnection/old-new-stats/">More constraints and stats</a></li>
 
+            <li><a href="src/content/peerconnection/per-frame-callback/">RTCPeerConnection and requestVideoFrameCallback()</a></li>
+
             <li><a href="src/content/peerconnection/create-offer/">Display createOffer output for various scenarios</a>
             </li>
 
--- a/src/content/peerconnection/per-frame-callback/css/main.css
+++ b/src/content/peerconnection/per-frame-callback/css/main.css
@@ -1,0 +1,37 @@
+/*
+ *  Copyright (c) 2021 The WebRTC project authors. All Rights Reserved.
+ *
+ *  Use of this source code is governed by a BSD-style license
+ *  that can be found in the LICENSE file in the root of the source
+ *  tree.
+ */
+video {
+  --width: 45%;
+  width: var(--width);
+  height: calc(var(--width) * 0.75);
+}
+
+button {
+  margin: 0 20px 0 0;
+  width: 96px;
+}
+
+video#localVideo {
+  margin: 0 20px 20px 0;
+}
+
+div.label {
+  display: inline-block;
+  font-weight: 400;
+  width: 120px;
+}
+
+div.graph-container {
+  float: left;
+  margin: 0.5em;
+  width: calc(50% - 1em);
+}
+
+a#viewSource {
+  clear: both;
+}
--- a/src/content/peerconnection/per-frame-callback/index.html
+++ b/src/content/peerconnection/per-frame-callback/index.html
@@ -48,15 +48,19 @@
         <button id="callButton">Call</button>
         <button id="hangupButton">Hang Up</button>
     </div>
+    <div class="graph-container" id="localDelayGraph">
+        <div>Local capture delay (converted to milliseconds)</div>
+        <canvas id="localDelayCanvas"></canvas>
+    </div>
     <div class="graph-container" id="networkDelayGraph">
         <div>Network delay (milliseconds)</div>
         <canvas id="networkDelayCanvas"></canvas>
     </div>
+    <br/>
     <div class="graph-container" id="timeGraph">
-        <div>Render delay (milliseconds): presentation time minus receive time</div>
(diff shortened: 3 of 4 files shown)


fippo:
deployed version here: https://fippo.github.io/webrtc/src/content/peerconnection/per-frame-callback/
:wave: @juberti 
I found this useful to look at the timing for insertable streams so this might be interesting for @alvestrand and @guidou 

@tguilbert-google I don't see `captureTime`. Also there is a (experimental) header extension for the [absolute capture time ](https://chromium.googlesource.com/external/webrtc/+/refs/heads/master/docs/native-code/rtp-hdrext/abs-capture-time/) which might actually be more interesting than the conversion of the RTP timestamp back to an absolute since iiuc this would include the capture delay so one could measure glass-to-glass delay

fippo:
tagging @drkron wrt lack of captureTime

drkron:
First of all, nice demo!


> tagging @drkron wrt lack of captureTime

I had to do some digging. At the moment captureTime is dependent on SenderReports in both directions to correctly do the mapping from RTP timestamp to time and to estimate the round trip time. Since the demo is one-directional you only have SenderReports in one direction. The missing part is the round trip estimation.

Two options for the demo:

- Either add an empty send track in the opposite direction.

- There's also a way to get round trip estimation to work in this case by enabling a special RTCP feedback message through SDP munging. What you do then is to add an extra RTCP feedback parameter line for each payload. The parameter to add is [rrtr](https://source.chromium.org/chromium/chromium/src/+/master:third_party/webrtc/media/base/media_constants.cc;drc=45f7206093c316686634b417e99d0f8ecdda317d;l=87). So something like: "a=rtcp-fb:[payload] rrtr

I haven't tried any of this myself :)

nit. Processing time is in seconds so you need to scale it to get ms.

fippo:
Nice functionality needs a nice demo :-)

Updated, good catch with the conversion! I added the track to pc2 too with a note which makes captureTime work indeed.
`rrtr` works when munged before pc2's setLocalDescription call but we want to avoid that in the samples.

drkron:
Thanks! Good to know that both options worked.

fippo:
What does not work is the local time. I'd expect the following to have `captureTime` in the [GUM sample](https://webrtc.github.io/samples/src/content/getusermedia/gum/)
```
document.querySelector("video").requestVideoFrameCallback((_, metaData) => console.log(metaData))
```

drkron:
> What does not work is the local time. I'd expect the following to have `captureTime` in the [GUM sample](https://webrtc.github.io/samples/src/content/getusermedia/gum/)

This is being worked on and should soon be resolved, see https://crbug.com/1169586 .

fippo:
Added local capture delay (i.e. presentationTime - captureTime) now that it is working.
Would it also make sense to graph `now - captureTime` in the same graph?
I am surprised that the local delay is 500ms+ (and I am seeing spikes of up to 6 seconds?!), OTOH this minimizes the delay between local and remote and makes things look natural.

Sadly this doesn't allow measuring a glass-to-glass delay yet since the local rVFC has no rtp timestamp for correlation.

fippo:
ready to merge?

drkron (approved):
LGTM!

KaptenJansson (approved):

chcunningham:
Sorry to jump on an old PR. Quick question about @fippo's comment

> Sadly this doesn't allow measuring a glass-to-glass delay yet since the local rVFC has no rtp timestamp for correlation.

I'm tempted to compute glass-to-glass delay = [expectedDisplayTime](https://wicg.github.io/video-rvfc/#dom-videoframecallbackmetadata-expecteddisplaytime) - [captureTime](https://wicg.github.io/video-rvfc/#dom-videoframecallbackmetadata-capturetime). What am I missing?

fippo:
@chcunningham that could work. There is some magic going on because you can't assume the remote clock to be the same as the local one but that is taken care of :crossed_fingers: 

I'd recommend comparing the result of that to "physical" ones, e.g. taking a photo of the result of filming a high resolution clock and displaying it.

josecpujol:
Question somehow related to this: 

I tried adding the rrtr support in the SDP, and everything works fine: I am getting `captureTime` values after a couple of seconds in the `requestVideoFrameCallback` function, which is good.

However, I am experimenting with the `http://www.webrtc.org/experiments/rtp-hdrext/abs-capture-time` extension, that includes both the NTP time of capture and the estimated capture clock offset. With this, I was expecting to get the same data in the callback, but `captureTime` is not populated. Instead, I have to inspect the `RTCRtpContributingSource` object, and then I get 

```
{"captureTimestamp":3884112907193,"rtpTimestamp":1552810795,"senderCaptureTimeOffset":0,"source":1823295184,"timestamp":1675124107212}
```
which does not give the data in receiver clock reference.

AFAIU, having the capture clock offset estimation allows the receiver to get the same data as in the rrtr case. Is it expected to have `requestVideoFrameCallback` function populate `captureTime`, in the same way that the rrtr (and having a reversed pc) do?

PR #1414, opened 2021-01-27
Measure negotiation with added/removed transceivers

alvestrand:
**Description**
Adds a demo page that adds and removes transceivers, and measures impact on the audio channel.

**Purpose**
Track down exactly what negotiations impact the audio quality.
This should also serve as a basis for a performance metric that can be tracked by bots, so that we do not regress.


src/content/peerconnection/negotiate-timing/js/main.js
```diff
@@ -0,0 +1,228 @@
...
+    if (stat.type == 'track') {
+      currentImpairment = stat.concealedSamples;
+    }
+  });
+  console.log('Found impairment value ', currentImpairment);
+  return currentImpairment;
+}
+
+async function baselineAudioImpairment(pc) {
+  audioTransceiver = pc.getTransceivers().filter(tr => tr.receiver.track.kind == 'audio')[0];
```
fippo: .find and remove the [0]. Should aduioTransceiver be set in ontrack?

src/content/peerconnection/negotiate-timing/css/main.css
```diff
@@ -0,0 +1,44 @@
+/*
+ *  Copyright (c) 2017 The WebRTC project authors. All Rights Reserved.
```
fippo, line 2: nit: 2021

src/content/peerconnection/negotiate-timing/test/test.js
```diff
@@ -0,0 +1,45 @@
+/*
```
fippo: is this test needed? They're in bad shape anyway

fippo (approved):
some nits, LGTM otherwise

src/content/peerconnection/negotiate-timing/js/main.js
```diff
@@ -0,0 +1,232 @@
...
+  console.log(`Remote video videoWidth: ${this.videoWidth}px,  videoHeight: ${this.videoHeight}px`);
+});
+
+remoteVideo.onresize = () => {
+  console.log(`Remote video size changed to ${remoteVideo.videoWidth}x${remoteVideo.videoHeight}`);
+  console.warn('RESIZE', remoteVideo.videoWidth, remoteVideo.videoHeight);
+  // We'll use the first onsize callback as an indication that video has started
+  // playing out.
+  if (startTime) {
+    const elapsedTime = window.performance.now() - startTime;
```
dogben, line 47: If elapsedTime is what you plan to measure, it will need to be at global scope.

src/content/peerconnection/negotiate-timing/js/main.js
```diff
@@ -0,0 +1,232 @@
...
+  await baselineAudioImpairment(pc2);
+  const previousVideoTransceiverCount = pc2.getTransceivers().filter(tr => tr.receiver.track.kind == 'video').length;
+  const startTime = performance.now();
+  const offer = await pc1.createOffer();
+  await pc1.setLocalDescription(offer);
+  await pc2.setRemoteDescription(offer);
+  const answer = await pc2.createAnswer();
+  await pc1.setRemoteDescription(answer);
+  await pc2.setLocalDescription(answer);
+  const elapsedTime = performance.now() - startTime;
```
dogben: Maybe de-duplicate with the code in call()?

src/content/peerconnection/negotiate-timing/js/main.js
```diff
@@ -0,0 +1,232 @@
...
+  pc1.close();
+  pc2.close();
+  pc1 = null;
+  pc2 = null;
+
+  const videoTracks = localStream.getVideoTracks();
+  videoTracks.forEach(videoTrack => {
+    videoTrack.stop();
+    localStream.removeTrack(videoTrack);
+  });
```
dogben, line 219: I don't understand the reason for this. It seems like if you stop all the local video tracks, I need to reload the page before I can click call again.

dogben (commented):
LGTM aside from fippo's comments!

src/content/peerconnection/negotiate-timing/test/test.js
alvestrand: Nuking the test directory. I copied it without thinking.

src/content/peerconnection/negotiate-timing/css/main.css
alvestrand: I cloned a 2017 sample, and then changed 90% of the lines, so I suppose 2021 is appropriate...

src/content/peerconnection/negotiate-timing/js/main.js
alvestrand: Could do that, but then ontrack has to distinguish between 3 cases (audio, video and "don't care").

src/content/peerconnection/negotiate-timing/js/main.js
alvestrand: Not the only thing I want to change when doing the performance metric, I suspect (I also want per-call timings), so leaving as-is for now.

src/content/peerconnection/negotiate-timing/js/main.js
alvestrand: May be a good idea. The performance of initial negotiation may be different from later negotiation, but both are interesting numbers. Done.

src/content/peerconnection/negotiate-timing/js/main.js
alvestrand: The most important thing to be done here is to release the camera and microphone (in case I need to use them elsewhere). So it makes most sense to enable "start" and disable "call".
This is copied from the previous sample, though; I haven't thought deeply about this. So this is a quick hack.

src/content/peerconnection/negotiate-timing/js/main.js
dogben: Maybe doc that baselineAudioImpairment must be called before measureAudioImpairment and/or that they should be called in pairs.

src/content/peerconnection/negotiate-timing/js/main.js
```diff
@@ -228,5 +221,7 @@ function hangup() {
   localVideo.srcObject = localStream;
```
dogben: Seems like this line should be deleted.

src/content/peerconnection/negotiate-timing/css/main.css
dogben: Did you intend to leave this 2017? IANAL, but generally people ask me to update the copyright when I make a copy.

src/content/peerconnection/negotiate-timing/js/main.js
alvestrand: Yep.

dogben (approved):

PR #1416, opened 2021-01-28
TVPF-5237 : Update the PTZ demo to be controlled by remote controller

NadiAbuMahfouz:


KaptenJansson:
AFAIK not aligned with the intention of the demo. Closing.

PR #1424, opened 2021-02-13
mediarecorder: add dropdown for codec selection

fippo:
for easier testing of issues like
  https://bugs.chromium.org/p/chromium/issues/detail?id=117681
and Safari support

### DIFF
--- a/src/content/getusermedia/record/js/main.js
+++ b/src/content/getusermedia/record/js/main.js
@@ -16,6 +16,8 @@
 let mediaRecorder;
 let recordedBlobs;
 
+const codecPreferences = document.querySelector('#codecPreferences');
+
 const errorMsgElement = document.querySelector('span#errorMsg');
 const recordedVideo = document.querySelector('video#recorded');
 const recordButton = document.querySelector('button#record');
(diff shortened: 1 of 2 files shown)


src/content/getusermedia/record/js/main.js
```diff
@@ -32,7 +34,8 @@ recordButton.addEventListener('click', () => {
 
 const playButton = document.querySelector('button#play');
 playButton.addEventListener('click', () => {
-  const superBuffer = new Blob(recordedBlobs, {type: 'video/webm'});
+  const mimeType = codecPreferences.options[codecPreferences.selectedIndex].value.split(';', 1)[0];
```
fippo, line 38: I would have preferred to use `mediaRecorder.mimeType` here but it is
```
"video/x-matroska;codecs=avc1,opus"
```
when one selects H264

fippo:
@yellowdoge can you please take a look?

src/content/getusermedia/record/js/main.js
yell0wd0g: Correct, and also l.75 won't work because MP4 container is not supported.

These two files should give an idea of what's exactly supported:
- https://source.chromium.org/chromium/chromium/src/+/master:third_party/blink/web_tests/fast/mediarecorder/MediaRecorder-isTypeSupported-avc1.html?q=istypesupported
- https://source.chromium.org/chromium/chromium/src/+/master:third_party/blink/web_tests/fast/mediarecorder/MediaRecorder-isTypeSupported.html

yell0wd0g: Also another note, IIRC the `ondataavailable` event listener's `event.data.type` should carry the actual `mimeType` used in the recording (which might be slightly different from the one specified):

https://source.chromium.org/chromium/chromium/src/+/master:third_party/blink/web_tests/fast/mediarecorder/MediaRecorder-requestData.html;l=68

yell0wd0g (approved):
Changes LGTM % comments left in the code. 
(I'm not sure if I can approve PRs in this repo BTW).

src/content/getusermedia/record/js/main.js
fippo: line 75 is for Safari which presumably only supports MP4 -- see https://webkit.org/blog/11353/mediarecorder-api/
It doesn't work in Chrome but that just means it will be filtered.

e.data.type is the same as mediaRecorder.mimeType

KaptenJansson (approved):

PR #1426, opened 2021-02-18
Add Plan B mode to negotiation-timing page

henbos:
This PR adds a "Start (legacy)" button that runs Plan B, and renames the old start-button to "Start (standard)".

Because transceivers are not available in Plan B, addTrack() is used instead of addTransceiver(). And because of Plan B limitations, the track has to actually exist at SDP offer time, so we can't replace(null) the dummy track immediately - we have to wait until after negotiation to do so.

### DIFF
--- a/src/content/peerconnection/negotiate-timing/js/main.js
+++ b/src/content/peerconnection/negotiate-timing/js/main.js
@@ -9,16 +9,20 @@
 'use strict';
 
 const startButton = document.getElementById('startButton');
+const startButtonPlanB = document.getElementById('startButtonPlanB');
 const callButton = document.getElementById('callButton');
 const renegotiateButton = document.getElementById('renegotiateButton');
 const hangupButton = document.getElementById('hangupButton');
 const log = document.getElementById('log');
 const videoSectionsField = document.getElementById('videoSections');
 
+let sdpSemantics = null;
+
 callButton.disabled = true;
 hangupButton.disabled = true;
 renegotiateButton.disabled = true;
-startButton.onclick = start;
+startButton.onclick = () => { sdpSemantics = "unified-plan"; start(); }
+startButtonPlanB.onclick = () => { sdpSemantics = "plan-b"; start(); }
 callButton.onclick = call;
 renegotiateButton.onclick = renegotiate;
 hangupButton.onclick = hangup;
@@ -27,7 +31,7 @@ let startTime;
 const localVideo = document.getElementById('localVideo');
 const remoteVideo = document.getElementById('remoteVideo');
 
-let audioTransceiver;
+let audioReceiver;
 let audioImpairmentAtStart = 0;
 
 let result;
@@ -53,8 +57,12 @@ remoteVideo.onresize = () => {
 };
 
 let localStream;
+let dummyVideoTrack;
+let dummyVideoStream;
 let pc1;
 let pc2;
+let videoSender;
+let audioSender;
 
 function logToScreen(text) {
   log.append(document.createElement('br'));
@@ -72,6 +80,7 @@ function getOtherPc(pc) {
 async function start() {
   console.log('Requesting local stream');
   startButton.disabled = true;
+  startButtonPlanB.disabled = true;
   const stream = await navigator.mediaDevices
       .getUserMedia({
         audio: true,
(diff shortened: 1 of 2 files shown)


henbos:
Please take a look @alvestrand

src/content/peerconnection/negotiate-timing/js/main.js
```diff
@@ -193,29 +233,47 @@ async function getAudioImpairment(audioTransceiver) {
...
+      if (additionalSenders[i].track) {
+        additionalSenders[i].track.stop();
+        await additionalSenders[i].replaceTrack(null);
+      }
+    }
+  }
   console.log(`Renegotiate finished after ${result.elapsedTime} milliseconds`);
-  const currentVideoTransceiverCount = pc2.getTransceivers().filter(tr => tr.receiver.track.kind == 'video').length;
+  const currentVideoTransceiverCount = pc2.getReceivers().filter(r => r.track.kind == 'video').length;
   result.audioImpairment = await measureAudioImpairment(pc2);
```
alvestrand, line 275: can you move the measurement before the clone cleanup, so that we don't get hit by any audio impact from that part?

src/content/peerconnection/negotiate-timing/js/main.js
```diff
@@ -72,6 +80,7 @@ function getOtherPc(pc) {
 async function start() {
   console.log('Requesting local stream');
   startButton.disabled = true;
+  startButtonPlanB.disabled = true;
```
alvestrand, line 83: hm. this is ugly (when we have start buttons, code referring to them should go with the button-specific thing), but startButton and startButtonPlanB have to know about each other, so perhaps this is the best way to write it. No change needed.

alvestrand (approved):
Looks good. If you fix the format issues to make the linter happy, and move the audio impairment measurement ahead of the cleanup code, I think we should just merge.

fippo:
I thought we had an agreement to not do non-standard stuff in the samples anymore? You might want to use https://github.com/webrtc/test-pages for such things

henbos:
> I thought we had an agreement to not do non-standard stuff in the samples anymore?

The samples pages should definitely not use Plan B as an example of "how to do things". I made an exception this time because this page is a demo for measuring the performance and audio glitches and has been very useful to debug performance issues. The UI displays it as "Start (legacy)" so I hope it is not seen as an encouragement on how to write an application.

henbos:
I'll remove it in a couple of chrome releases.

jan-ivar:
I agree with @fippo here this sends the wrong message. I don't think the word "legacy" is strong enough to counter the impression this was ever a standard.

> ...very useful to debug performance issues

It should be able to serve this same purpose in a personal github.com repo.

jan-ivar (changes requested):
Would rather we not merge this.

hthetiot:
> I'll remove it in a couple of chrome releases.

Can you share a timeline for removing PlanB support. Thx.

henbos:
> Would rather we not merge this.

Ok. Even if we add big red letters saying that this legacy code path will be deleted in August and only exists for temporary debugging purposes?

fippo:
@henbos past promises to fix nonstandard stuff by googlers or google contractors that were broken left a bad impression. I still think the other repo is the way to go, @KaptenJansson will be happy to help :-)

jan-ivar:
> Can you share a timeline for removing PlanB support. Thx.

image

henbos:
I agree that we shouldn't spread Plan B usage on sample pages in general. The reason why I thought it would have been acceptable to add Plan B to this particular page was that...
- The debugging I was going to do on Plan B vs Unified Plan was in the name of removing Plan B from Chromium. Our efforts into optimizing negotiation by profiling both SDP semantics have led to pinpointing issues that we have resolved with substantial improvements in M88-M90. This has been especially good news for Unified Plan.
- This is not a sample page for "how to do things", it is just a performance measurement tool.
- Plan B now triggers a deprecation warning and we have a target removal milestone (M93), lessening the risk that people think Plan B is OK after seeing it on a page like this.

But there seems to be a rough consensus against this PR, so I'll move it to my personal repo.

hthetiot:
> > Can you share a timeline for removing PlanB support. Thx.
> 
> image

Thx @jan-ivar 
I have not seen that yet, is that on chrome stable or canary only for now?

jan-ivar:
> I have not seen that yet, is that on chrome stable or canary only for now?

@hthetiot It was added in M89. See [release notes](https://groups.google.com/g/discuss-webrtc/c/Zrsn2hi8FV0/m/KIbn0EZPBQAJ). Type the following into web console to see it:
```js
new RTCPeerConnection({sdpSemantics: "plan-b"})
```

PR #1433, opened 2021-04-28
Use new canvas/WebGL APIs to draw VideoFrame directly and create VideoFrame from canvas

dogben:
Use new canvas/WebGL APIs to draw VideoFrame directly and create VideoFrame from canvas

The new APIs are supported on Chrome 91+, and the old createImageBitmap API has been removed on Chrome 92. This patch enables the sample to work again on Chrome 92, and fixes an issue on Chrome 90 where the video is vertically flipped.

Tested on Chrome 90, 91, and 92 on Mac and Linux. A previous revision was also tested on Chrome 89.

### DIFF
--- a/src/content/insertable-streams/video-processing/js/canvas-transform.js
+++ b/src/content/insertable-streams/video-processing/js/canvas-transform.js
@@ -14,14 +14,27 @@
  */
 class CanvasTransform { // eslint-disable-line no-unused-vars
   constructor() {
-    // All fields are initialized in init()
-    /** @private {?OffscreenCanvas} canvas used to create the 2D context */
+    /**
+     * @private {?OffscreenCanvas} canvas used to create the 2D context.
+     *     Initialized in init.
+     */
     this.canvas_ = null;
     /**
      * @private {?CanvasRenderingContext2D} the 2D context used to draw the
-     *     effect
+     *     effect. Initialized in init.
      */
     this.ctx_ = null;
+    /**
+     * @private {boolean} If false, pass VideoFrame directly to
+     * CanvasRenderingContext2D.drawImage and create VideoFrame directly from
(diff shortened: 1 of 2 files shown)


src/content/insertable-streams/video-processing/js/canvas-transform.js
```diff
@@ -51,27 +64,53 @@ class CanvasTransform { // eslint-disable-line no-unused-vars
...
+    const timestamp = frame.timestamp;
 
-    ctx.drawImage(inputBitmap, 0, 0);
-    inputBitmap.close();
+    if (!this.use_image_bitmap_) {
+      try {
+        // Supported for Chrome 90+.
+        ctx.drawImage(frame, 0, 0);
+      } catch (e) {
+        // This should only happen on Chrome <90.
```
KaptenJansson, line 74: So there are that many differences between versions resulting in these permutations (referring to the log statements with versions <90, <91 <92, 90+) ?

dogben: No, I just wanted to be precise in my comments. The ImageBitmap path is used for 88-90 and the direct path is used for 91-92. There is one additional wrinkle that the direct path should be supported on 90, and drawImage/texImage2D works fine, but creating a VideoFrame directly from an OffscreenCanvas does not, which is why we have 2 try/catch blocks.

Not sure how much to put in comments. I'm thinking once 91 is stable, I'll just delete the old path and update the text to say it works on 91+.

KaptenJansson: I see. Hard to accommodate for all versions :) I think then this is fine.

KaptenJansson (approved):

PR #1447, opened 2021-06-02
negotiate-timing: allow tweaking max-bundle et al

fippo:
@alvestrand any reason this does not bundle by default?
Either way, this should avoid issues like https://bugs.chromium.org/p/webrtc/issues/detail?id=12837 (in theory, whether it does in practice is a good question)


alvestrand (approved):

alvestrand:
The reason it doesn't bundle by default is twofold:
1) I didn't think of it at the time
2) it's likely that the performance bug this was created to hunt down only manifests when unbundled (and therefore shouldn't have affected users like meet that use bundle-by-default).

It would be interesting to run the tests with/without max-bundle.

PR #1449, opened 2021-06-08
Add async/await when call onIceCandidate

nhvu95:
**Description**
Function "onIceCandidate" is an async function, missing async/await will make it not stable.

**Purpose**


nhvu95:
@googlebot I signed it!

HieuNguyenVu:
This is not the root cause. I will fix it in another branch

PR #1462, opened 2021-07-31
fix: correct error name in handleError

upupming:
See: https://developer.mozilla.org/en-US/docs/Web/API/MediaDevices/getUserMedia#exceptions

**Description**

**Purpose**


upupming:
@googlebot I signed it!

alvestrand (approved):

PR #1463, opened 2021-07-31
fix: this -> rightVideo

upupming:
**Description**

**Purpose**


alvestrand:
Isn't "this" set to the object on which the event is fired?
The change makes it easier to read, but I wonder if it worked before.

upupming:
@alvestrand It doesn't work, see: https://webrtc.github.io/samples/src/content/capture/video-pc/

image

`this`is `window` here:

image

Or we can change the arrow function to an ordinary function to catch this as `rightVideo` itself.

PR #1472, opened 2021-09-06
Add synthetic video generation to "bandwidth" sample, and measure rampup

alvestrand:
**Description**
This allows checking how long it takes for resolution to adapt after
starting the call.

**Purpose**
Verifying that rampup takes too long.

### DIFF
--- a/src/content/peerconnection/bandwidth/js/main.js
+++ b/src/content/peerconnection/bandwidth/js/main.js
@@ -38,6 +38,9 @@ let packetSeries;
 let lastResult;
 
 let lastRemoteStart = 0;
+
+// lastRemoteFullSizeDelay is designed to be picked up by a test script.
+// eslint-disable-next-line no-unused-vars
 let lastRemoteFullSizeDelay = 0;
 
 const offerOptions = {
@@ -48,7 +51,7 @@ const offerOptions = {
 remoteVideo.addEventListener('resize', ev => {
   const elapsed = performance.now() - lastRemoteStart;
   console.log(elapsed, ': Resize event, size ',
-	      remoteVideo.videoWidth, 'x', remoteVideo.videoHeight);
+      remoteVideo.videoWidth, 'x', remoteVideo.videoHeight);
   if (localVideo.videoWidth == remoteVideo.videoWidth &&
       localVideo.videoHeight == remoteVideo.videoHeight) {
     lastRemoteFullSizeDelay = elapsed;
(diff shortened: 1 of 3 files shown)


src/content/peerconnection/bandwidth/js/main.js
```diff
@@ -288,3 +314,43 @@ window.setInterval(() => {
...
+  const ctx = canvas.getContext('2d');
+  const stream = canvas.captureStream();
+
+  let count = 0;
+  setInterval(() => {
+    // Use relatively-prime multipliers to get a color roll
+    const r = Math.abs((count*2)%512 - 256);
+    const g = Math.abs((count*3)%512 - 256);
+    const b = Math.abs((count*5)%512 - 256);
+    ctx.fillStyle = `rgb(${r}, ${g}, ${b})`;
```
dogben, line 338: rgb accepts numbers between 0 and 255 (incl) or percentages from 0% to 100%.

src/content/peerconnection/bandwidth/js/main.js
```diff
@@ -288,3 +314,43 @@ window.setInterval(() => {
...
+    count += 1;
+    const boxSize=80;
+    ctx.fillRect(0, 0, width, height);
+    // Add some bouncing boxes in contrast color to add a little more noise.
+    const rContrast = (r + 128)%255;
+    const gContrast = (g + 128)%255;
+    const bContrast = (b + 128)%255;
```
dogben: Did you perhaps mean 256 rather than 255 here?

dogben (approved):

src/content/peerconnection/bandwidth/js/main.js
alvestrand: I woke up in the night and said to myself "%255 will only go to 254!" :-)
This was code I copied from RTCPeerConnection-helper.js in WPT; it was improper from before.

(Once this is landed, I intend to take the signal generator back into RTCPeerConnection-helper)

src/content/peerconnection/bandwidth/js/main.js
alvestrand: The purpose of the fancy math is to create a signal that changes enough to give the codec something to do, but without to many bandwidth spikes. I may iterate once more over this.

dogben: Ok, so is it intentional to pass negative numbers to rgb? I assume they would be coerced to zero, but I'm not sure.

dogben: Oops, somehow I glossed over the Math.abs.

src/content/peerconnection/bandwidth/js/main.js
```diff
@@ -315,6 +315,13 @@ window.setInterval(() => {
   });
 }, 1000);
 
+// Return a number between 0 and maxValue based on the input number,
+// so that the output changes smoothly up and down.
+function triangle(number, maxValue) {
+  const modulus = (maxValue + 1) * 2;
+  return Math.abs(number % modulus - maxValue);
```
dogben, line 322: triangle(511, 255) returns 256.

dogben (approved):

PR #1477, opened 2021-09-14
Stop using deprecated policies in Network limiter extension

tonyherre:
The deprecated policies nonproxied_udp_enabled and multiple_routes_enabled were removed in crbug.com/982501. As such, we should only check and set the combined enum based policy/pref webRTCIPHandlingPolicy.

Also cleans up the legacy code for mapping from the two deprecated boolean policies to the new enum webRTCIPHandlingPolicy, introduced in M48.

fixes crbug.com/1247217


alvestrand (approved):

pmarks-net:
When I use the published "WebRTC Network Limiter" extension, the last option (Use my proxy server) is disabled, because this change was never actually pushed to production.  Could somebody do that please?

https://chrome.google.com/webstore/detail/webrtc-network-limiter/npeicpdbkakmehahjeeohfdhnlpdklia

PR #1479, opened 2021-09-19
Show more data from ICE candidates

nils-ohlmeier:
**Description**
ICE candidates expose more information, than rendered in this demo.

**Purpose**
Expose the data which is available to people who want to learn about WebRTC.


fippo (approved):
oh yes, we moved on at some point!

KaptenJansson (approved):

PR #1481, opened 2021-09-21
WebGPU sample pages that use Insertable Streams

palak8669:
**Description**
Adding sample page to show how video rendering can be done using WebGPU. These samples use Insertable streams API for input frames.


index.html
```diff
@@ -207,6 +207,12 @@ <h2 id="capture">Insertable Streams:</h2>
             <li><a href="src/content/insertable-streams/video-crop">Video cropping using MediaStream Insertable Streams in a Worker</a></li> (Experimental)
         </ul>
 
+        <h2 id="capture">WebGPU Rendering:</h2>
+        <p class="description">API for Rendering media on GPU</p>
```
tonyherre: How about "Integrations with <a href="https://www.w3.org/TR/webgpu/">WebGPU</a> for custom video rendering"

src/content/insertable-streams/webgpu/index.html
```diff
@@ -0,0 +1,77 @@
+<!DOCTYPE html>
+<!--
+ *  Copyright (c) 2020 The WebRTC project authors. All Rights Reserved.
```
tonyherre: 2021

src/content/insertable-streams/webgpu/index.html
```diff
@@ -0,0 +1,77 @@
...
+    <meta name="viewport" content="width=device-width, user-scalable=yes, initial-scale=1, maximum-scale=1">
+    <meta itemprop="description" content="Client-side WebRTC code samples">
+    <meta itemprop="image" content="../../../images/webrtc-icon-192x192.png">
+    <meta itemprop="name" content="WebRTC code samples">
+    <meta name="mobile-web-app-capable" content="yes">
+    <meta id="theme-color" name="theme-color" content="#ffffff">
+
+    <base target="_blank">
+
+    <title>Insertable Streams - Video</title>
```
tonyherre: Update the title

src/content/insertable-streams/webgpu/index.html
```diff
@@ -0,0 +1,77 @@
...
+
+<body>
+
+    <div id="container">
+        <h1><a href="//webrtc.github.io/samples/" title="WebRTC samples homepage">WebRTC samples</a>
+            <br>
+            <span>WebGPU rendering with insertable streams for multiple video feeds</span>
+        </h1>
+
+        <p>This sample shows how to perform WebGPU rendering on video streams using the
```
tonyherre: "... how to render multiple video streams to a single canvas using the <a ...>Insertable Steams</a> and <a ...>WebGPU</a> APIs"?

src/content/insertable-streams/webgpu/index.html
```diff
@@ -0,0 +1,77 @@
...
+
+        <div>Input:</div>
+        <video id="inputVideo" class="child" width="480" height="270"></video>
+        <video id="gumInputVideo" width="480" height="270"></video>
+
+        <div>Output:</div>
+        <div id="outputVideo" class="worker"></div>
+        <div id="gumOutputVideo" class="worker"></div>
+        <p>
+            <b>Note</b>: This sample is using WebGPU API that is in Origin Trial as
```
tonyherre, line 62: A supported check which displays a bigger warning if WebGPU and BreakoutBox aren't supported (eg too early Chrome, a browser which doesn't support it, or missing the webgpu flag) would be nice - similar to on the BreakoutBox samples but ideally not using alert(). This note is easy to miss if the window is too small to show it.

tonyherre (commented):
The "multi video rendering" option flickers for me occasionally when running on Canary, Mac. Looks like frames for the two videos get swapped? Anything you've seen?
I've not looking into where it comes from yet, I guess a race condition or similar.

src/content/insertable-streams/webgpu/index.html
```diff
@@ -0,0 +1,77 @@
...
+            the command line with
+            <code>--enable-unsafe-webgpu</code>.
+        </p>
+
+    </div>
+
+    <script src="js/main.js" async></script>
+    <script src="js/multi-video.js" async></script>
+    <script src="js/webpgu-worker.js" async></script>
+    <script src="js/worker.js" async></script>
```
dogben: I think you don't want to load your worker script here, right?

src/content/insertable-streams/webgpu/js/main.js
```diff
@@ -0,0 +1,128 @@
...
+    videoElement.play();
+
+    let sourceStream;
+    const mediaPromise = new Promise((resolve, reject) => {
+        videoElement.oncanplay = () => {
+            if (!resolve || !reject) return;
+            console.log('[VideoSource] Obtaining video capture stream');
+            if (videoElement.captureStream) {
+                sourceStream = videoElement.captureStream();
+                resolve("Success");
```
dogben: nit: just `resolve()` since we're ignoring the promise result anyway; also on line 22.

src/content/insertable-streams/webgpu/js/main.js
```diff
@@ -0,0 +1,128 @@
...
+                console.error(e);
+                reject(e);
+            }
+            resolve = null;
+            reject = null;
+        };
+    });
+    await mediaPromise;
+    console.log(
+        '[VideoSource] Received source video stream.',
```
dogben: nit: remove "[VideoSource]" -- I think that's left over from a copy/paste

src/content/insertable-streams/webgpu/js/main.js
```diff
@@ -0,0 +1,128 @@
...
+        `stream_ =`, sourceStream);
```
dogben: nit: remove "stream_ =" -- I think that's also left over from a copy/paste

src/content/insertable-streams/webgpu/js/main.js
```diff
@@ -0,0 +1,128 @@
...
+
+
+let gpuTransform;
+let gUMStream, gUMTrack;
+let gUMVideo;
+
+async function main(sourceType) {
+    gUMStream = await navigator.mediaDevices.getUserMedia({ audio: false, video: { width: 480, height: 270 } });
+    gUMTrack = gUMStream.getVideoTracks()[0];
+    const gum_processor = new MediaStreamTrackProcessor({ track: gUMTrack });
```
dogben: nit: camelCase for all variables

src/content/insertable-streams/webgpu/index.html
```diff
@@ -0,0 +1,77 @@
...
+        <div id="gumOutputVideo" class="worker"></div>
+        <p>
+            <b>Note</b>: This sample is using WebGPU API that is in Origin Trial as
+            of 2021-09-21 and is available in Chrome M94 if the experimental code is enabled on
+            the command line with
+            <code>--enable-unsafe-webgpu</code>.
+        </p>
+
+    </div>
+
```
dogben, line 71: nit: Most of the other samples have a "View source on GitHub" link, which is often useful. Could we add one here?

src/content/insertable-streams/webgpu/js/main.js
```diff
@@ -0,0 +1,128 @@
...
+            if(frame1) frame1.close();
+            if(frame2) frame2.close();
+        }
+    };
+
+    async function updateScreenImage() {
+        while (true) {
+            let chunk1, chunk2;
+            if(video_source){
+                let { value: chunk } = await video_source.read();
```
dogben: Hmm, I don't think we want to read the frames on the main thread for the worker sample. Especially since people will copy this sample for their own code, we should make sure it's the recommended approach, which is to transfer the ReadableStream to the worker and read the frames on the worker.

src/content/insertable-streams/webgpu/js/main.js
```diff
@@ -0,0 +1,128 @@
...
+async function updateSource() {
+    if (gpuTransform) {
+        await gpuTransform.destroy();
+        gpuTransform = null;
+    }
+    destroy_source();
+    const sourceType = sourceSelector.options[sourceSelector.selectedIndex].value;
+
+    console.log("New source is", sourceType);
+    if (sourceType != "stopped") {
```
dogben: nit: !==

src/content/insertable-streams/webgpu/js/multi-video.js
```diff
@@ -0,0 +1,245 @@
...
+            this.canvas_.height = 2500;
+        }
+
+        const canvas = this.canvas_;
+        const context = canvas.getContext('webgpu');
+        this.context_ = context;
+        const adapter = await navigator.gpu.requestAdapter();
+        const device = await adapter.requestDevice();
+        this.device_ = device;
+        if (this.device_ === null) return;
```
dogben: nit: Seems like we should at least log, if not show an error on the page.

src/content/insertable-streams/webgpu/js/multi-video.js
```diff
@@ -0,0 +1,245 @@
...
+            addressModeV: 'repeat',
+            addressModeW: 'repeat',
+            magFilter: 'linear',
+            minFilter: 'linear',
+        });
+        const videoTexture = this.videoTexture_;
+        let videoFrame, videoFrame2;
+        if (frame) {
+            videoFrame = await createImageBitmap(frame, { resizeWidth: 480, resizeHeight: 270 });
+            device.queue.copyExternalImageToTexture(
```
dogben: (No action required) Eventually we'll want to use the video textures directly. At that point, I assume we'll need to make quite a few changes to the shaders so that we can render from those textures.

src/content/insertable-streams/webgpu/js/multi-video.js
```diff
@@ -0,0 +1,245 @@
...
+    async destroy() {
+        if (this.device_) {
+            // Not yet in canary
+            // await this.device_.destroy();
+            this.verticesBuffer_.destroy();
+            this.device_ = null;
+            if (this.canvas_.parentNode) {
+                this.canvas_.parentNode.removeChild(this.canvas_);
+            }
+            console.log('[WebGPUTransform] WebGPU context is lost.',);
```
dogben: nit: I think maybe this log message is copy/pasted? Maybe something like "WebGPUTransform destroyed"?

src/content/insertable-streams/webgpu/js/webpgu-worker.js
```diff
@@ -0,0 +1,80 @@
...
+        const screenCanvas = this.screencanvas_;
+        document.getElementById('outputVideo').append(screenCanvas);
+        const offScreen = screenCanvas.transferControlToOffscreen();
+        this.offscreen_ = offScreen;
+
+        this.screencanvas2_ = document.createElement('canvas');
+        const screenCanvas2 = this.screencanvas2_;
+        document.getElementById('gumOutputVideo').append(screenCanvas2);
+        const offScreen2 = screenCanvas2.transferControlToOffscreen();
+        this.offscreen2_ = offScreen2;
```
dogben: I think maybe I'm missing the point of this sample. If I understand correctly, we're rendering each video feed into a separate canvas, and letting the browser do compositing of the canvases. I don't see a reason this would be better than just using video elements directly, and it seems much more complicated.

I think I'll need to understand the context a bit better before I can review the worker part.

src/content/insertable-streams/webgpu/index.html
```diff
@@ -0,0 +1,78 @@
...
+            <span>Choose type of rendering:</span>
+            <select id="sourceSelector">
+                <option selected value="stopped">(stopped)</option>
+                <option value="main">Main thread</option>
+                <option value="worker">Worker thread</option>
+            </select>
+        </div>
+
+        <div>Input:</div>
+        <video id="inputVideo" class="input" width="480" height="270"></video>
```
tonyherre, line 57: Nit: Add a little padding around these videos to make it visually clear that they're separate video elements?

src/content/insertable-streams/webgpu/js/main.js
```diff
@@ -0,0 +1,123 @@
...
+        };
+    });
+    await mediaPromise;
+    console.log(
+        'Received source video stream.', sourceStream);
+    return sourceStream;
+
+}
+
+async function getUserMediaStream() {
```
tonyherre: As discussed, simplify this to just something like

return  navigator.mediaDevices.getUserMedia({
            audio: false,
            video: { width: 480, height: 270 }
        }).catch(err => {
            throw new Error("Unable to fetch getUserMedia stream " + err);
        });

src/content/insertable-streams/webgpu/js/multi_video_main.js
```diff
@@ -0,0 +1,271 @@
...
+[[binding(1), group(0)]] var myTexture: texture_2d<f32>;
+
+[[stage(fragment)]]
+fn main([[location(0)]] fragUV : vec2<f32>) -> [[location(0)]] vec4<f32> {
+  return textureSample(myTexture, mySampler, fragUV);
+}
+`,
+};
+
+async function updateScreenImage(source) {
```
tonyherre: Can we just make sure source is always not undefined here and skip the check? If it is undefined, this will return an undefined frame and I'm not sure what will happen.

tonyherre: Also, rename to getFrame() or similar - or just inline the await source.read() call?

src/content/insertable-streams/webgpu/js/multi_video_main.js
```diff
@@ -0,0 +1,271 @@
...
+
+        const rectVerts = new Float32Array([
+            1.0, 1.0, 0.0, 1.0, 0.0,
+            1.0, -1.0, 0.0, 1.0, 1.0,
+            -1.0, -1.0, 0.0, 0.0, 1.0,
+            1.0, 1.0, 0.0, 1.0, 0.0,
+            -1.0, -1.0, 0.0, 0.0, 1.0,
+            -1.0, 1.0, 0.0, 0.0, 0.0,
+        ]);
+        //Creates a GPU buffer.
```
tonyherre: Nit: space at start of comment.

src/content/insertable-streams/webgpu/js/multi_video_main.js
```diff
@@ -0,0 +1,271 @@
...
+        this.vertexBuffer_ = null;
+    }
+
+    async init() {
+        console.log('[WebGPUTransform] Initializing WebGPU.');
+
+        if (!this.canvas_) {
+            this.canvas_ = document.createElement('canvas');
+            document.getElementById('outputVideo').append(this.canvas_);
+            this.canvas_.width = 5000;
```
tonyherre: src/css/main.css has a rule `canvas { max-width: 100%; }`, which means explicitly setting width and height here only sets the aspect ratio, so the actual rendered canvas has dimensions of 960x480, and squashed videos. How about just setting it to 960x540 here, so we get the right aspect ratios?

src/content/insertable-streams/webgpu/js/multi_video_worker.js
```diff
@@ -0,0 +1,249 @@
...
+[[binding(1), group(0)]] var myTexture: texture_2d<f32>;
+
+[[stage(fragment)]]
+fn main([[location(0)]] fragUV : vec2<f32>) -> [[location(0)]] vec4<f32> {
+  return textureSample(myTexture, mySampler, fragUV);
+}
+`,
+};
+
+async function updateScreenImage(source) {
```
tonyherre: Ditto above about probably inlining this, or at least renaming it.

src/content/insertable-streams/webgpu/js/multi_video_worker.js
```diff
@@ -0,0 +1,249 @@
...
+        while (true) {
+            const videoFrame = await updateScreenImage(videoSource);
+            const gumFrame = await updateScreenImage(gumSource);
+            renderOnScreen(videoFrame, gumFrame);
+        }
+    }
+    else if (operation === 'destroy') {
+        // Not yet in canary
+        // await device_.destroy();
+        if(device){
```
tonyherre: nit: space between if and paren

src/content/insertable-streams/webgpu/js/multi_video_main.js
palak8669: Do you mean to change `if(source)` to `if (source !== undefined)`  or checking the condition outside this function?

src/content/insertable-streams/webgpu/css/main.css
```diff
@@ -0,0 +1,24 @@
...
+ *  Copyright (c) 2021 The WebRTC project authors. All Rights Reserved.
+ *
+ *  Use of this source code is governed by a BSD-style license
+ *  that can be found in the LICENSE file in the root of the source
+ *  tree.
+ */
+
+ video {
+    width: 480px;
+    height: 270px;
```
dogben: nit: weird indentation

index.html
```diff
@@ -207,6 +207,12 @@ <h2 id="capture">Insertable Streams:</h2>
             <li><a href="src/content/insertable-streams/video-crop">Video cropping using MediaStream Insertable Streams in a Worker</a></li> (Experimental)
         </ul>
 
+        <h2 id="capture">Integrations with <a href="https://www.w3.org/TR/webgpu/">WebGPU</a> for custom video rendering:</h2>
+        <p class="description">API for Rendering media on GPU</p>
+        <ul>
+            <li><a href="src/content/insertable-streams/webgpu">Multi-video rendering using WebRTC Insertable Streams</a></li> (Experimental)
```
dogben: nit: seems like the title is inconsistent between here, the `<title>` tag in index.html, and the `<h1>` tag in index.html. Is that intentional?

src/content/insertable-streams/webgpu/index.html
```diff
@@ -0,0 +1,78 @@
...
+    <link rel="stylesheet" href="../../../css/main.css" />
+    <link rel="stylesheet" href="css/main.css" />
+
+</head>
+
+<body>
+
+    <div id="container">
+        <h1><a href="//webrtc.github.io/samples/" title="WebRTC samples homepage">WebRTC samples</a>
+            <br>
```
dogben: nit: for consistent style with other samples, there shouldn't be a `<br>` here

src/content/insertable-streams/webgpu/index.html
```diff
@@ -0,0 +1,78 @@
...
+            <span>WebGPU rendering with insertable streams for multiple video feeds</span>
+        </h1>
+
+        <p>This sample shows how to render multiple video streams to canvas using the <a
+                href="https://github.com/w3c/mediacapture-insertable-streams">insertable streams</a> and <a
```
dogben: nit: page has moved to https://github.com/w3c/mediacapture-transform

src/content/insertable-streams/webgpu/js/main.js
dogben: Sorry, I can't figure out how to comment on this line without reopening the old comment.

nit: doesn't need to be async

tonyherre (approved):
LGTM

src/content/insertable-streams/webgpu/js/multi_video_main.js
```diff
@@ -0,0 +1,274 @@
...
+        if (!this.canvas_) {
+            this.canvas_ = document.createElement('canvas');
+            document.getElementById('outputVideo').append(this.canvas_);
+            this.canvas_.width = 960;
+            this.canvas_.height = 540;
+        }
+
+        const canvas = this.canvas_;
+        const context = canvas.getContext('webgpu');
+        if (context === null || context === undefined) {
```
dogben: nit: `if (!context)`

Same comment in several other places where you're checking if objects are null or undefined.

src/content/insertable-streams/webgpu/js/multi_video_main.js
```diff
@@ -0,0 +1,274 @@
...
+        const videoSource = videoStream.getReader();
+        const gumSource = gumStream.getReader();
+        if (videoSource === undefined || videoSource === null) {
+            console.log('[WebGPUTransform] videoSource is undefined or null.')
+            return;
+        }
+        if (gumSource === undefined || gumSource === null) {
+            console.log('[WebGPUTransform] gumSource is undefined or null.')
+            return;
+        }
```
dogben: nit: getReader can't return undefined/null; it would throw if there's a problem.

src/content/insertable-streams/webgpu/js/multi_video_main.js
```diff
@@ -0,0 +1,274 @@
...
+
+        while (true) {
+            let { value: videoFrame } = await videoSource.read();
+            let { value: gumFrame } = await gumSource.read();
+            const rendered = await this.renderOnScreen(videoFrame, gumFrame);
+            if(!rendered){
+                break;
+            }
+        }
```
dogben: I wonder if it would be difficult to change things so that we're reading frames independently/concurrently from each stream instead of serially?

src/content/insertable-streams/webgpu/js/multi_video_main.js
```diff
@@ -0,0 +1,274 @@
...
+    }
```
dogben, line 245: nit: we should cancel videoSource and gumSource at the end of this function

src/content/insertable-streams/webgpu/js/multi_video_worker.js
```diff
@@ -0,0 +1,248 @@
...
+        const swapChainFormat = 'bgra8unorm';
+
+        const rectVerts = new Float32Array([
+            1.0, 1.0, 0.0, 1.0, 0.0,
+            -1.0, -1.0, 0.0, 0.0, 1.0,
+            -1.0, 1.0, 0.0, 0.0, 0.0,
+            1.0, 1.0, 0.0, 1.0, 0.0,
+            1.0, -1.0, 0.0, 1.0, 1.0,
+            -1.0, -1.0, 0.0, 0.0, 1.0,
+        ]);
```
dogben: I don't really understand why the coordinates are different between the main thread and the worker. Is there an extra flip happening somewhere? Seems like it might be a Chrome bug.

Other than these coordinates, it seems like this file and multi_video_main.js are very similar. Did you consider passing in a canvas and an error callback to WebGPUTransform so that it's usable on both main thread and worker rather than copy/pasting?

(Just in case the problem was with loading multiple files on the worker, check out https://developer.mozilla.org/en-US/docs/Web/API/WorkerGlobalScope/importScripts.)

src/content/insertable-streams/webgpu/js/multi_video_worker.js
```diff
@@ -0,0 +1,248 @@
...
+        if (videoSource === undefined || videoSource === null) {
+            console.log('[WebGPUWorker] videoSource is undefined or null.')
+            return;
+        }
+        if (gumSource === undefined || gumSource === null) {
+            console.log('[WebGPUWorker] gumSource is undefined or null.')
+            return;
+        }
+
+        while (true) {
```
dogben: nit: need some way to terminate the loop and cancel videoSource/gumSource when we get the 'destroy' operation

src/content/insertable-streams/webgpu/js/multi_video_main.js
```diff
@@ -0,0 +1,274 @@
...
+
+    async destroy() {
+        if (this.device_) {
+            // Not yet in canary
+            // await this.device_.destroy();
```
dogben: Is this comment still correct? I'm not seeing any way to close or destroy a GPU device in current Chrome.

src/content/insertable-streams/webgpu/js/multi_video_main.js
```diff
@@ -0,0 +1,274 @@
...
+            this.vertexBuffer_.destroy();
```
dogben: nit: maybe also destroy this.videoTexture_?

src/content/insertable-streams/webgpu/js/multi_video_worker_manager.js
```diff
@@ -0,0 +1,69 @@
...
+                if (msg.data.error) {
+                    console.log(msg.data.error);
+                    document.getElementById('errorMsg').innerText = msg.data.error;
+                    reject();
+                }
+                if (msg.data.result === 'Done') {
+                    resolve();
+                }
+            });
+        });
```
dogben: I'm a little surprised this works -- I would think we need to add the event listener before posting the `init` message.

src/content/insertable-streams/webgpu/js/multi_video_worker_manager.js
```diff
@@ -0,0 +1,69 @@
...
+        worker.postMessage(
+            {
+                operation: 'destroy',
+            });
+        if (screenCanvas.parentNode) {
+            screenCanvas.parentNode.removeChild(screenCanvas);
+        }
+        const onMessage = new Promise((resolve) => {
+            worker.addEventListener("message", function handleMsgFromWorker() {
+                this.terminate();
```
dogben: nit: avoid use of `this` outside of classes, because it's very confusing what it refers to

src/content/insertable-streams/webgpu/js/multi_video_worker_manager.js
```diff
@@ -0,0 +1,69 @@
...
+                    gumStream: gumStream,
+                }, [videoStream, gumStream]);
+        }
+    }
+
+    async destroy() {
+        worker.postMessage(
+            {
+                operation: 'destroy',
+            });
```
dogben: nit: I would think that terminating the worker is sufficient to clean up everything on the worker. If that's not the case, I wonder if it's actually a bug in Chrome.

dogben (commented):
Thanks for all the updates! Sorry, I added some nits that I missed on the first round.

We should probably have @Kangz review the WebGPU parts when you're ready.

src/content/insertable-streams/webgpu/js/multi_video_main.js
```diff
@@ -0,0 +1,274 @@
+
```
Kangz, line 1: While prototyping some stuff based on Ben's demo page I found it could be simplified a bit to not require a vertex buffer https://paste.googleplex.com/5043473867603968 In general there are a lot of defaults and other WebGPU contruct that can be used to make the code marginally simpler (but it wouldn't make it more efficient).

src/content/insertable-streams/webgpu/js/multi_video_main.js
```diff
@@ -0,0 +1,274 @@
...
+                    },
+                ],
+            },
+            primitive: {
+                topology: 'triangle-list',
+            },
+        });
+
+        this.videoTexture_ = device.createTexture({
+            size: {
```
Kangz: nit: texture defaults mean you only need to specify usage / format / size. And you can do `size: [480 * 2, 270 * 2]` too!

src/content/insertable-streams/webgpu/js/multi_video_main.js
```diff
@@ -0,0 +1,274 @@
...
+        const device = this.device_;
+        if (device === null || device === undefined) {
+            console.log('[WebGPUTransform] device is undefined or null.')
+            if (videoFrame) videoFrame.close();
+            if (gumFrame) gumFrame.close();
+            return false;
+        }
+        const videoTexture = this.videoTexture_;
+        let videoBitmap, gumBitmap;
+        if (videoFrame) {
```
Kangz: note: In the future we should be able to use GPUExternalTexture to avoid extra copies through an ImageBitmap. See the gPaste (though it requires a WIP Chromium CL)

src/content/insertable-streams/webgpu/js/multi_video_main.js
Kangz: It's currently being implemented.

src/content/insertable-streams/webgpu/js/multi_video_main.js
```diff
@@ -163,43 +155,46 @@ class WebGPUTransform {
...
-            gumBitmap.close();
-            gumFrame.close();
+
+        const videoPromise = videoSource.read().then(({ value }) => {
+            this.copyOnTexture(device, videoTexture, value, 0, 270);
+        });
+        const gumPromise = gumSource.read().then(({ value }) => {
+            this.copyOnTexture(device, videoTexture, value, 480, 0);
+        });
+        await videoPromise, gumPromise;
```
dogben: I'm no expert on JS, but I think this is only awaiting videoPromise, not gumPromise. You probably want `await Promise.all(videoPromise, gumPromise);`.

src/content/insertable-streams/webgpu/js/multi_video_main.js
```diff
@@ -163,43 +155,46 @@ class WebGPUTransform {
...
+
+        if (!this.device_) {
+            return false;
```
dogben: nit: I'd say this deserves a comment, e.g. "Check if destroy has been called asynchronously"

dogben (commented):
One little bug, otherwise LGTM!

src/content/insertable-streams/webgpu/js/multi_video_worker.js
```diff
@@ -1,18 +1,24 @@
-importScripts('./multi_video_main.js')
+// eslint-disable-next-line no-undef
+importScripts('./multi_video_main.js');
```
dogben: You should probably add importScripts to https://github.com/webrtc/samples/blob/gh-pages/.eslintrc.js

It might be worthwhile to add the WebGPU globals there as well -- looks like it's just GPUBufferUsage that's needed here.

dogben (approved):

PR #1491, opened 2021-11-04
change-codecs: show encoder implementation

fippo:
which might be useful feedback for issues like
  https://bugs.chromium.org/p/chromium/issues/detail?id=1237677


alvestrand (approved):

alvestrand:
do all codecs have implementation strings?

fippo:
For video yes. For vpx that is either "ExternalEncoder" or "libvpx", for H264 "ExternalEncoder" or "OpenH264" and for AV1 libaom (and similar strings for the decoder)
For audio it is not implemented but I don't think there is much choice there anyway. The question this answers is "are we decoding in hardware" or not.

PR #1502, opened 2021-12-15
crop: update to use frame copy constructor

fippo:
with visibleRect. See
  https://github.com/w3c/webcodecs/issues/281
(but not the small differences between the proposal and the actual api)

### DIFF
--- a/src/content/insertable-streams/video-crop/js/worker.js
+++ b/src/content/insertable-streams/video-crop/js/worker.js
@@ -8,17 +8,16 @@
 
 'use strict';
 
-const canvas = new OffscreenCanvas(640, 360);
-const ctx = canvas.getContext('2d', {desynchronized: true});
-
 function transform(frame, controller) {
-  // Cropping is a bit too complex. See https://github.com/w3c/webcodecs/issues/281
-  // See https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/drawImage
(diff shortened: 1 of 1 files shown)


fippo:
@dogben PTAL

fippo:
not really this sample but I wonder if we need to explain [sample aligned](https://w3c.github.io/webcodecs/#videoframe-copyto-options) which creates some restrictions on the x/y (making them divisible by 4 works from what I can see)

dogben (approved):

dogben:
> not really this sample but I wonder if we need to explain [sample aligned](https://w3c.github.io/webcodecs/#videoframe-copyto-options) which creates some restrictions on the x/y (making them divisible by 4 works from what I can see)

Agree it's not an ideal API. I think if the crop/scale constructor were extended to handle all CanvasImageSources then we would almost necessarily need to drop the sample-aligned requirement. I'm hoping that's the direction we'll head in; otherwise we may want to have accessors for alignmentX/Y.

dogben:
> not really this sample but I wonder if we need to explain [sample aligned](https://w3c.github.io/webcodecs/#videoframe-copyto-options) which creates some restrictions on the x/y (making them divisible by 4 works from what I can see)

Forgot to say -- I think it's supposed to work if they're divisible by 2. If you're seeing different behavior, it may be worth filing a bug.

fippo:
2 works too, thank you!

PR #1503, opened 2021-12-17
add broadcastchannel example

fippo:
showing actual signalling between two tabs without the need for a signaling server

( i miss apprtc)

### DIFF
--- a/src/content/peerconnection/channel/index.html
+++ b/src/content/peerconnection/channel/index.html
@@ -1,0 +1,63 @@
+<!DOCTYPE html>
+<!--
+ *  Copyright (c) 2021 The WebRTC project authors. All Rights Reserved.
+ *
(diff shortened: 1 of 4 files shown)


src/content/peerconnection/channel/index.html
```diff
@@ -0,0 +1,63 @@
...
+
+</head>
+
+<body>
+
+<div id="container">
+    <h1><a href="//webrtc.github.io/samples/" title="WebRTC samples homepage">WebRTC samples</a>
+        <span>Peer connection</span></h1>
+
+    <p>This sample shows how to setup a connection between two peers in different using
```
alvestrand: "in different using" -> "in different tabs using"

alvestrand (approved):
Can it really be that simple? :-)

PR #1511, opened 2022-01-25
Remove unneeded offerToReceive options

youennf:
These options are unnecessary and break Safari: the audio m-section becomes sendrecv instead of recvonly (Maybe due to some incompatibility between Safari and the shim).

### DIFF
--- a/src/content/capture/canvas-pc/js/main.js
+++ b/src/content/capture/canvas-pc/js/main.js
@@ -15,10 +15,6 @@ const video = document.querySelector('video');
 
 let pc1;
 let pc2;
-const offerOptions = {
-  offerToReceiveAudio: 1,
-  offerToReceiveVideo: 1
-};
 
 let startTime;
 
@@ -78,7 +74,7 @@ function call() {
   console.log('Added local stream to pc1');
 
   console.log('pc1 createOffer start');
-  pc1.createOffer(onCreateOfferSuccess, onCreateSessionDescriptionError, offerOptions);
+  pc1.createOffer(onCreateOfferSuccess, onCreateSessionDescriptionError);
 }
 
 function onCreateSessionDescriptionError(error) {


alvestrand:
Why would they break safari?
The purpose of the options is to ensure the m-sections become sendrecv and not sendonly; whether there's a good reason for doing that is something I don't know - but that shouldn't break anything.

youennf:
> The purpose of the options is to ensure the m-sections become sendrecv and not sendonly

Isn't the purpose of these options to have recvonly m-sections if no corresponding track is added?
From what I saw in Safari, adapter.js shim makes the corresponding audio m-section as sendrecv while other browsers are creating recvonly m-section.

I do not think this particular test is about testing that particular behavior so it seems fine to drop these options, right?
I do not understand why this page is negotiating an audio m-section while the sole purpose is exchanging video.
Just removing the offerToReceiveAudio value probably fixes the page on Safari.

> but that shouldn't break anything.

In Safari case, two track events are fired, which messes up the track event handler.
This in terms breaks setting the video.srcObject and the teapot does not show up.
In other browsers, only one track event is fired.

youennf:
Ping review?

alvestrand:
Hm. This rathole goes deeper .... setting options on the offer should not affect the number of stream events at the answerer.
This code *should* result in two m-sections in the SDP, but the audio section should be recvonly in the offer and either sendonly or inactive in the answer - receiving a recvonly m-section shouldn't cause a track event on the answerer.

It's also possible that multiple track events are fired in Chrome too, but that the two tracks generated have the same stream ID, in which case the code will not change the stream connected to the <video> tag.

I'm approving this CL because the argument is pointless, but I think there is a real Safari bug here causing the discrepancy.

alvestrand (approved):

youennf:
> I think there is a real Safari bug here causing the discrepancy.

offerToReceiveAudio is not implemented in Safari but shimmed.
Looking deeper, it seems to work fine with https://webrtchacks.github.io/adapter/adapter-latest.js but not https://webrtc.github.io/adapter/adapter-latest.js

Would it be possible to bump adapter's version? Can we merge that PR anyway?

alvestrand:
Merging. Can you make a PR for updating the shim?
(it's been so long, I've forgotten how to do that)

Issue #1514, opened 2022-02-09
Safari w/ Headphones capture failure in sample due to AudioContext not releasing resources


maikthomas:
- [x] I have provided steps to reproduce
- [x] I have provided browser name and version
- [x] I have provided a link to the sample here or a modified version thereof

## Browser affected

Safari Version 15.0 (16612.1.29.41.4, 16612) on Mac

## Description
When running the Audio stream volume example: https://webrtc.github.io/samples/src/content/getusermedia/volume/
On Safari with Bluetooth headphones, if you stop and start the audio track is not recaptured and you will see the error:
`A MediaStreamTrack ended due to a capture failure`

I believe this is due to [`AudioContext.close()`](https://developer.mozilla.org/en-US/docs/Web/API/AudioContext) not being called.
This doc for method says:
>Closes the audio context, releasing any system audio resources that it uses.

So it seems the bluetooth resource is not being released.

I have added a call to `audioContext.close()` and the issue appears to be fixed.
Note that since this only happens on Safari and with bluetooth devices, we have also opened a ticket to WebKit:
https://bugs.webkit.org/show_bug.cgi?id=236350

## Steps to reproduce
- On a Mac (haven't tested on windows)
- Connect bluetooth headphones as default audio device
- Open in safari: https://webrtc.github.io/samples/src/content/getusermedia/volume/
- Click Start, Stop, and Start again
- Observe in console: `A MediaStreamTrack ended due to a capture failure`

## Expected results
- Stream and be started and stopped freely with audioMeter working

## Actual results
- Stream and audioMeter only work first time before become permanently broken with audio track unavailable

fippo:
fixed!

PR #1515, opened 2022-02-09
Add call to AudioContext.close() to free up bluetooth device

maikthomas:
**Description**
As detailed in #1514, bluetooth resources are not released by the AudioContext in Safari when using bluetooth headphones.
This PR adds a manual call to `audioContext.close()` to release the resource.

**Purpose**
Fix getusermedia/volume demo for safari users with bluetooth headphones

### DIFF
--- a/src/content/getusermedia/volume/js/main.js
+++ b/src/content/getusermedia/volume/js/main.js
@@ -82,6 +82,7 @@ function stop() {
 
   window.stream.getTracks().forEach(track => track.stop());
   window.soundMeter.stop();
+  window.audioContext.close();
   clearInterval(meterRefresh);
   instantMeter.value = instantValueDisplay.innerText = '';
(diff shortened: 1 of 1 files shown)


alvestrand:
I'd consider this a workaround for a Safari bug, but that's life.
However, you need to do the CLA dance before contributing (see instructions at top level).

maikthomas:
@alvestrand I agree, we also opened this as a bug on WebKit.
I guess it depends on whether it's good practice to manually close the audioContext or not?

CLA updated btw

alvestrand:
The main reason for closing contexts (as we do religiously in web platform tests, for instance) is to avoid running out of resources
Here I don't think that's the problem. But I don't see a problem with doing it either, so approving.

alvestrand (approved):

PR #1521, opened 2022-02-23
test: get rid of gulp for linting

fippo:
because 2022.


package.json
```diff
@@ -19,32 +19,18 @@
     "url": "https://github.com/webrtc/samples.git"
   },
   "scripts": {
-    "test": "gulp",
-    "start": "http-server . -c-1"
+    "start": "http-server . -c-1",
+    "test": "npm run eslint && npm run stylelint",
+    "eslint": "eslint 'src/content/**/*.js'",
```
fippo, line 24: note the single quotes here. This is cross-os glob matching, not shell glob matching

fippo:
Note that nightwatch has been broken forever (while passing the tests)
It should go away too... sadly I don't have time to restore the old tests.

alvestrand (approved):

PR #1524, opened 2022-02-24
fix autoplay woes detection in canvas-pc for  firefox

fippo:
paused is initially true there, listening for the loadedmetadata
event allows hiding the text again

follow-up to #1513


fippo:
@alvestrand I ported one test to the new [mocha-based](https://www.npmjs.com/package/mocha) setup:
  src/content/peerconnection/trickle-ice/js/test.js
mocha is similar to jasmine used by apprtc once upon a time but more popular

KaptenJansson (approved):
LGTM

PR #1528, opened 2022-02-28
bring back selenium-based e2e tests

fippo:
removes the "new" tests, re-instantiates the old ones and then replace tape with mocha because 2022


fippo:
@alvestrand I ported one test to the new [mocha-based](https://www.npmjs.com/package/mocha) setup:
src/content/peerconnection/trickle-ice/js/test.js
mocha is similar to jasmine used by apprtc once upon a time but more popular

KaptenJansson (approved):
Nostalgia :) LGTM with some indentation nits.

fippo:
heh. Will fix those when porting the other tests but I first wanted some feedback on how the test tooling looks in general :-)

alvestrand:
The approach has my approval!

fippo:
a couple of things to note:
- firefox currently doesn't work since executeScript can not access the ES6 const/let variables for some scoping reasons I don't understand
~- I plan to squash the "new tests" a bit more~
- the ice restart behavior seems broken, i'll run a bisect
~- tests are not running yet~

fippo:
looks ready now. I recommend reviewing the individual commits. Some tests are deactivated, partly due to flakyness but that will be easier to look at in separate changes.

src/content/peerconnection/upgrade/js/test.js
```diff
@@ -15,7 +15,7 @@ let driver;
 const path = '/src/content/peerconnection/upgrade/index.html';
 const url = `${process.env.BASEURL ? process.env.BASEURL : ('file://' + process.cwd())}${path}`;
 
-describe('peerconnection upgrade from audio-only to audio-video', () => {
+xdescribe('peerconnection upgrade from audio-only to audio-video', () => {
```
alvestrand: I take it x-prefixing is the accepted idiom for "disable"? Or does this need a comment?

src/content/datachannel/datatransfer/js/test.js
```diff
@@ -1,20 +1,57 @@
...
+        .sendKeys(megsToSend + '\n');
+
+    await driver.findElement(webdriver.By.id('sendTheData')).click();
+
+    await Promise.all([
+      driver.wait(() => driver.executeScript(() => {
+        return localConnection && localConnection.connectionState === 'connected'; // eslint-disable-line no-undef
+      })),
+      await driver.wait(() => driver.executeScript(() => {
+        return remoteConnection && remoteConnection.connectionState === 'connected'; // eslint-disable-line no-undef
```
alvestrand, line 45: that's a lot of no-undef comments. is remoteConnection that hard to see?

alvestrand:
Looks good for the parts I understand. Running tests beat non-running tests!

src/content/peerconnection/upgrade/js/test.js
fippo: yes. I should try describe.skip which is more readable

src/content/datachannel/datatransfer/js/test.js
fippo: it is more :exploding_head: 

The tests are linted in the node environment (see the eslint-env at the top) but the executeScript is actually run in the browser. The linter can not know that the variables are defined in the main.js file hence these suppression for no-undef is  needed.

The alternative would be to turn the "in browser" tests into strings (which is how it was done in the past) but those can't be linted at all.

alvestrand: I think I've seen comment syntax that declares the variables to Lint.
Second answer in https://stackoverflow.com/questions/41552041/global-variables-in-javascript-and-eslint says tht

/* global remoteConnection */

should be enough to make Lint shut up (and still catch other misspelled variables on that line, if any).

fippo: [global](https://eslint.org/docs/user-guide/configuring/language-options#specifying-globals) is just that - a global in the node environment which we don't want.

https://eslint.org/docs/rules/no-undef is more close to the use-case. Risk should be minimal since the test needs to work in the end. This should be sufficient to protect us from my butterfingers typing a utf8 character in the wrong place ;-)

fippo:
going to merge, :crossed_fingers: the tests are reasonably stable

PR #1534, opened 2022-04-04
ice-restart: fix functionality

fippo:
by listening to the transport selectedcandidatepairchange event.
This sample has been relying on the somewhat buggy but useful behavior
of the "legacy" iceconnectionstatechange firing a "connected" event even
if the previous state was already connected.

This behavior is no longer supported in the spec iceconnectionstatechange
event. Alternatively listen to the selectedcandidatepairchange event and
determine the availability of new candidates based on that

### DIFF
--- a/package.json
+++ b/package.json
@@ -30,7 +30,7 @@
   ],
   "devDependencies": {
     "chai": "^4.3.6",
-    "chromedriver": "^98.0.1",
+    "chromedriver": ">98.0.1",
     "eslint": "^8.9.0",
     "eslint-config-google": "^0.14.0",
     "geckodriver": "^3.0.1",
--- a/package.json
+++ b/package.json
@@ -22,7 +22,7 @@
     "start": "http-server . -c-1",
     "test": "npm run eslint && npm run stylelint",
     "eslint": "eslint 'src/content/**/*.js'",
-    "mocha": "mocha 'src/content/**/test.js'",
+    "mocha": "mocha --timeout 5000 'src/content/**/test.js'",
     "stylelint": "stylelint 'src/**/*.css'"
   },
   "eslintIgnore": [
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -30,7 +30,7 @@
   ],
   "devDependencies": {
     "chai": "^4.3.6",
-    "chromedriver": "^98.0.1",
```
fippo, line 33: drive-by fix, otherwise chromedriver gets stuck on a version incompatible with the chrome version that gets installed

src/content/datachannel/filetransfer/js/test.js
```diff
@@ -15,7 +15,7 @@ let driver;
 const path = '/src/content/datachannel/filetransfer/index.html';
 const url = `${process.env.BASEURL ? process.env.BASEURL : ('file://' + process.cwd())}${path}`;
 
-describe('datachannel basic', () => {
+describe('datachannel filetransfer', () => {
```
fippo, line 18: drive-by too

src/content/peerconnection/restart-ice/js/main.js
```diff
@@ -32,6 +32,8 @@ remoteVideo.addEventListener('loadedmetadata', function() {
   console.log(`Remote video videoWidth: ${this.videoWidth}px,  videoHeight: ${this.videoHeight}px`);
 });
 
+const useTransportIceEvent = window.RTCIceTransport && 'onselectedcandidatepairchange' in RTCIceTransport.prototype;
```
alvestrand: There are many ICE events (I think).
What about calling it "useTransportCandidatePairChange"? Verbose, but accurate.

alvestrand (approved):

src/content/peerconnection/restart-ice/js/main.js
fippo: useSelectedCandidatePairChange

PR #1535, opened 2022-04-05
Allow `stuns:` URIs in the trickle ICE sample

rhansen:
The `stuns` URI scheme is defined in https://www.rfc-editor.org/rfc/rfc7064.html#section-3

### DIFF
--- a/src/content/peerconnection/trickle-ice/js/main.js
+++ b/src/content/peerconnection/trickle-ice/js/main.js
@@ -83,7 +83,7 @@ function selectServer(event) {
 
 function addServer() {
   const scheme = urlInput.value.split(':')[0];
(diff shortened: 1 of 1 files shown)


alvestrand (approved):
Looks good to me. Note that allowing it in the sample is no guarantee of browser support, but being able to enter the URL is required to test that, so it's goodness anyway.

fippo:
iirc the reason we do the filtering and alerting manually is that we don't catch failures to construct the peerconnection - chrome throws when it encounters `stuns:`

rhansen:
I pushed a commit to display the Chrome error.

fippo:
thank you for splitting up the commits nicely too!

PR #1538, opened 2022-04-06
test: update selenium to selenium@4

fippo:
which is the current stable version which also allows fancy things like
controlling multiple windows


KaptenJansson (approved):
LGTM

PR #1540, opened 2022-04-06
test: try three times

fippo:
hoping that this increases stability

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,7 +22,7 @@
     "start": "http-server . -c-1",
     "test": "npm run eslint && npm run stylelint",
     "eslint": "eslint 'src/content/**/*.js'",
-    "mocha": "mocha --timeout 5000 'src/content/**/test.js'",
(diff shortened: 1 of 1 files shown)


alvestrand:
The experience in WebRTC and Chrome is that this decreases flakiness to a level where it hides code that is actually wrong and needs to be fixed.....  I suspect we'd be better off in many cases by disabling the corresponding tests and filing bugs on them.

fippo:
to be fair, chromium CQ will automatically retry once, no?

alvestrand:
actually I believe Chromium CQ will try 3 times on each bot run, and
reschedule the bot run if it fails, so at least 6 times.
The list of flaky tests is *horribly* long in Chrome, which is a reason why
I don't want to slide towards that state.

fippo:
true. I'll try to make this work with a single driver instance - creating and closing so many instances may well overload the CI bots

fippo:
closing for now

PR #1546, opened 2022-04-14
states: add e2e tests

fippo:
and fix misc bugs and terminology

### DIFF
--- a/src/content/peerconnection/states/js/test.js
+++ b/src/content/peerconnection/states/js/test.js
@@ -1,0 +1,79 @@
+/*
+ *  Copyright (c) 2022 The WebRTC project authors. All Rights Reserved.
+ *
+ *  Use of this source code is governed by a BSD-style license
+ *  that can be found in the LICENSE file in the root of the source
+ *  tree.
+ */
+/* eslint-env node, mocha */
+
+'use strict';
+const webdriver = require('selenium-webdriver');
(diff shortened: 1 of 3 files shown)


fippo:
@alvestrand @henbos there is an interesting thing about this sample that I only noticed when writing the test:
signaling state and ice connection state get a state change to "new/stable" (from "new/stabe"), connectionState does not.
That seems inconsistent...

henbos:
Yeah I think we should calculate connectionState the same way we calculate iceConnectionState, in blink, so that it matches the JS exposed transport states.

But incidentally connectionState still exposes the ”failed” heuristic that was removed from iceConnectionState.

My kingdom for an end-of-candidates

fippo:
ah... turns out there was an extra call adding the 'new' / 'stable' state to the div. Not a spec issue, a sample issue. Updated the sample and the test.

KaptenJansson (approved):

src/content/peerconnection/states/js/test.js
```diff
@@ -0,0 +1,79 @@
+/*
+ *  Copyright (c) 2015 The WebRTC project authors. All Rights Reserved.
```
KaptenJansson: Update copyright year?

KaptenJansson (approved):

PR #1553, opened 2022-04-25
datachannel: add example for two tabs

### LINKED ISSUE #1325: No Example To Run Datachannel Peer To Peer
Installing DataChannel Basic Sample, I managed to get it to text chat itself. However, there was no way to load this peer to peer, such as two different computers on two different IP addresses.

fippo:
using broadcast channel like for
  https://webrtc.github.io/samples/src/content/peerconnection/channel/

Fixes #1325


fippo:
I wonder if we should remove https://webrtc.github.io/samples/src/content/datachannel/messaging/ -- it ends up being rather confusing by trying to do things in a single tab

fippo:
looks like the other sample is bust as has been for a while without anyone noticing which suggests it is safe to remove.

KaptenJansson (approved):
LGTM

PR #1560, opened 2022-06-21
permission: need to call resume()

hakkisabah:
**Description**
In this page : https://developer.chrome.com/blog/autoplay/#web-audio
When we read the relevant page, we may need to make changes from the
beginning, so we ask the user what to do before the AudioContext is
called.

**Purpose**
`"The AudioContext was not allowed to start. It must be resumed (or created) after a user gesture on the page."` It was not working because of this error, it has been fixed.


alvestrand:
The patch does not call resume(). Does it call something that calls resume()?
If not, the title of the patch should be changed.

PR #1561, opened 2022-06-22
trickle-ice: show url and relayProtocol if present

fippo:
note: this will work in M105+ once
  https://chromiumdash.appspot.com/commit/1fe14f2752a35932bea6d6ccff7433a6716c6391
is available

### DIFF
--- a/src/content/peerconnection/trickle-ice/js/main.js
+++ b/src/content/peerconnection/trickle-ice/js/main.js
@@ -258,10 +258,10 @@ async function iceCallback(event) {
     appendCell(row, candidate.address);
     appendCell(row, candidate.port);
     appendCell(row, formatPriority(candidate.priority));
+    appendCell(row, candidate.relayProtocol || '');
     appendCell(row, candidate.sdpMid);
     appendCell(row, candidate.sdpMLineIndex);
     appendCell(row, candidate.usernameFragment);
-    appendCell(row, candidate.relayProtocol || '');
     appendCell(row, candidate.url || url || '');
     candidates.push(candidate);
   }
--- a/src/content/peerconnection/trickle-ice/js/main.js
+++ b/src/content/peerconnection/trickle-ice/js/main.js
@@ -238,12 +238,15 @@ async function iceCallback(event) {
       return;
     }
     const {candidate} = event;
-    let candidateStats;
-    if (['srflx', 'relay'].includes(candidate.type)) {
+    let url;
+    // Until url is available from the candidate, to to polyfill.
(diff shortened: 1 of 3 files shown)


src/content/peerconnection/trickle-ice/js/main.js
```diff
@@ -247,6 +245,16 @@ function iceCallback(event) {
       return;
     }
     const {candidate} = event;
+    const stats = await pc.getStats();
+    let candidateStats;
+    if (['srflx', 'relay'].includes(candidate.type)) {
+      stats.forEach(report => {
+        if (!candidateStats && report.type === 'local-candidate' && report.address === candidate.address && report.port === candidate.port) {
```
fippo: I /think/ this sufficient to match the candidate

fippo:
@henbos @alvestrand PTAL. The API inconsistency bothers me, I don't see a reason not to expose the same properties in onicecandidates as we do in stats.

src/content/peerconnection/trickle-ice/js/main.js
```diff
@@ -247,6 +245,16 @@ function iceCallback(event) {
       return;
     }
     const {candidate} = event;
+    const stats = await pc.getStats();
```
henbos: nit: move "await pc.getStats()" to inside the condition to avoid unnecessarily polling getStats

src/content/peerconnection/trickle-ice/js/main.js
```diff
@@ -240,6 +238,16 @@ function iceCallback(event) {
       return;
     }
     const {candidate} = event;
+    let candidateStats;
+    if (['srflx', 'relay'].includes(candidate.type)) {
+      const stats = await pc.getStats();
+      stats.forEach(report => {
+        if (!candidateStats && report.type === 'local-candidate' && report.address === candidate.address && report.port === candidate.port) {
+          candidateStats = report;
```
alvestrand: should there be a 'break' here, or is there some fear that we'll want to match the last one not the first one?

alvestrand (approved):
I assume this will be OBE once the URL is available in the candidate, but needed for 3 months while the next version rolls out.
Is the relay protocol available in stats and should be extracted the same way?

src/content/peerconnection/trickle-ice/js/main.js
fippo: you can't break a .forEach :-(

fippo:
relayProtocol is polyfilled in Chrome/Safari in adapter-latest.js as of yesterday!

fippo:
changed to actually have a built-in expiry date so the getStats call no longer happens when url is there

PR #1572, opened 2022-09-08
trickle-ice: remove icecandidatepoolsize option

fippo:
since this makes no sense for this case as we only gather a single candidate immediately after creating the connection


KaptenJansson (approved):

KaptenJansson:
LGTM

PR #1573, opened 2022-09-21
Make the gDM example request audio

eladalon1983:
As a Chromium committer, I use this page often to trigger gDM. I know many others who do the same. And we often need audio, which then requires us to use the DevTools console. Or worse, we interact with someone trying to reproduce an issue, and they report that "audio has now also stopped working" after interacting with this page.


fippo (approved):

fippo:
i herewith delegate the 🍰 you owe for the review to @KaptenJansson

eladalon1983:
The cake is a lie.

fippo:
But promises were made about different rules...

eladalon1983:
I am rejecting the `Promise` with an `InvalidCharacterError`.

PR #1575, opened 2022-09-30
getdisplaymedia: add displaysurface

fippo:
described in https://developer.chrome.com/docs/web-platform/screen-sharing-controls/

@eladalon1983 it seems https://developer.chrome.com/docs/web-platform/screen-sharing-controls/#displaySurface has displaySurface at the wrong level, no?

### DIFF
--- a/src/content/getusermedia/getdisplaymedia/index.html
+++ b/src/content/getusermedia/getdisplaymedia/index.html
@@ -37,7 +37,7 @@
     <button id="startButton" disabled>Start</button>
     <select id="preference" style="display:none">
       <option value="" selected>Share the default</option>
-      <option value="browser">Share a browser window</option>
+      <option value="browser">Share a browser tab</option>
       <option value="window">Share another window</option>
       <option value="monitor">Share the entire screen</option>
     </select>
--- a/src/content/getusermedia/getdisplaymedia/index.html
+++ b/src/content/getusermedia/getdisplaymedia/index.html
@@ -36,7 +36,7 @@
     <video id="gum-local" autoplay playsinline muted></video>
     <button id="startButton" disabled>Start</button>
     <select id="displaySurface" style="display:none">
-      <option value="" selected>Share the default</option>
+      <option value="default" selected>Share the default</option>
       <option value="browser">Share a browser tab</option>
       <option value="window">Share a window</option>
       <option value="monitor">Share the entire screen</option>
--- a/src/content/getusermedia/getdisplaymedia/index.html
+++ b/src/content/getusermedia/getdisplaymedia/index.html
@@ -35,12 +35,15 @@
 
     <video id="gum-local" autoplay playsinline muted></video>
     <button id="startButton" disabled>Start</button>
-    <select id="displaySurface" style="display:none">
-      <option value="default" selected>Share the default</option>
-      <option value="browser">Share a browser tab</option>
-      <option value="window">Share a window</option>
-      <option value="monitor">Share the entire screen</option>
-    </select>
+    <fieldset id="options" style="display:none">
+      <legend>Advanced options</legend>
+      <select id="displaySurface">
+        <option value="default" selected>Show default sharing options</option>
+        <option value="browser">Prefer to share a browser tab</option>
+        <option value="window">Prefer to share a window</option>
+        <option value="monitor">Prefer to share the entire screen</option>
+      </select>
+    </fieldset>
     <div id="errorMsg"></div>
 
     <p>Display the screensharing stream from <code>getDisplayMedia()</code> in a video element.</p>
--- a/src/content/getusermedia/getdisplaymedia/index.html
+++ b/src/content/getusermedia/getdisplaymedia/index.html
@@ -41,7 +41,7 @@
         <option value="default" selected>Show default sharing options</option>
         <option value="browser">Prefer to share a browser tab</option>
         <option value="window">Prefer to share a window</option>
-        <option value="monitor">Prefer to share the entire screen</option>
+        <option value="monitor">Prefer to share an entire screen</option>
       </select>
     </fieldset>
     <div id="errorMsg"></div>
--- a/src/content/getusermedia/getdisplaymedia/index.html
+++ b/src/content/getusermedia/getdisplaymedia/index.html
@@ -35,7 +35,7 @@
 
     <video id="gum-local" autoplay playsinline muted></video>
     <button id="startButton" disabled>Start</button>
-    <select id="preference" style="display:none">
+    <select id="displaySurface" style="display:none">
       <option value="" selected>Share the default</option>
       <option value="browser">Share a browser tab</option>
       <option value="window">Share a window</option>
--- a/src/content/getusermedia/getdisplaymedia/index.html
+++ b/src/content/getusermedia/getdisplaymedia/index.html
@@ -35,7 +35,12 @@
 
     <video id="gum-local" autoplay playsinline muted></video>
     <button id="startButton" disabled>Start</button>
-
+    <select id="preference" style="display:none">
+      <option value="" selected>Share the default</option>
+      <option value="browser">Share a browser window</option>
+      <option value="window">Share another window</option>
+      <option value="monitor">Share the entire screen</option>
+    </select>
     <div id="errorMsg"></div>
 
     <p>Display the screensharing stream from <code>getDisplayMedia()</code> in a video element.</p>
--- a/src/content/getusermedia/getdisplaymedia/index.html
+++ b/src/content/getusermedia/getdisplaymedia/index.html
@@ -38,7 +38,7 @@
     <select id="preference" style="display:none">
       <option value="" selected>Share the default</option>
       <option value="browser">Share a browser tab</option>
-      <option value="window">Share another window</option>
+      <option value="window">Share a window</option>
       <option value="monitor">Share the entire screen</option>
     </select>
     <div id="errorMsg"></div>
--- a/src/content/getusermedia/getdisplaymedia/js/main.js
+++ b/src/content/getusermedia/getdisplaymedia/js/main.js
@@ -8,6 +8,7 @@
 'use strict';
 
 const preferredDisplaySurface = document.getElementById('displaySurface');
+const startButton = document.getElementById('startButton');
 
 if (adapter.browserDetails.browser === 'chrome' &&
     adapter.browserDetails.version >= 107) {
@@ -46,12 +47,12 @@ function errorMsg(msg, error) {
   }
 }
 
-const startButton = document.getElementById('startButton');
+
 startButton.addEventListener('click', () => {
-  let options = {audio: true, video: true};
+  const options = {audio: true, video: true};
   const displaySurface = preferredDisplaySurface.options[preferredDisplaySurface.selectedIndex].value;
   if (displaySurface !== 'default') {
-    options.video = { displaySurface };
+    options.video = {displaySurface};
   }
   navigator.mediaDevices.getDisplayMedia(options)
       .then(handleSuccess, handleError);
--- a/src/content/getusermedia/getdisplaymedia/js/main.js
+++ b/src/content/getusermedia/getdisplaymedia/js/main.js
@@ -7,9 +7,14 @@
  */
 'use strict';
 
-// Polyfill in Firefox.
-// See https://blog.mozilla.org/webrtc/getdisplaymedia-now-available-in-adapter-js/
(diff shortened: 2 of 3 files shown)


eladalon1983:
> @eladalon1983 it seems https://developer.chrome.com/docs/web-platform/screen-sharing-controls/#displaySurface has displaySurface at the wrong level, no?

Yes, that's wrong atm. Thanks for pointing it out! The article will be fixed soon.

beaufortfrancois:
Thanks @fippo for raising this issue! We're fixing the article in https://github.com/GoogleChrome/developer.chrome.com/pull/3881

src/content/getusermedia/getdisplaymedia/index.html
```diff
@@ -35,7 +35,12 @@ <h1><a href="//webrtc.github.io/samples/" title="WebRTC samples homepage">WebRTC
 
     <video id="gum-local" autoplay playsinline muted></video>
     <button id="startButton" disabled>Start</button>
-
+    <select id="preference" style="display:none">
+      <option value="" selected>Share the default</option>
+      <option value="browser">Share a browser window</option>
```
beaufortfrancois:
```suggestion
      <option value="browser">Share a browser tab</option>
```

src/content/getusermedia/getdisplaymedia/index.html
```diff
@@ -35,7 +35,12 @@ <h1><a href="//webrtc.github.io/samples/" title="WebRTC samples homepage">WebRTC
...
+      <option value="window">Share another window</option>
```
beaufortfrancois:
```suggestion
      <option value="window">Share a window</option>
```

src/content/getusermedia/getdisplaymedia/js/main.js
```diff
@@ -7,9 +7,14 @@
  */
 'use strict';
 
-// Polyfill in Firefox.
-// See https://blog.mozilla.org/webrtc/getdisplaymedia-now-available-in-adapter-js/
-if (adapter.browserDetails.browser == 'firefox') {
+const sharingPreference = document.getElementById('preference');
```
beaufortfrancois: How about `preferredDisplaySurface` instead of `sharingPreference` so that it's clear?

src/content/getusermedia/getdisplaymedia/js/main.js
```diff
@@ -7,9 +7,14 @@
...
+
+if (adapter.browserDetails.browser === 'chrome' && adapter.browserDetails.version >= 107) {
+  // See https://developer.chrome.com/docs/web-platform/screen-sharing-controls/
```
beaufortfrancois, line 15: I'd prefer not to promote this pattern. How about simply letting the `select` always visible and add a link to the article or MDN later?

beaufortfrancois (commented):
I've left few comments!

Note that https://developer.chrome.com/docs/web-platform/screen-sharing-controls/#displaySurface has been updated ;)

src/content/getusermedia/getdisplaymedia/index.html
eladalon1983: Why "browser window"? It can be any native window.

eladalon1983: ("Another" is also wrong, btw; one can share the current window too.)

src/content/getusermedia/getdisplaymedia/js/main.js
fippo: this pattern is the sad reality (also here in the samples) unless there is a way to feature detect :-(

beaufortfrancois: 👍

src/content/getusermedia/getdisplaymedia/index.html
beaufortfrancois: Good catch. I've edited my comment to "Share a window"

src/content/getusermedia/getdisplaymedia/js/main.js
fippo: and a better name for the element

src/content/getusermedia/getdisplaymedia/js/main.js
```diff
@@ -7,14 +7,20 @@
  */
 'use strict';
 
-// Polyfill in Firefox.
-// See https://blog.mozilla.org/webrtc/getdisplaymedia-now-available-in-adapter-js/
-if (adapter.browserDetails.browser == 'firefox') {
+const preferredDisplaySurface = document.getElementById('displaySurface');
+
+if (adapter.browserDetails.browser === 'chrome' && adapter.browserDetails.version >= 107) {
```
eladalon1983: nit: Mind breaking up this line?

src/content/getusermedia/getdisplaymedia/js/main.js
```diff
@@ -40,7 +47,12 @@ function errorMsg(msg, error) {
 
 const startButton = document.getElementById('startButton');
 startButton.addEventListener('click', () => {
-  navigator.mediaDevices.getDisplayMedia({video: true, audio: true})
+  const options = {audio: true, video: {}};
+  const displaySurface = preferredDisplaySurface.options[preferredDisplaySurface.selectedIndex];
+  if (displaySurface.value !== '') {
```
eladalon1983: nit: Is this the idiomatic thing to do? What about `if (displaySurface.value) {`?

eladalon1983 (approved):

eladalon1983:
Generally LGTM, but I wonder if adding more than the minimal functionality to an example, might make it less approachable. Please consider if this might deserve its own sample page, possibly along with other similar controls, like in François' page [here](https://screen-sharing-controls.glitch.me/).

src/content/getusermedia/getdisplaymedia/js/main.js
fippo: idiomatic yes, good for a sample no

src/content/getusermedia/getdisplaymedia/js/main.js
beaufortfrancois: +1
```suggestion
if (adapter.browserDetails.browser === 'chrome' && 
    adapter.browserDetails.version >= 107) {
```

beaufortfrancois (approved):
LGTM with nit

https://webrtc.github.io/samples/src/content/getusermedia/getdisplaymedia/ is a canonical page for getDisplayMedia. I think it's fine to have `displaySurface` there.

eladalon1983:
> https://webrtc.github.io/samples/src/content/getusermedia/getdisplaymedia/ is a canonical page for getDisplayMedia. I think it's fine to have `displaySurface` there.

I think it's fine but sub-optimal. I think ideally, this page should have the minimal required functionality.
* Beginners can make sense of it more easily.
* When people are sent there for debugging purposes, they don't end up puzzled over irrelevant controls. (I often use "does the problem manifest on \*this\* page too?")

beaufortfrancois:
How about bundling those advanced options like this?

image


```diff
 
     <video id="gum-local" autoplay playsinline muted></video>
+    <fieldset>
+      <legend>Advanced options</legend>
+      <select id="preference" style="display:none">
+        <option value="" selected>Share the default</option>
+        <option value="browser">Share a browser window</option>
+        <option value="window">Share another window</option>
+        <option value="monitor">Share the entire screen</option>
+      </select>
+    </fieldset>
     <button id="startButton" disabled>Start</button>
-    <select id="preference" style="display:none">
-      <option value="" selected>Share the default</option>
-      <option value="browser">Share a browser window</option>
-      <option value="window">Share another window</option>
-      <option value="monitor">Share the entire screen</option>
-    </select>
```

eladalon1983:
> How about bundling those advanced options like this?
> ...

Yeah, that makes sense to me. Especially if the example can still demonstrate that `video: true` is an option. (I think a lot of developers end up not needing constraints. But I've not checked this hypothesis.)

beaufortfrancois:
@fippo How does that sound? If that's too much, I can take this PR over. Let me know.

src/content/getusermedia/getdisplaymedia/js/main.js
fippo: changed to `!= 'default'` which is easier to understand

fippo:
I like the fieldset! Borrowed with updated labels. Also changed the labels to "prefer to share"

src/content/getusermedia/getdisplaymedia/index.html
```diff
@@ -35,7 +35,15 @@ <h1><a href="//webrtc.github.io/samples/" title="WebRTC samples homepage">WebRTC
 
     <video id="gum-local" autoplay playsinline muted></video>
     <button id="startButton" disabled>Start</button>
-
+    <fieldset id="options" style="display:none">
```
beaufortfrancois, line 38: You may want to update CSS with something like:

```diff
+fieldset {
+  margin: 0 0 1em 0;
+}
+
+fieldset > select {
+  margin-top: 1em;
+}
+
```

src/content/getusermedia/getdisplaymedia/index.html
```diff
@@ -35,7 +35,15 @@ <h1><a href="//webrtc.github.io/samples/" title="WebRTC samples homepage">WebRTC
...
+      <legend>Advanced options</legend>
+      <select id="displaySurface">
+        <option value="default" selected>Show default sharing options</option>
+        <option value="browser">Prefer to share a browser tab</option>
+        <option value="window">Prefer to share a window</option>
+        <option value="monitor">Prefer to share the entire screen</option>
```
beaufortfrancois:
```suggestion
        <option value="monitor">Prefer to share an entire screen</option>
```
There can be multiple screens plugged.

src/content/getusermedia/getdisplaymedia/js/main.js
```diff
@@ -40,7 +48,12 @@ function errorMsg(msg, error) {
 
 const startButton = document.getElementById('startButton');
 startButton.addEventListener('click', () => {
-  navigator.mediaDevices.getDisplayMedia({video: true, audio: true})
+  const options = {audio: true, video: {}};
```
beaufortfrancois: As we may be updating `options` after, we should use `let`.
```suggestion
  let options = {audio: true, video: true};
```

src/content/getusermedia/getdisplaymedia/js/main.js
```diff
@@ -40,7 +48,12 @@ function errorMsg(msg, error) {
...
+  const displaySurface = preferredDisplaySurface.options[preferredDisplaySurface.selectedIndex];
+  if (displaySurface.value !== 'default') {
+    options.video.displaySurface = displaySurface.value;
+  }
```
beaufortfrancois:
```suggestion
  const displaySurface = preferredDisplaySurface.options[preferredDisplaySurface.selectedIndex].value;
  if (displaySurface !== 'default') {
    options.video = { displaySurface };
  }
```

src/content/getusermedia/getdisplaymedia/index.html
fippo: added to main css, tagging this as TBR=@samdutton ;-)

src/content/getusermedia/getdisplaymedia/js/main.js
fippo: JS semantics: you can change properties of a const map 🤷‍♂️

eladalon1983: > JS semantics: you can change properties of a const map 🤷‍♂️

I usually think of it as the difference between a const-ptr and a ptr-to-const, then it makes sense.

src/content/getusermedia/getdisplaymedia/js/main.js
```diff
@@ -38,9 +47,14 @@ function errorMsg(msg, error) {
   }
 }
 
-const startButton = document.getElementById('startButton');
```
beaufortfrancois, line 41: Why did you remove this line?

beaufortfrancois (approved):

beaufortfrancois:
LGTM with nit

@fippo I'll merge after `const startButton = document.getElementById('startButton');` line is restored if that looks good to you.

image

src/content/getusermedia/getdisplaymedia/js/main.js
fippo: not removed, moved up. We do declare the the UI-related accessors at the top

beaufortfrancois:
And it's live at https://webrtc.github.io/samples/src/content/getusermedia/getdisplaymedia/! Thank you @fippo!

image

PR #1576, opened 2022-10-11
add Chrome/Firefox interop tests

fippo:
**Description**

adds interoperability tests between Chrome unstable and Firefox Nightly. These tests are designed to run nightly as github actions. Github provides a "free" budget of 2000 minutes per month. In the current configuration with four tests (a matrix of Chrome/Firefox) that each take around a minute the nightly runs have a "cost" of 120 minutes per month.

The tests do not rely on a signaling server, instead use the mocha-based test that is controlling the individual selenium webdriver instances to act as a signaling channel which exchanges offers and answers with candidates.
This pattern is described in the testbed repository
  https://github.com/fippo/testbed
from 2016, this is a "more modern" take on the same subject.

To run the tests locally,
* npm run mocha test/interop/connection

Running them with Chrome/Firefox unstable requires Linux for the integration with travis-multirunner and setting the
* BROWSER_A
* BROWSER_B
* BVER environment variables.

**Purpose**
Getting rid of the [dependency on apprtc in tests](https://bugs.chromium.org/p/webrtc/issues/detail?id=14376) while keeping a minimum level of interoperability tests

### DIFF
--- a/.github/workflows/interop-tests.yml
+++ b/.github/workflows/interop-tests.yml
@@ -1,0 +1,22 @@
+on:
+  schedule:
+    - cron: "30 5 * * *"
+  push:
+jobs:
+  interop:
+    runs-on: ubuntu-latest
+    timeout-minutes: 5
+    strategy:
+      fail-fast: false
+      matrix:
+        browserA: [chrome, firefox]
+        browserB: [firefox, chrome]
+        bver: ['unstable']
+    steps:
+    - uses: actions/checkout@v2
+    - uses: actions/setup-node@v3
+    - run: npm install
+    - run: BROWSER=${{matrix.browserA}} BVER=${{matrix.bver}} ./node_modules/travis-multirunner/setup.sh
+    - run: BROWSER=${{matrix.browserB}} BVER=${{matrix.bver}} ./node_modules/travis-multirunner/setup.sh
+    - run: Xvfb :99 &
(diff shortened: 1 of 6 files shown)


fippo:
I've run those tests for ~2 weeks now and they were surprisingly reliable (unlike the samples tests...)
Adding Safari to the mix is possible (but that has a [minute multiplier of 10](https://docs.github.com/en/billing/managing-billing-for-github-actions/about-billing-for-github-actions#minute-multipliers)) too but I'd rather run the basic case that we want to replace apprtc first for a month or two before getting more ambitious.

One of the tricky things is that chromedriver normally has a version check which allow it to operate +- 1 release of the chrome version. This broke when Chrome Unstable went to M108 and required adding the `disable-build-check` flag.
It might actually be useful to extend the chromedriver flag to specify a range of 2 as an argument instead of completly disabling the check

KaptenJansson (approved):
LGTM overall but with a disclaimer: it's been a while since I worked on this but I trust you that it works.

NIT: Please add copyright headers on each file.

alvestrand (approved):
Where do test failures show up?

fippo:
They will show up on https://github.com/webrtc/samples/actions which can be filtered.
I plan to add a "send email" functionality once running this for a month shows it is solid enough that a failure justifies nagging people ;-)

PR #1577, opened 2022-10-11
test: add cli option to retry three times to mocha

fippo:
to increase reliability

https://mochajs.org/#-retries-n


fippo:
ofc I read the docs as `--retries 3` but as with any `--` argument they are `key=value` ;-)

KaptenJansson (approved):
LGTM

PR #1581, opened 2022-10-28
trickle-ice: trim information about candidates

fippo:
since some of the current information is not relevant for this use-case and revert the width of the table which was enlarged in #1561 since this is no longer necessary.


KaptenJansson (approved):
LGTM

PR #1582, opened 2022-10-30
trickle-ice: only acquire audio stream

fippo:
which avoids adding audio and video and gathering candidates for multiple components


KaptenJansson (approved):
LGTM

PR #1584, opened 2022-11-21
Update video-crop demo note

ggarber:
Not requiring an experimental flag anymore

https://developer.mozilla.org/en-US/docs/Web/API/MediaStreamTrackProcessor#browser_compatibility


fippo (approved):
🍰

PR #1588, opened 2023-01-22
Volume with wasm

roy000731:
**Description**


alvestrand:
I'd completely forgotten that I had written this ... does it still work?
Do you think it's worth putting into the repository?

PR #1593, opened 2023-02-27
add video_replay sample

fippo:
which allows replaying IVF files generated by video_replay in the browser. This is useful for debugging issues like
  https://bugs.chromium.org/p/chromium/issues/detail?id=1418596
where the native video_replay is using software decoder which do not show the issue.


src/content/insertable-streams/video-replay/js/main.js
```diff
@@ -0,0 +1,121 @@
...
+  }
+
+  const decoder = new VideoDecoder({
+    output: async (frame) => {
+      await writer.write(frame);
+      frame.close();
+      const nextFrame = await ivf.readFrame();
+      if (nextFrame) {
+        decoder.decode(new EncodedVideoChunk({
+          timestamp: Number(nextFrame.timestamp - firstFrame.timestamp) * 1000,
```
fippo, line 101: I am not sure this is correct or it takes the headers timescale into account. But I don't have a great sample file to check at hand.

fippo:
Test failure is due to an unrelated E2E test failing

PR #1600, opened 2023-03-22
use https:// for external pages instead of //

fippo:
since `//` times out on `file://` contexts used for testing.


fippo:
This should take care of the test timeouts... I wonder if we really need Google analytics, did anyone care about this in the last five years?

alvestrand (approved):
Good question about the analytics.
Would actually be interesting to see what they are, but I don't know if anyone currently working on the project knows how to find them. I'll ask around.

PR #1601, opened 2023-03-22
Fix Chrome(>=109) removed event.path

simo-an:
**Description**
When i switch audio output, the error occured.
Because of Chrome(>=109) removed event.path, so change the method of get audio element

more see:
https://bugs.chromium.org/p/chromium/issues/detail?id=1277431


fippo:
Superseeded by #1638

PR #1603, opened 2023-04-03
Update bandwidth sample

fippo:
modernizing it a bit and allowing easier testing of the lowest limit (yep, it works at 5kbps even)

### DIFF
--- a/src/content/peerconnection/bandwidth/js/main.js
+++ b/src/content/peerconnection/bandwidth/js/main.js
@@ -200,8 +200,14 @@ function onSetSessionDescriptionError(error) {
 bandwidthSelector.onchange = () => {
   bandwidthSelector.disabled = true;
   const bandwidth = bandwidthSelector.options[bandwidthSelector.selectedIndex].value;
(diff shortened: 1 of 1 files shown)


src/content/peerconnection/bandwidth/js/main.js
```diff
@@ -215,36 +221,34 @@ bandwidthSelector.onchange = () => {
...
       parameters.encodings = [{}];
     }
-    if (bandwidth === 'unlimited') {
+    if (bandwidthInKbps === 'unlimited') {
       delete parameters.encodings[0].maxBitrate;
     } else {
-      parameters.encodings[0].maxBitrate = bandwidth * 1000;
+      parameters.encodings[0].maxBitrate = bandwidthInKbps * 1000;
     }
+    return sender.setParameters(parameters);
```
beaufortfrancois, line 229: Code below won't be executed with this added `return`

fippo: Away with it, thanks for 👀

henbos (approved):

PR #1605, opened 2023-04-09
Update volume example to use web audio API

### LINKED ISSUE #1509: Code sample using depreciated createScriptProcessor function
- [x] I have provided steps to reproduce
- [x] I have provided browser name and version
- [x] I have provided a link to the sample here or a modified version thereof

## Browser affected

Internet Explorer (all versions)

## Description

The [Audio Stream Volume](https://github.com/webrtc/samples/blob/gh-pages/src/content/getusermedia/volume/js/soundmeter.js) sample uses the depreciated `BaseAudioContext.createScriptProcessor()` instead of `AudioWorkletNode`

## Steps to reproduce

NA

## Expected results

NA

## Actual results

NA

porjo:
**Description**

Update volume example to use web audio API. Closes #1509 

NOTE: soundmeter.js is no longer loaded via index.html but instead via Javascript using [Worklet `addModule()`](https://developer.mozilla.org/en-US/docs/Web/API/Worklet/addModule) .  As such, [CORS](https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS) comes into effect and files must be served from a HTTP file server (not referenced with `file:///`)

Credit to this Stackoverflow post which was used as a reference: https://stackoverflow.com/a/62732195/202311

### DIFF
--- a/src/content/getusermedia/volume/index.html
+++ b/src/content/getusermedia/volume/index.html
@@ -65,7 +65,7 @@
         about a second.</p>
     <p>Note that you will not hear your own voice; use the <a href="../audio">local audio rendering demo</a> for that.
     </p>
-    <p>The <code>audioContext</code>, <code>stream</code> and <code>soundMeter</code> variables are in global scope, so
+    <p>The <code>audioContext</code> and <code>stream</code> variables are in global scope, so
         you can inspect them from the console.</p>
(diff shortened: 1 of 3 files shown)


porjo:
Obsoleted by #1723

PR #1610, opened 2023-07-13
record: Correct AV1 MIME type

rosetta-jpn:
The correct AV1 MIME is "av01".

### DIFF
--- a/src/content/getusermedia/record/js/main.js
+++ b/src/content/getusermedia/record/js/main.js
@@ -69,7 +69,7 @@ function handleDataAvailable(event) {
 
 function getSupportedMimeTypes() {
   const possibleTypes = [
-    'video/webm;codecs=av1,opus',
+    'video/webm;codecs=av01,opus',
     'video/webm;codecs=vp9,opus',
     'video/webm;codecs=vp8,opus',
     'video/webm;codecs=h264,opus',


alvestrand:
The registered media type for AV1 at IANA is AV1: https://www.iana.org/assignments/media-types/video/AV1

Where did you get a reference saying "AV01" from?

rosetta-jpn:
The media type should be used for real time streaming.

> Applications which use this media:
> Video over IP, video conferencing, video streaming.

I think, for Media Streaming API, the regular MIME type should be used.
https://aomediacodec.github.io/av1-isobmff/#av1sampleentry-section

rosetta-jpn:
I searched more. 
So there are two ways how to specify the mime, mime and mime+codecs. 
https://developer.mozilla.org/en-US/docs/Web/API/MediaRecorder/mimeType

(1) With MIME only, `video/av1` should be used.

See https://developer.mozilla.org/en-US/docs/Web/API/MediaRecorder/mimeType
> The media type strings are standardized by the Internet Assigned Numbers Authority (IANA). For their official list of defined media type strings, see the article [Media Types](https://www.iana.org/assignments/media-types/media-types.xhtml) on the IANA site.

(2) With MIME + codecs, 'video/webm; codecs="av01"' should be used.

See https://developer.mozilla.org/en-US/docs/Web/Media/Formats/codecs_parameter#av1


I will clarify the commit message.

rosetta-jpn:
I created a new pull request https://github.com/webrtc/samples/pull/1611.

Issue #1614, opened 2023-08-02
E2E interop test failures


fippo:
The nightly interop tests started failing with
  https://github.com/webrtc/samples/actions/runs/5723012987

This looks like a mix of
* chromedriver refusing to recognize the symlink in browsers/bin
* Chromedriver 115 being incompatible with M117 after a certain version.

The path is easy to fix (tested [here](https://github.com/fippo/interop/actions/runs/5737241319)) but after that it is still running into timeout issues

fippo:
The difference between the last known good run
https://github.com/webrtc/samples/actions/runs/5711022621/job/15472020011
and the first bad one 
https://github.com/webrtc/samples/actions/runs/5723012987/job/15509825644#step:8:31
is that the latter shows
```
Selenium Manager binary found at 
    /home/runner/work/samples/samples/node_modules/selenium-webdriver/bin/linux/selenium-manager
```

fippo:
@bonigarcia you *might* know since I saw your twe..whatever one calls it now :-)

It seems that upgrading from 4.10.0 to 4.11.0 broke the custom path we use (together with the very old travis-multirunner) via `setChromeBinaryPath`. Does that ring a bell?

bonigarcia:
It might be a bug on Selenium. Can you please open a bug report [here](https://github.com/SeleniumHQ/selenium)?

I am reporting this in the Selenium TLC Slack channel as well.

fippo:
Already fixed, just needs to be released before unpinning this. Thank you @bonigarcia!

PR #1617, opened 2023-08-23
pin selenium-webdriver to 4.10.0

fippo:
which avoids issues in the integration of travis-multirunner since the custom Chrome path is no longer found when being used with selenium manager

See #1614


fippo:
We're back to general test flakyness \o/

KaptenJansson (approved):
LGTM

PR #1623, opened 2023-09-18
trickle-ice: remove custom url check

fippo:
and let the browser throw an error which shows up in the field at the bottom


KaptenJansson (approved):
LGTM

PR #1624, opened 2023-09-19
test: port to jest

fippo:
to simplify the testing stack.
Also remove unused babelrc

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,8 +1,0 @@
-{
-  "presets": [
-    "env"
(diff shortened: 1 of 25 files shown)


fippo:
Slightly concerned that jest doesn't support --retries (yet) but limiting concurrency might get rid of the datachannel flakes we see sometimes.

alvestrand (approved):
Anything that makes more people understand how to do tests is good.
I hope.

PR #1625, opened 2023-09-20
Fix jest interop tests

fippo:
follow-up to #1624 that fixes the nightly interop pipeline


alvestrand (approved):

fippo:
The complaint about using M116 with a M117 chromedriver is surprising since travis-multirunner does install M117... maybe more setBinaryPath woes :-(

fippo:
Turns out that bver vs version was a thing, thanks for noticing that @alvestrand!

PR #1629, opened 2023-10-09
fix: add async for createPeerConnection when use await call it

grootgordon:
**Description**
When use VSCode edit main.js, if we call ```await createPeerConnection();```

there is a warning of: `'await' has no effect on the type of this expression.ts(80007`
**Purpose**
resolve warning


grootgordon:
sign CLA done.

PR #1630, opened 2023-10-13
Remove wrong AV1 codecs string

rosetta-jpn:


alvestrand (approved):
I've verified that the av01 string works in current Chrome.

PR #1631, opened 2023-10-16
Remove chromedriver dependency

fippo:
since this gets automagically installed by Selenium these days

### DIFF
--- a/.github/workflows/interop-tests.yml
+++ b/.github/workflows/interop-tests.yml
@@ -1,6 +1,7 @@
 on:
   schedule:
     - cron: "30 5 * * *"
+
 jobs:
   interop:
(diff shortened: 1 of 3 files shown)


alvestrand (approved):

alvestrand:
Still chromedriver woes....

fippo:
Without the chromedriver dependency selenium found the [preinstalled](https://github.com/actions/runner-images/blob/main/images/linux/Ubuntu2204-Readme.md#browsers-and-drivers) chromedriver.
Forcefully removing that from the path as part of the github action made selenium download a matching version.

fippo:
I wonder about geckodriver but... separate PR and it hasn't shown any issues.

PR #1632, opened 2023-10-18
test: remove geckodriver

fippo:
follow-up to #1631


fippo:
https://github.com/webrtc/samples/actions/runs/6558809718/job/17813136593?pr=1632 shows that the nightly workflow works with this change.

KaptenJansson (approved):

PR #1640, opened 2023-11-20
change-codecs: move codec selection to the receiver

fippo:
While using setCodecPreferences on the sending side works because the answerer will match the sender's order, using it on the receiving end is more semantically correct.

See this comment
https://github.com/w3c/webrtc-encoded-transform/pull/186#issuecomment-1818658226

PR #1642, opened 2023-11-25
move codec selection to the receiver for more samples

fippo:
follow-up to #1640 which got copy-pasted quite a bit.

### DIFF
--- a/src/content/insertable-streams/endtoend-encryption/js/videopipe.js
+++ b/src/content/insertable-streams/endtoend-encryption/js/videopipe.js
@@ -10,14 +10,14 @@
 //
 // The usage of this abstraction:
 // var pipe = new VideoPipe(mediastream, handlerFunction);
-// handlerFunction = function(mediastream) {
+// handlerFunction = function(MediaStreamTrackEvent) {
 //   do_something
 // }
 // pipe.close();
 //
 // The VideoPipe will set up 2 PeerConnections, connect them to each
-// other, and call HandlerFunction when the stream is available in the
-// second PeerConnection.
+// other, and call HandlerFunction when the stream's track is available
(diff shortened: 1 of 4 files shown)


src/content/insertable-streams/endtoend-encryption/js/videopipe.js
```diff
@@ -36,36 +36,37 @@ function VideoPipe(stream, forceSend, forceReceive, handler) {
...
-    const selectedCodecIndex = codecs.findIndex(c => c.mimeType === preferredVideoCodecMimeType);
-    const selectedCodec = codecs[selectedCodecIndex];
-    codecs.splice(selectedCodecIndex, 1);
-    codecs.unshift(selectedCodec);
-    const transceiver = this.pc1.getTransceivers()
-        .find(t => t.sender && t.sender.track === stream.getVideoTracks()[0]);
-    transceiver.setCodecPreferences(codecs);
+  if ('setCodecPreferences' in window.RTCRtpTransceiver.prototype) {
+    this.pc2.ontrack = (e) => {
+      if (e.track.kind === 'audio' && preferredAudioCodecMimeType ) {
```
alvestrand, line 41: Hm. From surrounding text, it looks as if this should be a handler function, not an innate property of VideoPipe.
But that's a different refactoring from this one, so not a showstopper.

alvestrand (approved):
Looks straightforward. Only comment is on a broken abstraction that was already present in the previoius version.

src/content/insertable-streams/endtoend-encryption/js/videopipe.js
fippo: The good news is that we can now actually move it to the outside ontrack handler

PR #1643, opened 2023-11-28
fix(change-codecs): audio receiver set video codec throw error

changzhn:
**Description**
Line 183 e.transceiver.setCodecPreferences(codecs); e.transceiver may be for audio, set video will throw error.
error message: Failed to execute 'setCodecPreferences' on 'RTCRtpTransceiver': Invalid codec preferences: Missing codec from recv codec capabilities.

**Purpose**
fix bug


fippo:
Thank you for catching that! The test did not catch it because ontrack gets called both for audio and video and this works for the video part which ends up setting the srcObject and show the video.

fippo (approved):

PR #1645, opened 2023-12-01
Correct name of video element

eladalon1983:


fippo (approved):
while unused, it should not be called gum-local

PR #1653, opened 2024-04-17
trickle-ice: spelling correction in error note

renandincer:
Just correcting a small spelling error I happened to notice.


fippo (approved):
Thanks for catching that!

PR #1654, opened 2024-04-20
test: replace travis-multirunner with @puppeter/browsers

fippo:
following what was done in adapter in https://github.com/webrtcHacks/adapter/pull/1155 which is now back to supporting Firefox stable


fippo:
This one turned out to be mean... showing that adapter did not test in the downloaded chrome because CHROME_BIN was already set in the environment.

KaptenJansson (approved):

PR #1655, opened 2024-04-23
Remove old-new-stats

fippo:
since old "legacy" stats are no more


KaptenJansson (approved):
Thank you for the delete delete :)

fippo:
Tagging @henbos for papertrail

PR #1657, opened 2024-04-29
refactor: small simplification

Keuklar:
**Description**
small refactoring

**Purpose**
code factorisation / simplification / concision

### DIFF
--- a/src/content/datachannel/basic/js/main.js
+++ b/src/content/datachannel/basic/js/main.js
@@ -150,15 +150,12 @@ function onReceiveMessageCallback(event) {
 function onSendChannelStateChange() {
   const readyState = sendChannel.readyState;
   console.log('Send channel state is: ' + readyState);
-  if (readyState === 'open') {
-    dataChannelSend.disabled = false;
+  const closed = readyState !== 'open'
+  dataChannelSend.disabled = closed;
+  sendButton.disabled = closed;
+  closeButton.disabled = closed;
+  if (!closed) {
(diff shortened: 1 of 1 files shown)


alvestrand:
The purpose of samples is to have code that is easy to understand.
While the suggestion uses many more Javascript constructs, I have trouble seeing how it is easier to understand than the old code.

Keuklar:
ok I see, sorry for that, the main goal was to factor this code :
```javascript
if (readyState === 'open') {
    dataChannelSend.disabled = false;
    dataChannelSend.focus();
    sendButton.disabled = false;
    closeButton.disabled = false;
  } else {
    dataChannelSend.disabled = true;
    sendButton.disabled = true;
    closeButton.disabled = true;
  }
```
  
  which is a bit verbose, repetitive.
But I got a little carried away, I could just focus on the 'if...else' if you wish.
Something like that: 

```javascript
  const closed = readyState !== 'open'
  dataChannelSend.disabled = closed;
  sendButton.disabled = closed;
  closeButton.disabled = closed;
  if (!closed) {
    dataChannelSend.focus();
  }

```

alvestrand:
That seems more readable, yes. (I would call the variable "isOpen" just because that's what the ready state is named. But that's taste.

Keuklar:
So, I've made a new proposal, I let `closed` const, as you said it's a matter of taste, since I find it more straightforward.

stepancar (approved):

PR #1664, opened 2024-05-26
Fix error message handling in getusermedia/gum.

bc-lee:
**Description**

Currently, when some errors occur in the getusermedia/gum demo, the error message isn't handled correctly because the constraints object might not have required properties in some cases. This patch fixes the error message handling by using the error object directly and dumping it as a JSON string.

**Purpose**

### DIFF
--- a/src/content/getusermedia/gum/js/main.js
+++ b/src/content/getusermedia/gum/js/main.js
@@ -24,10 +24,9 @@ function handleSuccess(stream) {
 
 function handleError(error) {
   if (error.name === 'OverconstrainedError') {
-    const v = constraints.video;
-    errorMsg(`The resolution ${v.width.exact}x${v.height.exact} px is not supported by your device.`);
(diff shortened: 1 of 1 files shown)


fippo:
I assume this only happens if one manually sets the video constraints to cause an overconstrained error.

Can you revert to using errorMsg *and* (for consistency) prefix with the error name in addition to the somewhat helpful error message?

bc-lee:
> I assume this only happens if one manually sets the video constraints to cause an overconstrained error.

I encountered this issue without modifying the constraints. I was testing Linux WebKitGTK to see if media capture streams (and WebRTC) were working. Its implementation seems to be buggy, as I got an OverconstrainedError when trying to get the video stream, even though I didn't set any constraints.

fippo (approved):

fippo:
Thank you! Maybe @philn knows more about what WebKitGTK might be doing in wrong ways, it did show this code was wrong which is 👍

philn:
That demo works just fine in my development build.

PR #1665, opened 2024-06-06
fix: Add workaround for Firefox DTMF compatibility

bc-lee:
**Description**

`src/content/peerconnection/dtmf/js/main.js` was incompatible with Firefox due to its reliance on the `canInsertDTMF` function, which is not supported in Firefox. This commit implements a workaround by assuming DTMF insertion is always available when `canInsertDTMF` is not present.

**Purpose**

### DIFF
--- a/src/content/peerconnection/dtmf/js/main.js
+++ b/src/content/peerconnection/dtmf/js/main.js
@@ -173,7 +173,9 @@ function dtmfOnToneChange(tone) {
 }
 
 function sendTones(tones) {
-  if (dtmfSender && dtmfSender.canInsertDTMF) {
+  // firefox doesn't implement canInsertDTMF, so assume it's always available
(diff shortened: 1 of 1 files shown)


fippo:
looks like you have an upstream fix which is the preferred way? Now we just need @Pehrsons to agree but that should be easy ;-)

bc-lee:
Yes, I've uploaded my patch to the Mozilla Phabricator to implement `canInsertDTMF` method. However, I still believe using the workaround for older Firefox versions is worthwhile. Even web platform tests utilizes the workaround, as you can see [here](https://github.com/web-platform-tests/wpt/blob/2bd26e124cf17b2f0a25c150794d640b07b2a870/webrtc/RTCDTMFSender-helper.js#L31).

fippo (approved):
we have done more terrible things in the past :-)

PR #1668, opened 2024-07-04
multiple: modernize code

fippo:
which was still stuck in the 2012 style

### DIFF
--- a/src/content/peerconnection/multiple/js/main.js
+++ b/src/content/peerconnection/multiple/js/main.js
@@ -29,10 +29,6 @@ let pc1Local;
 let pc1Remote;
 let pc2Local;
 let pc2Remote;
-const offerOptions = {
-  offerToReceiveAudio: 1,
-  offerToReceiveVideo: 1
-};
 
 const supportsSetCodecPreferences = window.RTCRtpTransceiver &&
   'setCodecPreferences' in window.RTCRtpTransceiver.prototype;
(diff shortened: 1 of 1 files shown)


src/content/peerconnection/multiple/js/main.js
```diff
@@ -80,67 +68,33 @@ function call() {
...
-
-  localStream.getTracks().forEach(track => pc2Local.addTrack(track, localStream));
-  console.log('Adding local stream to pc2Local');
-  pc2Local.createOffer(offerOptions)
-      .then(gotDescription2Local, onCreateSessionDescriptionError);
-}
-
-function onCreateSessionDescriptionError(error) {
-  console.log(`Failed to create session description: ${error.toString()}`);
+  localStream.getTracks().forEach(track => {
```
nikolov9996, line 80: looks fine to me, but there is no changes in tests, do the tests work the same way after replacing callbacks with async/await?

nikolov9996 (commented):
No changes in the tests after async/await changes.
Please leave a comment or tests - if required.
I see the check is passing, but does the current test working as expected for the new style.

henbos (approved):

src/content/peerconnection/multiple/js/main.js
henbos: This PR is just refactoring from old style to new style so if the old code was well tested then it will have ensured that this PR did not regress any behavior. Not sure how well tested webrtc samples is in general but what this PR is doing is LGTM

PR #1679, opened 2024-09-09
don't use exact property for device ID

yhc002:
Use ideal property instead of exact property when calling getUserMedia API. 
**Description**
Since enumerateDevices API does not request device permission, the test page will not receive any device ID if camera permission is not already set nor will it trigger a device permission popup. Since getUserMedia API will be called with empty device ID using 'exact' property, test page will not display any video.

**Purpose**
Allow the test page to get camera video on initial launch or after camera permission for test page has been reset.


fippo:
cc @handellm

henbos (approved):

PR #1683, opened 2024-10-17
Add HEVC to MediaRecorder mimetypes list

StaZhu:
HEVC MediaRecorder support will be available after: https://chromium-review.googlesource.com/c/chromium/src/+/4707115

It will support `video/x-matroska` and `video/mp4` mimetypes.

Note that the latest Chromium also support VP9/AV1 `video/mp4` encoding, so add them to the list as well.


handellm (approved):

henbos (approved):
LGTM on behalf of handellm

PR #1687, opened 2024-11-23
test: use Chrome Canary on Linux

fippo:
which is now available:
  https://groups.google.com/a/chromium.org/g/chromium-dev/c/pAPo4zM12m8/m/P77lbkh0AAAJ?utm_medium=email&utm_source=footer

### DIFF
--- a/test/webdriver.js
+++ b/test/webdriver.js
@@ -38,7 +38,7 @@ if (os.platform() === 'win32') {
 function mapVersion(browser, version) {
   const versionMap = {
     chrome: {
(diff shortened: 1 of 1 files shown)


fippo:
https://github.com/webrtc/samples/actions/runs/11989364046 ran the PR tests with canary by using `unstable` [ħere](https://github.com/webrtc/samples/blob/gh-pages/.github/workflows/test.yml#L21) so this should work for interop tests too

alvestrand (approved):

PR #1688, opened 2024-12-04
Use valid mimetype for HEVC/H.264/AV1 and add `hev1`/`avc3` mimetypes

StaZhu:
Based on conversation with Chrome Media Team, we should only support valid mimetype for future new codecs like HEVC.

For HEVC/H.264/AV1, it is recommend to use valid mimetype instead of just `hvc1`, `avc1` or `av01`.

There are some limitation for `avc1` / `hvc1` as they only allow "out of band parameter sets", so if capture device can produce dynamic resolution video stream, only `avc3` / `hev1` can satisfy the needs as they supports "in band parameter sets".

Chromium may support `hev1` and `avc3` in the future, so also added these mimetypes for future testing.


StaZhu:
cc @handellm

fippo (approved):
LGTM!

PR #1689, opened 2024-12-11
Add ondevicechange event support to demo

markafoltz:
Demo does not use `devicechange` event to keep device selectors in sync with hotplug.  This fixes the demo to do that.

It also fixes some other issues:
- The demo now works if only a camera or mic is available (but not both)
- The demo waits for enumeration to complete before calling gUM.  Previously, this would result in a failed request.

`console.log` will be removed before merging.

### DIFF
--- a/src/content/devices/input-output/js/main.js
+++ b/src/content/devices/input-output/js/main.js
@@ -48,7 +48,9 @@ function gotDevices(deviceInfos) {
   });
 }
 
-navigator.mediaDevices.enumerateDevices().then(gotDevices).catch(handleError);
+function getDevices() {
+  navigator.mediaDevices.enumerateDevices().then(gotDevices).catch(handleError);
+}
 
 // Attach audio output device to video element using device/sink ID.
 function attachSinkId(element, sinkId) {
@@ -79,8 +81,7 @@ function changeAudioDestination() {
 function gotStream(stream) {
   window.stream = stream; // make stream available to console
   videoElement.srcObject = stream;
-  // Refresh button list in case labels have become available
-  return navigator.mediaDevices.enumerateDevices();
+  getDevices();
 }
 
 function handleError(error) {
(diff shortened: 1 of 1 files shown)


markafoltz:
This should be ready for review.  Who can take a look? @handellm @henbos @fippo

fippo:
Nice! `eslint --fix` should be able to deal with the lint failures

fippo:
(maybe reduce the console.log amount but... I like console.log)

handellm:
I tried the patched demo, with repository checked out and `python3 -m http.server` serving the files and navigating to localhost. Things started out not working, probably there was something strange about permissions. After accepting cam & mike permissions things jumped to life and worked well.

+1 on fixing lint failures otherwise lgtm.

guidou (approved):
lgtm % the automated comments (var -> let/const and spaces)

markafoltz:
> eslint --fix should be able to deal with the lint failures

This seems broken for me:

```
$ eslint --fix main.js

Oops! Something went wrong! :(

ESLint: 6.4.0.

ESLint couldn't find the config "google" to extend from. Please check that the name of the config is correct.

The config "google" was referenced from the config file in "/home/mfoltz/github/markafoltz/webrtc-samples/.eslintrc.js".
```

I will fix manually instead.

markafoltz:
> Things started out not working, probably there was something strange about permissions.

Yeah.  If you call `enumerateDevices()` without permissions it returns dummy values with empty `deviceIds`, but the demo would populate the dropdowns anyway.

I updated it to avoid populating the dropdowns with dummy values, so permission is requested right away and then the dropdowns can get the true values.

markafoltz:
Okay, this should be ready for another look and a merge if there are no more comments.

fippo (approved):

fippo:
(something weird with webdriver these days, needs investigation)

PR #1691, opened 2024-12-25
Don's ajust audio track if there is no audio track

StaZhu:
On a M4 Mac Mini, there is no built-in microphone or camera, so `getAudioTracks()` will always return an empty array. In this case, we don't need to ajust sampling rate to 48khz...

Also change av1 codecs string from `av01.1.x.y` to `av01.0.x.y` since MediaRecorder only support av1 main profile instead of high profile.

### DIFF
--- a/src/content/getusermedia/record/js/main.js
+++ b/src/content/getusermedia/record/js/main.js
@@ -93,7 +93,7 @@ function getSupportedMimeTypes() {
     'video/mp4;codecs=avc3.64003E,mp4a.40.2',
     'video/mp4;codecs=hvc1.1.6.L186.B0,mp4a.40.2',
     'video/mp4;codecs=hev1.1.6.L186.B0,mp4a.40.2',
-    'video/mp4;codecs=av01.1.19M.08,mp4a.40.2',
+    'video/mp4;codecs=av01.0.19M.08,mp4a.40.2',
     'video/mp4',
   ];
   return possibleTypes.filter(mimeType => {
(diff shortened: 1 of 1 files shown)


alvestrand:
Looks straightforward, but makes the relevant selenium test fail (see bot logs). Please investigate, and fix if possible.

StaZhu:
```

> webrtc-samples@1.0.0 jest
> node test/download-browsers.js && jest --testTimeout 5000 --maxWorkers=1 src/content/ --retries=3

/home/runner/work/samples/samples/node_modules/selenium-webdriver/lib/error.js:521
    let err = new ctor(data.message)
              ^

SessionNotCreatedError: session not created: Chrome failed to start: exited normally.
  (session not created: DevToolsActivePort file doesn't exist)
  (The process started from chrome location /home/runner/work/samples/samples/browsers/chrome/linux-131.0.6778.204/chrome-linux64/chrome is no longer running, so ChromeDriver is assuming that Chrome has crashed.)
```

It seems the failure not related to the change, could you help re-trigger the CI? Thanks

StaZhu:
Hi @alvestrand, is it possible to skip the CI, I run the CI locally and don't have any issue.

StaZhu:
@handellm Could you help take a look markus? Seems like you previous merge request also has similar CI issue.

alvestrand:
@fippo found an error that might have been the reason why this workflow failed. Will approve once the test passes.

fippo:
Still picking up the wrong CI image it seems ... lets try again tomorrow and if it still does not merge ignore the failure

StaZhu:
Thanks!

alvestrand (approved):

alvestrand:
Merging, as the latest fix to the CI config seems to have fixed the issue.
Thanks for your patience!

PR #1694, opened 2025-01-03
Pin ubuntu version to 22.04 LTS

fippo:
fixes #1692 (well, works around)

PR #1695, opened 2025-01-04
pin browser tests to ubuntu 22.04 LTS

fippo:
follow-up to #1694 which only replaced the wrong (first) test runner.
Use ubuntu-latest for lint


fippo:
There was more :-)

alvestrand (approved):

PR #1715, opened 2025-08-12
munge-sdp: remove audio/video device selectors

fippo:
which are unrelated to the munge sample

(and show warnings for audiooutput devices in enumerateDevices, suggesting this hasn't been updated for ... the better part of a decade)


fippo:
`sourceId` ... I have not heard that constraint name in a looooong time!

alvestrand:
Given that the code still uses {optional: []}, it's been this way for a long time.
And it doesn't seem to allow dropping the audio or video anyway.

Let it go.

alvestrand (approved):

PR #1717, opened 2025-09-09
Resolution page: Change to exact deviceId

### LINKED ISSUE #1716: "Resolution" page uses ideal instead of exact for deviceId
It's confusing that you can easily get the wrong camera depending on which resolution you ask for. I think if you have explicitly said in the drop-down menu that you want camera X, you should not get camera Y. So we should use exact instead of ideal.

henbos:
Fixes #1716


alvestrand (approved):
Since this page only populates the popdown with existing cameras, I think this is OK.
(if there had been a "default" or "any" option, this would have been a bad change. But there isn't, at least when I run it).

PR #1728, opened 2026-04-02
resolution: improve compat with non-persistent permissions

fippo:
by not calling it with a empty deviceId and by
repopulating the devices select after getUserMedia

### DIFF
--- a/src/content/getusermedia/resolution/js/main.js
+++ b/src/content/getusermedia/resolution/js/main.js
@@ -153,6 +153,7 @@ function gotStream(mediaStream) {
     widthInput.value = constraints.width.min;
     widthOutput.textContent = constraints.width.min;
   }
+  navigator.mediaDevices.enumerateDevices().then(gotDevices).catch(handleError);
 }
 
 function errorMessage(who, what) {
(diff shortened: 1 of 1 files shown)


fippo:
@nils-ohlmeier says this is a problem in Chrome with non-persistent permissions too

fippo:
it also seems that OverConstrainedError has an empty error message in that case? cc @guidou

fippo:
regressed in #1717, @henbos to write tests ...

nils-ohlmeier:
Confirmed locally. With this PR applied I'm getting a gUM prompt again and afterwards the demo works as expected.

nils-ohlmeier (approved):

src/content/getusermedia/resolution/js/main.js
```diff
@@ -153,6 +153,7 @@ function gotStream(mediaStream) {
     widthInput.value = constraints.width.min;
     widthOutput.textContent = constraints.width.min;
   }
+  navigator.mediaDevices.enumerateDevices().then(gotDevices).catch(handleError);
```
guidou, line 156: indentation?

guidou (approved):
lgtm % comment

src/content/getusermedia/resolution/js/main.js
fippo: that is ok (and linter is happy), what isn't ok is that this shows a promise style where handleError will handle any error thrown in gotDevices.

Will ask minion to fix this as a follow-up.

